PR #149, opened 2014-12-03
cosmetic

athanclark:


dylanmc:
I like the idea of this change, but it looses some key precedence information. I agree it's prettier, but I think we should stick with the current format.

PR #172, opened 2015-02-24
Issue 167

### LINKED ISSUE #167: Empty module is a parser error
This probably shouldn't be the case. The following should be a perfectly fine module:

```
module Foo where
```

bryant:
This pull request fixes #167.

Part of the problem is that `Cryptol.Parser.LexerUtils.layout` never has a chance to insert a matching `VCurlR` to a `VCurlL` inserted immediately an EOF:

```
*Main> mapM_ print . map thing . fst $ lexer defaultConfig "module M where"
Token {tokenType = KW KW_module, tokenText = "module"}
Token {tokenType = Ident "M", tokenText = "M"}
Token {tokenType = KW KW_where, tokenText = "where"}
Token {tokenType = Virt VCurlyL, tokenText = "beginning of layout block"}
Token {tokenType = EOF, tokenText = "end of file"}
```

### DIFF
--- a/src/Cryptol/Parser/LexerUtils.hs
+++ b/src/Cryptol/Parser/LexerUtils.hs
@@ -245,6 +245,9 @@ layout cfg ts0 = loop implicitScope [] ts0
 
     -- add any block start tokens, and push a level on the stack
     (startToks,stack')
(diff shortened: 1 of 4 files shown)


elliottt:
Seems pretty straight forward to me.  Thanks for the fix!

PR #213, opened 2015-05-08
Import Bits Bool orphan instance from base-orphans

RyanGlScott:
This pull request replaces `cryptol`'s orphan `Bits Bool` instance with one exported from the `base-orphans` library. This helps mitigate the possibility of other libraries defining a `Bits Bool` instance and failing to compile if used together with `cryptol` on GHC 7.6 or earlier.

### DIFF
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -43,6 +43,7 @@ library
   Build-depends:       base            >= 4.6 && < 5,
                        array           >= 0.4,
                        async           >= 2.0,
+                       base-orphans    >= 0.1,
                        containers      >= 0.5,
                        deepseq         >= 1.3,
                        directory       >= 1.2,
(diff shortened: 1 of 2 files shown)


RyanGlScott:
This instance doesn't appear to be defined anymore in `cryptol`, so I'll go ahead and close this.

acfoltzer:
Indeed, unfortunately we no longer support GHC versions earlier than 7.8. Let us know if 7.6 support is important for you, though, and we could take another look at what would be required.

RyanGlScott:
It's not a problem for me, personally. This pull request was just one of many I made in a ([still-ongoing](https://github.com/haskell-compat/base-orphans/issues/16)) attempt to get packages that backport instances from newer versions `base` to use `base-orphans` so as to avoid instance conflicts on older GHCs. Now that `cryptol` doesn't backport `instance Bits Bool`, it's no longer a concern.

acfoltzer:
Ah, that's a really useful endeavor. Great work!

Issue #216, opened 2015-05-14
Add a ':write" command


TomMD:
The CLI could use a `:write` command morally in the form `write : [a][8] -> FilePath -> IO ()` (or flipped).   It is getting tedious, copy-pasting 8KB of text into a Haskell program that then writes the hex into a binary file... and that's after only doing it once.

acfoltzer:
Can you say more about the use case? I've thought about something similar before but for input, rather than output (say you want to have a _really_ big key, but don't want to force the parser to read it all in).

Also, does the type signature imply that you'd want it to just be raw binary, or ASCII chars?

TomMD:
@acfoltzer Sure.  I was working on a file encapsulation format and even a bare hand-full of times is too much for copy and pasting an expression into GHCi to perform `writeFile . pack $ someCryptolOutput`.  So the idea, as implemented in patch ab54e71b052ac30ae9efa33579f00a30c2315075, is that the user could:

```
let v = someCryptolExpression
:set outputFile=/tmp/junkfile
:write v
```

There are a few short-comings and warts currently (some we might not care to tackle):

- No `:read` support
- Inconsistent names (`outputFile` vs `smtfile` notice the camel case)
- No stripping of quotes (`:set outputFile="/tmp/ThisDoesNotWork"`)
- I have no idea what I'm doing with Cryptol ASTs - are my assumptions valid? Do `[n][8]` types always come in the form of VSeq-VSeq-BV?  Can the type be in a different "non-normalized" form my hacky check won't recognize?  It seems possible.

TomMD:
As a matter of update:

* You can observe `VSeq [VSeq [VBit ..` but a `pack` function exists to tranform the inner seq to `VWord`.
* There is no resistance to adding n-ary commands (`:write FileName Expr`), it was not preexisting but don't take that as an intentional design decision.

acfoltzer:
> No :read support

Definitely want this eventually, but we should probably be careful about what formats we adopt here. Are there any standardized formats for blobs of bytes we could use, other than just the blob of bytes itself?

> Inconsistent names (outputFile vs smtfile notice the camel case)

I've been meaning to do a cleanup pass of these anyway. The inconsistency is not intended.

> No stripping of quotes (:set outputFile="/tmp/ThisDoesNotWork")

We should probably support this better, particularly for those users who might dare to use paths with spaces in them :wink:

[...]

PR #220, opened 2015-05-20
Add a ':write' command

TomMD:
This pull addresses issue #216 by allowing users to write out bytes (`fin n => [n][8]`) via `:write filename expression`.

I'd appreciate someone (anyone, @acfoltzer ?) else glancing at this.  Certainly any functionality issue should be raised but I'm more concerned about getting a useful API so no one feels like this is half-done.

@yav suggested a `Binary` instance to allow serialization of all sensible Cryptol types into some format, or a hex serialization of `[n]`.  Such a design would satisfy any user who can munge data post-Cryptol but I wanted to avoid making any format implicit in `:write` (including endian-ness).  Certainly yav's comments don't preclude my use case, but I feel what is currently implemented meets the majority of interesting uses.  Other opinions welcome, but I am using this version of `:write` in IRAD and would appreciate either a merge thumbs up or unambiguous requirement for merging.

----

The intent here is to allow users to write expressions to disk without
copy and pasting the ASCII representation into some other system (ex: a
Haskell script) to perform the actual file write.

Use:

```
Cryptol> :write /tmp/foo "hello Cryptol!\n"
Cryptol> :!cat /tmp/foo
hello Cryptol!
Cryptol> let var = "Complex computation\n"
Cryptol> :write /tmp/foo var
Cryptol> :!cat /tmp/foo
Complex computation
Cryptol> let var = 32 : [32]
Cryptol> :write /tmp/foo var
Can not write expression of types other than [n][8].  Type was:  [32]
```


acfoltzer:
This looks quite good. Given the discussion above, maybe we could change `:read` and `:write` to `:readByteArray` and `:writeByteArray` or similar, leaving the more generic names open for if/when we decide to make serialization work for more complex types. Also, a test case or two would be great; something involving roundtripping, something involving putting a KAT in a file, etc?

TomMD:
@acfoltzer   Merged in the latest, fixed merge issues, renamed to read/writeByteArray, added a test write/read/==.  I think this is in a reasonably good state now.

TomMD:
Now with ~ and quoted filepath support.

TomMD:
Merge issues fixed (again).  IMO it should be merged before they diverge again.

acfoltzer:
Thanks!

PR #221, opened 2015-05-21
UTF-8 support

acfoltzer:
Parking this here for discussion. I want to add a few tests cases before merging, but this looks :100:


elliottt:
As there's been no discussion, I'm inclined to just merge this :)

weaversa:
Does this add the support desired in #199?

acfoltzer:
@elliottt: sorry for the delay, I'm ready for discussion now :)

Aside from getting some tests in, would it be possible to switch to non-`Lazy` text? It's not a big deal if not, but the 0MQ interface for the notebook expects strict text instead.

elliottt:
@acfoltzer It would be nice to stick with lazy text, as that's what the unlit process uses.  If we switched to strict, we would likely re-allocate the whole program text before getting to parsing, whereas using lazy text we just break it up into chunks.

@weaversa This change doesn't address #199's request for new operator synonyms, as it only expands the set of allowable characters for identifiers.  Additionally, it makes it more difficult to add special case synonyms like `⊕` for `^`.  We would really need to implement user-defined operators, as in #141, which would then allow all the synonyms to be defined in the Prelude.

acfoltzer:
Closed by c04446b53aeaf8bc9cb30aab57e67ab8bc8aa758

Issue #210, opened 2015-04-30
Add a step in the .msi that prompts for installation location


acfoltzer:
The default path of `C:\Program Files\Galois\Cryptol 2.x.x\` might not work for all users given varying user privileges and the problems that `.msi` installers seem to have with network drives. Let's add a location selection screen to the installer so that users can pick instead.

acfoltzer:
Added in 1b1cdeb

PR #257, opened 2015-07-07
#210 Add install location prompt

hidekiishikawa:
Letting the user pick the installation directory for Cryptol could
potentially help ease a few issues and headaches sometimes encountered
when doing a Windows install. Seems like such a customization option is
pretty standard in most installers.

*User now sees one additional page in the install wizard after the
license agreement screen.


acfoltzer:
Thanks for contributing this! I'll make sure we support this before the next release.

acfoltzer:
This will be part of the 2.4 release. Thanks again!

Issue #258, opened 2015-07-10
'include' of literate files fails


msaaltink:
While I can :load a tex file, and it or a .cry file can `include "something.cry"`, if a file I am loading says `include "somthing.tex"` I get a parse error on the \documentclass line at the top.  Surely it should be possible to `include` a Literate Cryptol file.

elliottt:
Fixed by #266.

PR #266, opened 2015-08-06
Fix for issue #258 (incude of literate file fails)

msaaltink:
Guesses the pre-processing from the file extension, as is done by parseModule.


elliottt:
Looks great, thanks for the fix!

PR #299, opened 2015-12-26
Add some Haskell-like prelude functions.

TomMD:
@elliottt We discussed this before, not sure if you want to weigh in or are willing to do a quick sanity check.

@dylanmc @kiniry Some project direction is implicit in this patch -  I personally use Cryptol to drive SAT.  I hear other people behave this way as well.  The implication is we should add more general-purpose operations to either the language, interprer, or prelude.  I've started on that with the `:{read,write}ByteArray` patch as well as this one.

---

Included in this new prelude are Implication  (==>), not, and, or, all, any, map, foldl, sum, scanl, extend, uniquely, and elem.

Obvious omissions: right fold, right scan, zip operations.

Rational:

I've had to implement these functions several times for different problems.
While my problems were admittedly toy, not cryptographic, the functions are
generally applicable and unlikely to clash with many, if any, pre-existing
operations of different semantic meaning.


lib/Cryptol.cry
```diff
@@ -323,3 +324,87 @@ groupBy : {each,parts,elem} (fin each) =>
...
+ */
+and : {n} (fin n) => [n] -> Bit
+and xs = ~zero == xs
+
+/**
+ * Disjuntion
+ */
+or : {n} (fin n) => [n] -> Bit
+or xs = zero != xs
+
```
acfoltzer: Might it make sense to specialize these first four operations to `Bit` values? 

I suppose `and` and `or` already are, but when I'm dealing with a sequence of bits that I want to treat more as "list of bits" than "word", I like to expand the type abbreviation `[n]` -> `[n]Bit`

Also, typo in "Disjuntion"

TomMD: I disagree with specialization at least for not, and, and or.  These function are rather useful in the polymorphic form:

```
Cryptol> not True
False
Cryptol> not [True, True]
0x0
Cryptol> not [True, False]
0x1
```

We couldn't use `not : Bit -> Bit` if we say `not : [a]Bit -> [a]Bit` and that would almost certainly frustrate people.

On the other hand, specialization for implication seems to make sense.  If we want a vector form them we could add `zipWith` and `zipWIth (==>)`, though then I'd want to work out an under-the-covers optimization.

lib/Cryptol.cry
```diff
@@ -10,6 +10,7 @@ module Cryptol where
  */
 primitive demote : {val, bits} (fin val, fin bits, bits >= width val) => [bits]
 
+infix  5 ==>
```
brianhuffman: I think this should be `infixr` rather than `infix`.

lib/Cryptol.cry
```diff
@@ -323,3 +324,87 @@ groupBy : {each,parts,elem} (fin each) =>
...
+/**
+ * Scan left is like a fold that emits the intermediate values
+ */
+scanl : {b, a, n} (fin n) => (b -> a -> b) -> b -> [n]a -> [n+1]b
+scanl f acc xs = ys
+ where
+  ys = [acc] # [f a x | a <- ys | x <- xs]
+
+/**
+ * Zero extention.
```
brianhuffman: Typo: "extention" should be "extension"

lib/Cryptol.cry
TomMD: Fixed

lib/Cryptol.cry
TomMD: Fixed

lib/Cryptol.cry
TomMD: Fixed for all but `not`.

acfoltzer: Hmm, on one hand, `not` is more consistent with `and` and `or` with the more polymorphic type. But it is also just a prefix synonym for `~`. Is your thought that `not` will just signal to the reader  that we're thinking logically rather than bitwise?

TomMD: Yes, that's exactly my thinking.  A logical skin over `~`.

lib/Cryptol.cry
```diff
@@ -323,3 +324,87 @@ groupBy : {each,parts,elem} (fin each) =>
...
+and xs = ~zero == xs
+
+/**
+ * Disjunction
+ */
+or : {n} (fin n) => [n]Bit -> Bit
+or xs = zero != xs
+
+/**
+ * Conjunction after apping a predicate to all elements
```
brianhuffman: Typo: "apping" should be "applying"

lib/Cryptol.cry
```diff
@@ -323,3 +324,87 @@ groupBy : {each,parts,elem} (fin each) =>
...
+/**
+ * Compute the sum of the words in the array.
+ */
+sum : {a,n} (fin n, Arith a) => [n]a -> a
+sum xs = foldl (+) zero xs
+
+/**
+ * Scan left is like a fold that emits the intermediate values
+ */
+scanl : {b, a, n} (fin n) => (b -> a -> b) -> b -> [n]a -> [n+1]b
```
brianhuffman: Why the `fin n` constraint on `scanl`?

lib/Cryptol.cry
```diff
@@ -323,3 +324,87 @@ groupBy : {each,parts,elem} (fin each) =>
...
+extend : {n,m} (fin m, fin n, m >= n) => [n] -> [m]
+extend n = zero # n
+
+/**
+ * Returns true if the predicate is satisfied by at most one element in the
+ * array.
+ */
+uniquely : {n,a} (fin n) => (a -> Bit) -> [n]a -> Bit
+uniquely p xs =  0b10 > (us ! 0)
+  where us = [0b00] # [ u || u + extend [p x] | x <- xs | u <- us ]
```
brianhuffman: Based on the name of `uniquely`, I expected it to be a unique existence quantifier. I was surprised to realize that it also holds for an always-false predicate, so it is a zero-or-one quantifier. Is this even a concept that people use? There's not a standard mathematical notation for it that I'm aware of. Wouldn't a unique existence quantifier be more useful?

Also, it would be nice to have a definition whose meaning is a little more obvious; the current definition is practically impenetrable. We should also consider the size/complexity of generated SBV/SMTLib expressions of whatever definition we use.

lib/Cryptol.cry
TomMD: Fixed

lib/Cryptol.cry
TomMD: Because I lazily didn't think about it.  Fixed.

lib/Cryptol.cry
TomMD: I've found this notion useful in a couple cases so I threw it in, but admittedly this is rather questionable so I'll remove it.  As for its definition and size, I found both of them to be workable.

weaversa: For `extend`, why `fin m`?

Why not add signed extension as well?

TomMD: @weaversa fin m because:

```
(#) : {front, back,
       a} (fin front) => [front]a -> [back]a -> [front + back]a
```

and `front ~ m`.

@brianhuffman I've removed `uniquely`, though it was useful in some past instances.  I considered the construction efficient, especially when compared with the obviously more  bloated `(< 2) . sum . map extend`, which must include `width n` integral for the sum instead of just a 2 bit state.

TomMD:
I've added uncurry, curry, foldr, scanr, zip, zipWith, repeat, and extendSigned.  Since there didn't seem to be any objections on the basic principle it seemed sensible to address the completeness issue.

lib/Cryptol.cry
brianhuffman: @TomMD I was thinking of a definition that uses only boolean operations instead of mixing bitwise and arithmetic. I've used something like this before:

    oneof : {n} (fin n) => [n] -> Bit
    oneof xs = bs ! 0
      where
        as = [True] # [ a && (~ x) | a <- as | x <- xs ]
        bs = [False] # [ if x then a else b | a <- as | b <- bs | x <- xs ]

lib/Cryptol.cry
```diff
@@ -323,3 +324,133 @@ groupBy : {each,parts,elem} (fin each) =>
...
+ where
+  ys = [acc] # [f a x | a <- ys | x <- xs]
+
+/**
+ * Scan right
+ */
+scanr : {a,b,n} (fin n) => (a -> b -> b) -> b -> [n]a -> [n+1]b
+scanr f acc xs = ys
+    where
+     ys = [acc] # [f x a | a <- ys | x <- reverse xs]
```
brianhuffman: If `scanr` is supposed to match the Haskell one, then the result `ys` should also be reversed; i.e. `acc` should end up on the right end, and `foldr f acc xs` should end up on the left.

TomMD: Gah, foolishness.  Yes, fixed.

lib/Cryptol.cry
```diff
@@ -323,3 +324,133 @@ groupBy : {each,parts,elem} (fin each) =>
...
+ * Signed extension. `extendSigned 0bwxyz : [8] == 0bwwwwwxyz`.
+ */
+extendSigned : {total,n} (fin total, fin n, n >= 1, total >= n+1) => [n]Bit -> [total]Bit
+extendSigned  xs = repeat (xs @ 0) # xs
+
+/**
+ * Repeat a value
+ */
+repeat : {n,a} (fin n, n >= 1) => a -> [n]a
+repeat x = [x | _ <- [1..n] : [n][width n]]
```
brianhuffman: Simpler and more general:

    repeat : {n, a} a -> [n]a
    repeat x = [ x | _ <- zero ]

TomMD: Nice, I love and hate how useful zero is - it seems so hackish in practice.  The fact that zero is necessary to accomplish something this general (I think) is something of a wart. Clearly the next patch should be a 20Kloc change adding dependent types.

lib/Cryptol.cry
```diff
@@ -323,3 +324,133 @@ groupBy : {each,parts,elem} (fin each) =>
   [parts * each] elem -> [parts][each]elem
 groupBy = split`{parts=parts}
 
+/**
+ * Logical implication
+ */
+(==>) : {a} a -> a -> a
+a ==> b = not a || b
```
brianhuffman: Weren't we going to specialize this to type Bit? I would actually prefer to redefine it as

    (==>) : Bit -> Bit -> Bit
    a ==> b = if a then b else True

to make it explicitly lazy in the second argument. (According to the Cryptol semantics that hasn't been written yet, I think bitwise operators like `(||)` will be strict in both arguments.) This will be important for using `(==>)` to state properties with preconditions that guard against undefined behavior.

TomMD: Done.

lib/Cryptol.cry
```diff
@@ -323,3 +324,133 @@ groupBy : {each,parts,elem} (fin each) =>
...
+zipWith f xs ys = [f x y | x <- xs | y <- ys]
+
+/**
+ * Transform a function into uncurried form.
+ */
+uncurry : {a,b,c} (a -> b -> c) -> (a,b) -> c
+uncurry f = \(a,b) -> f a b
+
+/**
+ * Trasform a function into curried form
```
brianhuffman: Typo: "Trasform" should be "Transform"

brianhuffman:
I think things are looking pretty good now. One more general comment about docstrings: We should be consistent in our use of periods vs no-periods. (It seems that most existing docstrings in Cryptol.cry use a period, so we should follow that.)

And while I'm nitpicking, you also have a spelling error in your commit message: "Rational" (adj) should be "Rationale" (noun).

TomMD:
I believe the spelling, comments, horribly incorrect behavior, type signature style, and objectionable functions have all been addressed now.  Any remaining thoughts on either the nuance or the concept anyone?

elliottt:
Looks good to me. I have had to implement a lot of these as well, so I think they'll make a nice addition to the prelude :)

TomMD:
Talked to @dylanmc and he sounds on board with the note that we'd like to think about how to teach Cryptol gradually without introducing all of FP in the beginner tutorial.

Seems @brianhuffman, @elliottt are signed off, so I'm merging.  Blame me, not anyone else, if this is premature in your eyes.

acfoltzer:
Hey, I'm no BDFL :+1:

acfoltzer:
I cherry-picked the non-merge commit into the release branch at c338f9d; @TomMD could you make sure everything is in place?

TomMD:
@acfoltzer Looks good.

weaversa:
Do we have one of those for Cryptol? If not, I recommend FST.

acfoltzer:
@TomMD it looks like this change makes the Prelude take over 10x longer to typecheck (see purple line at the right of the graph). I'm inclined to split the new functions out from the Prelude until performance improves further. Maybe we could start including a `lib` directory in the distribution?

image

PR #301, opened 2016-01-19
Example: miniLock in Cryptol

TomMD:
As requested.

Notice there is much type-signature cruft as this work was done during the worst of the 2.2-2.3 typechecker upset.


TomMD:
Minilock is in.

PR #309, opened 2016-02-05
CAST5 example

clinty:
advice for improvement welcome


acfoltzer:
@TomMD, would you mind picking up this thread?

TomMD:
@acfoltzer Certainly.  I think we can collapse a lot of the manually expanded rounds into a loop but in general having an example of CAST5 is great and I'm all for merging with or with out changes... maybe with a few more KATs but that'd be it.

robdockins:
@TomMD, ping? Any reason not to just merge this?

TomMD:
No, lets merge.

PR #317, opened 2016-03-25
tweak for deepseq-generics-0.2

trofi:
deepseq-generics-0.2 stopped reexporting NFData from deepseq.
Patch imports it explicitly.

Signed-off-by: Sergei Trofimovich <siarheit@google.com>


acfoltzer:
Many thanks! For some reason when I tried to do this, I got tons of unused import warnings as a result, but it seems fine now.

Issue #156, opened 2014-12-30
Precedence of xor is wrong


TomMD:
This is ment as a discussion, the fix is to move one line in Parsery.y.

Consider the parse of:

```
1 ^ 2 == 3
```

Most users would expect:

```
(1 ^ 2) == 3
```

This is true in Haskell and this is true in Cryptol for addition (`+`). But in cryptol the parse is

```
1 ^ (2 == 3)
```

Does anyone actually want this parse?  The only place in the examples that appears to use both these operators together uses parentheses (the first parse).  It's a breaking change but seeing as xor is just vector addition in R(2) I also consider it a consistency issue.

PR #326, opened 2016-05-05
Proposal to fix issue 241

robdockins:
Proposal to fix issue #241 for further discussion.

The main change in this proposal is to move the precedence of `&&` and `||` to bind tighter than the comparison operators `==`, `<=`, etc, and so that `&&` binds more tightly than `^`.  This mostly undoes the changes introduced in #156.

Instead, two new operators are introduced (`/\` and `\/` ) which are short-cutting boolean operations with higher precedence than the comparison operators.

This change is backward incompatible, as it changes parses involving, e.g., `==` and `&&`, as well as the intended change with respect to `^` and `&&`.  The symptom, in most cases, is a parse error.  The fix, in pretty much every case, is to replace `&&` with `/\` and `||` with `\/`.

Is this change too disruptive to existing code?


kiniry:
Define “short-cutting” in this context?

robdockins:
Short-cutting means that if evaluating the first argument is sufficient to determine the value of the boolean operator, the second argument is not evaluated.  This is similar to the C logical operators `&&` and `||`.  In particular, values which would otherwise fail to terminate or cause an error will not be forced.  For example:

```
Cryptol> True \/ undefined
True
Cryptol> undefined \/ True

Run-time error: undefined
Cryptol> False \/ undefined

Run-time error: undefined
```

This follows from the definitions given for /\ and \/, which are:
```
x /\ y = if x then y else False
x \/ y = if x then True else y
```

yav:
@robdockins Did you mean to say that the logical connectives should have _lower_ precedence than the comparison operators?   To summarize, is this the proposal:

    (&&) : {a} a -> a -> a        // pointwise lifted "and", higher precedence than comparison
    (||) : {a} a -> a -> a        // pointwise lifted "or", higher precedence than comparison
    (/\) :: Bit -> Bit -> Bit     // logical "and", lower precedence than comparison
    (\/) :: Bit -> Bit -> Bit     // logica "or", lower precedence than comparison

I guess we still have the choice for the strictness no Bit of the lifted operations.

This would allow us to avoid parens (for better or worse) in expressions like this:

     a && mask == 0x7 /\ b && mask == 0x8

robdockins:
@yav, yes, I meant lower precedence.

Indeed, the question of whether /\ and \/ should have short-cutting semantics or not is a separable question from whether they should exist or not.

robdockins:
An additional thought I had this morning: I wonder if is is useful to have an implication operator with lower precedence than /\ and \/, for use in, e.g., stating properties with preconditions.

```
(==>) : Bit -> Bit -> Bit
a ==> b = ~a \/ b
```

atomb:
There is an implication operator in the `Cryptol::Extras` module, written in terms of `if` rather than `\/`.

robdockins:
This branch is now quite behind, no point in keeping it around.  Continuing discussion of this issue should take place in issue #241.

PR #356, opened 2016-07-18
lib/extras: add iterate

thoughtpolice:
This adds the common `iterate` function to the `Cryptol::Extras` module.


yav:
Thanks!  (This is also related to #350)

PR #374, opened 2016-08-20
Fix up terminal title usage and color output

thoughtpolice:
This adds a `--color` command line option, which allows you to control how colors are output by the frontend. It has three settings: `none`, `always`, and `auto`.

This also adds a fix for the usage of `setTitle` in Cryptol, to set the terminal title. Previously this was always set no matter what, when the REPL was reloaded or a new module loaded, but this breaks inside Emacs, as it can't handle those values within the `ansi-terminal` mode, via `comint`. It can handle ANSI color escape codes, however.

Notably, while color can be controlled explicitly, the autodetection of whether to set the terminal title *cannot* be controlled explicitly. Mostly, I see no reason to support extra code for this. Also, while there may be cases like Emacs where color autodetection should be overridden, I think the general case of title autodetection -- checking if `stdout` is a terminal -- is perfectly adequate.

With both of these commits, Cryptol 2 works far better in `cryptol-mode` (I have some unreleased changes which polish it up a bit).

This does not add documentation, yet.

### DIFF
--- a/cryptol/Main.hs
+++ b/cryptol/Main.hs
@@ -210,7 +210,7 @@ setupREPL opts = do
                 4 (vcat (map pp smoke)))
       exitFailure
   displayLogo True
-  setUpdateREPLTitle setREPLTitle
+  setUpdateREPLTitle (shouldSetREPLTitle >>= \b -> when b setREPLTitle)
   updateREPLTitle
   mCryptolPath <- io $ lookupEnv "CRYPTOLPATH"
   case mCryptolPath of
--- a/cryptol/Main.hs
+++ b/cryptol/Main.hs
@@ -41,12 +41,16 @@ import System.IO (hClose, hPutStr, openTempFile)
 import Prelude ()
 import Prelude.Compat
 
+data ColorMode = AutoColor | NoColor | AlwaysColor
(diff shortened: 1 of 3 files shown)


weaversa:
I'm super excited about an updated `cryptol-mode`!

acfoltzer:
@robdockins @brianhuffman might one of you guys be able to review this? I don't want to keep @thoughtpolice waiting, but I don't have any time to tackle this in the next couple weeks

robdockins:
I don't have any expertise in terminal gunk, but I can take a look. @thoughtpolice, do you consider this PR ready to go?  It would be nice to add the documentation you mention in your comment above.

thoughtpolice:
@robdockins Actually, looking around, there's no actual documetation for several other flags like any of the `cryptolrc` options that can be passed to the executable. So maybe this patch is fine just how it is?

I could make another PR as a follow up documenting the missing set of options, though.

thoughtpolice:
Great, thanks! I've pushed my updates to `cryptol-mode`, so if people are using MELPA it should update soon. `M-x cryptol-repl` behaves *much* better with Cryptol 2 now with these changes.

Issue #408, opened 2017-05-05
Reference evaluator lazier than normal evaluator


msaaltink:
Cryptol> :eval [True, 0b0 == 1/0] @ 0b0
    True
    Cryptol> [True, 0b0 == 1/0] @ 0b0
    
    division by 0

Shouldn't these results agree?

PR #409, opened 2017-05-05
Perform more careful handling of divide-by-0 conditions.

robdockins:
Make values lazier with respect to when they produce divide-by-0
errors.  Divide-by-0 should now propigate in the same way that
user errors do.

In addition, fix a bug in the reference evaluator with respect to
divide-by-0 handling of pmod and pdiv.  Previously, the `polyDivMod`
function would raise a division by zero error, in contravariance
to the reference implementation invariant that errors may only
occur when forcing values at type `Bit`.


robdockins:
Fixes issue #408

PR #434, opened 2017-07-18
Modifications to make cryptol compile with SBV 7.0

LeventErkok:
Also, bumped up the version to 2.4.1

### DIFF
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -1,5 +1,5 @@
 Name:                cryptol
-Version:             2.4.0
+Version:             2.4.1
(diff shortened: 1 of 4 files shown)


LeventErkok:
Obviously the builds will fail since SBV 7,.0 isn't released on hackage yet! Can you do your testing locally?

LeventErkok:
I should add that SBV now requires GHC >= 8.0.

If you guys still need to support GHC 7 for whatever reason, we'll have to find a compromise. (I hope that's not the case though!)

PR #439, opened 2017-08-02
Add operations for signed arithmetic, and carry condition testing.

robdockins:
For your bikeshedding pleasure, a proposed enhancement to address issue #428.

Still to do: add some sanity checks to the test suite.


docs/CryptolPrims.md
```diff
@@ -53,6 +71,10 @@ Boolean
     (^)   : a -> a -> a
     (~)   : a -> a
 
+    (==>) : Bit -> Bit -> Bit
+    (/\)  : Bit -> Bit -> Bit
+    (\/)  : Bit -> Bit -> Bit
+
```
yav: These seem unrelated to the signed arithmetic change?

robdockins: It's true, they're a bit unrelated.  The document had fallen a bit out of date in general.

I can split it into separate patches, if you think it's important.

yav: Honestly, it is such a small change that I don't think it is really worth---it is only a change to docs after all. 

I'd say just merge it in.

PR #440, opened 2017-08-10
Adds the :all command for printing out a parseable AST

sliverdragon37:


cryptol-ast
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+CRYPTOL=/Users/emullen/cryptol/dist/build/Cryptol/cryptol
```
yav: I am not sure that we need this file, but if we have it should not contain hard-coded local paths.

src/Cryptol/Parser/AST.hs
```diff
@@ -172,17 +172,23 @@ data Bind name = Bind { bName      :: Located name -- ^ Defined thing
                       , bPragmas   :: [Pragma]     -- ^ Optional pragmas
                       , bMono      :: Bool         -- ^ Is this a monomorphic binding
                       , bDoc       :: Maybe String -- ^ Optional doc string
-                      } deriving (Eq, Show, Generic, NFData)
+                      } deriving (Eq, Generic, NFData, Functor, Show)
+
+-- instance (Show name) => Show (Bind name) where
+--   show b = "(Bind (" ++ show (bName b) ++ ") (" ++ show (bDef b) ++ "))"
```
yav: Why is this commented out bit here?

src/Cryptol/Parser/AST.hs
```diff
@@ -172,17 +172,23 @@ data Bind name = Bind { bName      :: Located name -- ^ Defined thing
...
-                    deriving (Eq, Show, Generic, NFData)
+                    deriving (Eq, Show, Generic, NFData, Functor)
 
 data Fixity   = Fixity { fAssoc :: !Assoc
                        , fLevel :: !Int
-                       } deriving (Eq, Show, Generic, NFData)
+                       } deriving (Eq, Generic, NFData, Show)
+
+-- instance Show Fixity where
+--   show _ = ""
```
yav: Remove commented out instance

src/Cryptol/Parser/AST.hs
```diff
@@ -985,4 +990,156 @@ instance NoPos (Prop name) where
       CLocated c _  -> noPos c
       CType t       -> CType (noPos t)
 
+-- --Typeclass for prettying up the AST for exporting to Coq
```
yav: Maybe the class and its instances should go into a separate module

src/Cryptol/Parser/Position.hs
```diff
@@ -22,7 +22,10 @@ import Control.DeepSeq
 import Cryptol.Utils.PP
 
 data Located a  = Located { srcRange :: !Range, thing :: !a }
-                  deriving (Eq, Show, Generic, NFData)
+                  deriving (Eq, Generic, NFData)
+
```
yav: Generally we've tried to keep all the information in the `Show` instances, as they are mostly used for debugging.
Do we really need to drop the location info?

src/Cryptol/REPL/Command.hs
```diff
@@ -183,6 +184,10 @@ nbCommandList  =
     "do type specialization on a closed expression"
   , CommandDescr [ ":eval" ] (ExprArg refEvalCmd)
     "evaluate an expression with the reference evaluator"
+  , CommandDescr [ ":ast" ] (ExprArg astOfCmd)
+    "print out an AST of a given term"
+  , CommandDescr [ ":all" ] (ExprArg allTerms)
```
yav: *  The name `all` seems a bit generic, perhaps `ast-all` would work better.
*  Also, it seems wrong to use `ExprArg` as the command has no parameter?

src/Cryptol/REPL/Command.hs
```diff
@@ -589,6 +594,20 @@ refEvalCmd str = do
   val <- liftModuleCmd (rethrowEvalError . R.evaluate expr)
   rPrint $ R.ppValue val
 
+nameToNumber :: M.Name -> Int
+nameToNumber n = M.nameUnique n
```
yav: This seems like an odd place for this declaration, let's move it wherever `M.Name` lives.

cryptol-ast
sliverdragon37: I agree completely.

src/Cryptol/Parser/AST.hs
sliverdragon37: No good reason. Could be deleted.

src/Cryptol/Parser/AST.hs
sliverdragon37: Thanks!

src/Cryptol/Parser/AST.hs
sliverdragon37: Sounds good to me.

src/Cryptol/TypeCheck/AST.hs
```diff
@@ -112,22 +114,98 @@ groupDecls dg = case dg of
...
-                       } deriving (Show, Generic, NFData)
+                       } deriving (Generic, NFData, Show)
 
 data DeclDef    = DPrim
                 | DExpr Expr
                   deriving (Show, Generic, NFData)
 
---------------------------------------------------------------------------------
-
+-- ShowAST prints out just enough information to model the program in Coq
```
yav: * Perhaps this should go into a separate module
* It might be nice to use the pretty printing combinators, in case we ever want to look at the output.

src/Cryptol/Parser/Position.hs
sliverdragon37: I believe it'll work if we keep the Show instance.

src/Cryptol/REPL/Command.hs
sliverdragon37: * ast-all is great
* I didn't fully understand ExprArg when I wrote this. If there's a more appropriate alternative I'm happy with it.

src/Cryptol/REPL/Command.hs
sliverdragon37: Great!

src/Cryptol/TypeCheck/AST.hs
sliverdragon37: Separate module sounds great. What are the pretty printing combinators you mention?

src/Cryptol/Parser/AST.hs
yav: * Also, maybe this should be called something else? 
* Add some comments on what's the purpose of these instances.
* I am not quite clear on what's this class is for, but could we get what you need by generalizing `NoPos` rather than having so much duplicated code?

src/Cryptol/Parser/Position.hs
sliverdragon37: I can make the Show instance into an instance of ShowAST, and it works. I've restored the Show derivation for Located.

src/Cryptol/Parser/AST.hs
sliverdragon37: * How about ShowParseable instead of ShowAST?
* Will do
* Short answer: I don't believe so. NoPos will just set all positions to (0,0), but will leave positions on the object. Further, NoPos only removes the top level of positions, and doesn't recurse on the inner structure. noPos has type t -> t, where showParseable has type t -> String. showParseable also has to print out Let as MLet (to avoid the reserved word in Coq), and perform some other small but necessary transformations in order for the definition printed out to be accepted in Coq.

PR #561, opened 2019-01-15
Make Cryptol compile with SBV 8.0

LeventErkok:
SBV 8.0 is now on hackage, with several backwards compatibility breaking changes. You'll need this patch to compile cryptol against it.


LeventErkok:
@atomb Looking at the build failures, I'm pretty sure they have nothing to do with my commit!

atomb:
Indeed. Keeping our three different CI systems up to date is somewhat time consuming, so sometimes lags a little. ;)

PR #583, opened 2019-03-19
Command to dump generated tests

david-christiansen:
Here's a REPL command to dump generated tests instead of running them, so that they can be consumed by other tools.

I have a couple of questions, pre-merge:

1. The docs for this command are longer than the others due to describing its interaction with options. Should it be shortened, should they be lengthened, or is it OK?

2. Is there a CHANGELOG or NEWS file that I should update?

### DIFF
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -317,8 +317,10 @@ dumpTestsCmd outFile str =
               do argOut <- mapM (rEval . E.ppValue ppopts) args
                  resOut <- rEval (E.ppValue ppopts x)
                  return (renderOneLine resOut ++ "\t" ++ intercalate "\t" (map renderOneLine argOut) ++ "\n")
-     io $ writeFile outFile (concat out)
-
+     io $ writeFile outFile (concat out) `X.catch` handler
+  where
+    handler :: X.SomeException -> IO ()
+    handler e = putStrLn (X.displayException e)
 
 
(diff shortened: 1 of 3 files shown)


yav:
I think that having a longer description is fine.  It would be nice to say what are the arguments to the command (i.e., FILE and EXPR).   I typed `:dumptests True "out.tests"` and was very confused as to why it "didn't work" :)

Other commands like `dumpArray` probably ought to say that too.

Also there is a small typo in the help doc: `noutput` -> `output.

As for question 2, I don't think we currently update any `CHANGES` file or anything.  There is the Cryptol book and `Syntax.md` that we should update when we change the language, but I am not sure that
there is any need to update either of those for this command.

atomb:
This could be an opportunity to start a CHANGELOG file, though!

david-christiansen:
I'm happy to start a CHANGELOG file if it's wanted.

david-christiansen:
So I've fixed the misspelling (it was a remnant of a \n that had been deleted :-] ) and added argument metavariables to all the commands, for the sake of consistency.

ZojaA:
Hi... Why I cannot installed cryptol from official site? Thanks for answer

kiniry:
> Hi... Why I cannot installed cryptol from official site? Thanks for answer

We are not sure what problem you are having nor why you commented in this particular discussion for help.  File a new issue if you are still having difficulty.

PR #630, opened 2019-07-03
Prelude Prims

yav:
This commit moves a bunch of the type level primitives into the Cryptol Prelude.

It also fixes a few problems with the handling of fixities on type level things, and does a little clean up.


src/Cryptol/TypeCheck/TCon.hs
```diff
@@ -1,174 +1,86 @@
...
-  , tInfix  "/^"               TF TCCeilDiv    (l 90)
-    "Divide numeric types, rounding up."
-
-  , tInfix  "%^"               TF TCCeilMod    (l 90)
-    "How much we need to add to make a proper multiple of the second argument."
-
-  , tPrefix "lengthFromThenTo" TF TCLenFromThenTo
-    "The length of an enumeration."
-  ]
+-- | This is used for pretty prinitng.
```
robdockins: What is difficult about getting fixity information from the Prelude?

tests/issues/issue226.icry.stdout
```diff
@@ -16,6 +16,35 @@ Constraint Synonyms
     type constraint i <= j = j >= i
     type constraint i > j = i >= 1 + j
 
+Primitive Types
```
robdockins: We're printing the kind of consraints as `Prop` here, but we're using the `@` symbol in the prelude.  We should pick one.

yav: Good point, I forgot that we do actually print kinds.  Maybe `Prop` makes more sense.

src/Cryptol/TypeCheck/TCon.hs
yav: It's quite annoying.  It's two things, either of which could probably be fixed somehow:  a) the pretty printer does not have access to additional information, such as an environment of fixities, b) alternatively the fixities could be stored directly on the constructors, but that makes it difficult to make new constructors, as now you have to know their fixities (e.g., look at `tMul` in `Cryptol.TypeCheck.Type`).

Another option is the following "hack":  we already use TH to bake in the Prelude in Cryptol, but currently it is just stored as a string.  We could, however, parse it further, and generate the fixity table.

yav: I'll leave this as is for the moment, as I think fixing it requires some changes that are otherwise not very related to this pull request.    I made ticket #634 to help us remember to do it.

yav:
I am merging this as it seems to work OK.

PR #655, opened 2019-10-28
Fix build with GHC 8.8

atomb:
This PR also updates the Travis and AppVeyor configurations to successfully build and test with GHC 8.4, 8.6, and 8.8.

Tests currently fail on AppVeyor, but only because the specifications for correct output aren't flexible WRT file names.


atomb:
Tests now succeed on AppVeyor, so I'm going to merge this.

PR #696, opened 2020-03-31
Rational type

robdockins:
Add a new type of `Rational` values, which are explicitly constructed internally as pairs of integers in the usual way (#694).  Partially, this was a test to see where all the places are that need to be touched to add a new base type. I think I found them all.

As an experiment in language design, along the lines of #686, I added a new typeclass `Field` for field division and reciprocal.

Depends on #684


robdockins:
Things still to do/discuss:
* reference interpreter still needs some updating
* Is the `Field` class a good idea?
* Should `Field` be a structural class in the way `Arith` and `Logic` are?
* How should we pretty print rationals?

robdockins:
Folded into #724

PR #721, opened 2020-05-06
Updates for haskeline 0.8 and use of exceptions package.

kquick:
The version of haskeline was updated for GHC 8.10 (haskeline is one of the bundled libraries in the GHC distribution), and the newer version of haskeline contains some non-backwards-compatible changes that are addressed by this pull request.

More specifically, haskeline dropped its internal MonadExceptions module in favor of using the standard `exceptions` package (pull request 97 in haskeline) which necessitates the addition of some new class instances (and the removal of the MonadException instance) for GHC 8.10.

### DIFF
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -50,6 +50,7 @@ library
                        cryptohash-sha1   >= 0.11 && < 0.12,
                        deepseq           >= 1.3,
                        directory         >= 1.2.2.0,
+                       exceptions,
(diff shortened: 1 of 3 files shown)


kquick:
Good call: I updated the pull request description, @yav.

yav:
@kquick thanks for the update.  My only suggestion would be that in the CPP macros you check for the version of `haskeline` rather than the version of `base` as that's the library that changed.

PR #724, opened 2020-05-15
Split arith

robdockins:
Addreses #686, as well as several other issues.

This splits the `Arith` as discussed in the above ticked, and adds several others.  The arithmetic typeclasses are now `Ring`, `Integral`, `Field` and `Round`.  The types of `(^^)`, `lg2`, `(/$)`, and `(%$)` are specialized to bitvectors.  The types of sequence index and update operators as well as shifts and rotates are generalized to accept any `Integral` type.

There are still some things to clean up here in the reference interpreter and test suite, but currently the basic functionality seems to be working.  A few examples now fail with "ambiguous type" errors that previously were defaulted or whose types were somehow otherwise fixed.  We should track those down and figure out if we can improve defaulting to eliminate them.


robdockins:
Some interesting observations regarding defaulting with the new class arrangement:

* warnings seem to have shifted away from indexing and shifting primitives and instead attach to `number` and `fromTo`, etc.
* `(^^)` is now generating defaulting warnings in its second parameter because it's type is no longer fixed by the first parameter
* There is one case in the examples (`MISTY1.cry`) where we now get an ambiguous type error instead of a defaulting warning.

```
[error] at contrib/MISTY1.cry:230:14--230:22:
   Failed to infer the following types:
     • ?a`1339, type argument 'a' of 'zero' at contrib/MISTY1.cry:230:18--230:22
   subject to the following constraints:
     • Integral ?a`1339
         arising from
         use of expression (!)
         at contrib/MISTY1.cry:230:14--230:22
 [error] at contrib/MISTY1.cry:230:18--230:22:
   Failed to infer the following types:
     • ?a`1339, type argument 'a' of 'zero' at contrib/MISTY1.cry:230:18--230:22
   subject to the following constraints:
     • Zero ?a`1339
         arising from
         use of expression zero
         at contrib/MISTY1.cry:230:18--230:22
```

I think we should change the defaulting rules for `Literal`: if the type is a variable which is unconstrained or has a `Zero`, `Ring` or `Integral` constraint, it should default to Integer without a warning.  If it has a `Field` or `Round` constraint, it should default to `Rational` without warning.  This would squash quite a lot of warnings.

I'm not sure what to do with `(^^)`.  Maybe we could apply the same special-case defaulting rules we currently have for `(@)` and such where we suppress the defaulting warning because the defaulting choice doesn't affect the semantics.

@yav, @brianhuffman, ideas?

yav:
I generally agree with @robdockins's comments on defaulting.  We should probably make a ticket to track and discuss the changes we need to make to the defaulting system.

weaversa:
I'm starting to appreciate #720, aptly titled "Avoid using `fromInteger`".

robdockins:
That is a really interesting interaction between `(^^)` and `fromInteger`.  I agree, we'll have to think pretty carefully about how this should work and how to migrate.  I was disappointed by how many additional defaulting warnings I got when the exponent type of `(^^)` was no longer constrained by the base type.

robdockins:
One additional point: I intended to suggest that the defaulting warning for `(^^)` be suppressed only when defaulting for `Literal`, which wouldn't affect the interaction with `fromInteger` from above.

In fact, I might go so far as to say that we should raise this sort of defaulting situation to an error instead; that is, _only_ default widths for class `Literal` (with or without warnings, depending on situation) and be an error otherwise.

atomb:
One alternative, which I considered while writing that ElectionGuard code, would be to restrict the second argument of `^^` to `Integer`. There are definite downsides to that, though. For one, symbolic exponents couldn't be dealt with in a purely bit-vector development. Ultimately, what I'd really want to do would be to rule out `Z n` as the exponent type, but there's not really a clean way to do that. The three types I can imagine actually being useful are `[n] -> [m] -> [n]`, `Integer -> Integer -> Integer`, and `Z n -> Integer -> Z n` (as well as, maybe, `[n] -> Integer -> [n]`). What do others think about the actually useful instantiations of the general type?

robdockins:
Is there a particular reason to rule out `Z n` as the exponent?  It's at least bounded, so we can do the same square and multiply computation as with `[n]`.

brianhuffman:
This definition looks like it needs the constraint `p >= q`, otherwise it will produce the wrong result:
```
pow : {p, q} (fin p, fin q, p >= 1, q >= 1) => Z p -> Z q -> Z p
pow z n = z ^^ fromInteger (fromZ n)
```
Similarly for this one; if `p` is smaller than `q`, then reducing the exponent mod `p` is going to give the wrong result.
```
pow''' : {p, q, n} (fin p, fin q, fin n, p >= 1, q >= 1, n >= width p) => Z p -> Z q -> Z p
pow''' z n = z ** ((fromInteger (fromZ n) : [n]) % `p)
```

atomb:
We can implement it, for sure, but is it ever useful? My impression (based on very little background, so I may be wrong) is that standard notation for exponentiation on modular integers assumes the exponents to be unbounded integers (even if lots of discussions leave that vague). And if you accidentally interpret them as modular integers, lots of things go silently wrong.

brianhuffman:
I think that raising numbers to `Integer` exponents makes sense only for fields, where we have multiplicative inverses and negative powers are defined.

For rings, the only types that make sense to me for exponents are `Natural` (which we only have at the type level) or `[n]`.

Type `Z n` seems awkward as an exponent because it doesn't fit the usual view of `Z n` values as equivalence classes of integers. Since exponentiation doesn't preserve equivalence modulo `n` in its second argument, we can't treat an exponent of type `Z n` as an equivalence class; rather we have to think of it as representing some particular, arbitrarily-chosen representative from that class. So it seems weird to me, much in the same way that div/mod or `<`/`>` seem weird on `Z n`.

robdockins:
Well, I was contemplating `(^^) : {a, b} (Ring a, Integral b) => a -> b -> a`, which would then include the case `b ~ Z n`.  I have a hard time imagining when you would actually want to do that.  It seems weird to introduce a new class (containing `Integer` and `[n]`) just for this purpose, though.  Alternately, we could remove `Z n` from `Integral` (I always thought including it was a little weird).

CHANGES.md
```diff
@@ -1,8 +1,51 @@
...
+  sequence update and shifting operations are generalized over
+  `Integral`.  The `toInteger` operation is also generalized over this
+  class.  `Integral` contains the bitvector types, `Integer` and `Z n`.
+
+  The new `Field` class contains types representing mathematical
+  fields (or types that are approximately fields). For now, it is
+  inhabited only by the new `Rational` type, but will eventually
+  contain `Real` and floating-point types. It has the operation
+  `recip` for reciprocal and `(/.)` for field division (not to be
+  confused for `(/)`, which is Eucledian integral division).
```
brianhuffman, line 23: "Euclidean"

atomb:
Yeah, I'm starting to think that removing `Z n` from `Integral` might be the right way to go, given @brianhuffman's comment above about using elements of `Z n` as exponents. It also seems weird to use `Z n` elements as array indices or shift amounts.

robdockins:
Well, I suppose there's two useful ways to think about `Z n`.  One is as equivalence classes of integers, as @brianhuffman says.  The other is as the initial segment of the integers from `0` to `n-1` with the various arithmetic operations reduced as necessary.

They're isomorphic in pretty much all the ways that matter, but one view supports an ordered integral view, and the other doesn't.

yav:
@robdockins (This is unrelated to the discussion above).

While you are doing this, would you group the primitives in `Cryptol.cry` together with their
corresponding "class", kind of like a class declaration?

This would help us in the future when adding new primitive types to Cryptol---as it
would make it more obvious as to which primitives you have to implement if you want the new
type to belong to the given class.

robdockins:
I've rearranged the cryptol prelude as @yav suggested.  In addition I've made the following changes after our various discussions:

1) I've removed `Z n` from the integral class.  It therefore does not have integer division or indexing operations.  The `fromZ` operation becomes primitive again.

2) The `(^^)` operation now has type `{a, e} (Ring a, Integral e) => a -> e -> a`.  This reintroduces the possibility of raising to unbounded integer powers, which causes problems with the symbolic backends, but I think this will be a better UX story by allowing `e` to default to Integer for literals when we rework the defaulting.

At the point, I think the only remaining task is to update the reference interpreter.

weaversa:
Maybe I misunderstood, but, without division, how can we compute the mod inverse?

```
isEven : Integer -> Bit
isEven x = ~((fromInteger x) : [1]) ! 0

type constraint isOdd a = (a / 2) * 2 == a - 1

half : {p} (fin p, p >= 3, isOdd p) => Z p -> Z p
half x = if isEven (fromZ x) then x/2 else (fromInteger ((fromZ x + `p) / 2))

mp_mod_inv : {a} (fin a, a >= 1) => Z a -> Z a
mp_mod_inv c = if c == 0 then error "Zero does not have a multiplicative inverse"
               else fromInteger (if u'' == 1 then x1'' else x2'')
  where
    innercond (a, x) = isEven a
    innerbody (a, x) = (a / 2, if isEven x then x / 2 else (x + `a) / 2)

    outtercond (u, v, x1, x2) = (u != 1) /\ (v != 1)
    outterbody (u, v, x1, x2) = if(u' >= v') then (u' - v', v', x1' - x2' % `a, x2')
                                             else (u', v' - u', x1', x2' - x1' % `a)
      where
        (u', x1') = while innercond innerbody (u, x1)
        (v', x2') = while innercond innerbody (v, x2)

    (u'', _, x1'', x2'') = while outtercond outterbody (fromZ c, `a, 1, 0)
```

robdockins:
To get this to typecheck, I had to simply replace `x/2` with `fromInteger (fromZ x / 2)`, which is basically the original semantics of division on `Z n` anyway.  It looks like most of this computation is occuring in `Integer` already anyway.

weaversa:
I see, transformations are the solution  Thanks!

robdockins:
I'm going to make another pass over the documentation to tweak things according to our most recent discussions, but I think this is otherwise ready to go.

src/Cryptol/Eval/Concrete.hs
```diff
@@ -248,6 +269,9 @@ primTable = let sym = Concrete in
                     updatePrim sym updateBack_word updateBack)
 
     -- Misc
+  , ("fromZ"      , {-# SCC "Prelice::fromZ" #-}
```
robdockins, line 272: typo in SCC

src/Cryptol/Eval/Generic.hs
```diff
@@ -319,89 +324,229 @@ arithNullary sym opw opi opz = loop
...
+
+{-# SPECIALIZE computeExponent ::
+      Concrete -> TValue -> GenValue Concrete -> [SBit Concrete] -> SEval Concrete (GenValue Concrete)
+  #-}
+computeExponent :: Backend sym =>
+  sym -> TValue -> GenValue sym -> [SBit sym] -> SEval sym (GenValue sym)
+computeExponent sym aty a bs0 =
+  do onei <- integerLit sym 1
+     one <- intV sym onei aty
+     loop one bs0
```
robdockins, line 454: Note to self: maybe it makes sense to drop concrete leading zeros here before going into the loop?

src/Cryptol/Eval/Generic.hs
```diff
@@ -1346,6 +1522,30 @@ rotateRightReindex sz i shft =
...
+
+-- | Compute the list of bits in an integer in big-endian order.
+--   The integer argument is a concrete upper bound for
+--   the symbolic integer.
+enumerateIntBits' :: Backend sym =>
+  sym ->
+  Integer ->
+  SInteger sym ->
+  SEval sym [SBit sym]
+enumerateIntBits' sym n idx =
```
robdockins, line 1545: Is this the best way to do this?

src/Cryptol/Eval/Reference.lhs
```diff
@@ -592,7 +628,30 @@ Cryptol primitives fall into several groups:
 >   , ("!="         , binary (cmpOrder (\o -> o /= EQ)))
 >   , ("<$"         , binary signedLessThan)
 >
->   -- Sequences:
+>   -- Bitvector
+>   , ("/$"         , vFinPoly $ \n ->
```
robdockins, line 632: Oops, this should use the signed version of `fromVWord`

robdockins:
@jared-w it looks like the GitHub Action for OSX is failing in the test phase because we are missing the CVC4 binary in that environment.  Would it be hard to get that set up?

robdockins:
@jared-w, I'm hoping to merge this soon; what do we need to do to fix the CI?

@brianhuffman, @yav, updates that track this up through SAW now exist, and I plan to merge it pretty soon; if you have comments, please get them in soon.  The only remaining task is to fix some merge conflicts and ensure that shifts and rotates with negative integer amounts work correctly.

yav:
I am fine with you merging it in whenever you are ready.

PR #736, opened 2020-05-26
Real float 2

yav:
This branch implements floating point literals.


src/Cryptol/Symbolic/SBV.hs
```diff
@@ -286,6 +289,14 @@ parseValue (FTTuple ts) cvs = (Eval.VTuple (map Eval.ready vs), cvs')
 parseValue (FTRecord fs) cvs = (Eval.VRecord (Map.fromList (zip ns (map Eval.ready vs))), cvs')
   where (ns, ts)   = unzip (Map.toList fs)
         (vs, cvs') = parseValues ts cvs
+parseValue (FTFloat e p) _ =
```
robdockins, line 292: should pass `cvs` through unchanged here rather than returning `[]`

lib/Cryptol.cry
```diff
@@ -77,6 +77,17 @@ primitive type SignedCmp : * -> Prop
 /** 'Literal n a' asserts that type 'a' contains the number 'n'. */
 primitive type Literal : # -> * -> Prop
 
+/** 'FLiteral n d r a' asserts that the type `a' contains the
+fraciton `n/d`.  The flag `r` indicates if we should round (`r >= 1`)
```
kquick, line 81: fraciton -> fraction

yav:
There's a new version on top of the `split-arith` changes

PR #765, opened 2020-06-11
Float implementation, rebased on current `master`

yav:
This is still work in progress:  I've made things build but there are more changes needed to catch
up with the class refactor.


src/Cryptol/Eval/Concrete/Value.hs
```diff
@@ -313,9 +319,43 @@ instance Backend Concrete where
   znNegate _ 0 _ = evalPanic "znNegate" ["0 modulus not allowed"]
   znNegate _ m x = pure $! (negate x) `mod` m
 
+  ------------------------------------------------------------------------
+  -- Floating Point
+  fpLit                  = FP.fpLit
+  fpEq _sym x y          = pure (FP.bfValue x == FP.bfValue y)
+  fpLessThan _sym x y    = pure (FP.bfValue x <  FP.bfValue y)
+  fpGreaterThan _sym x y = pure (FP.bfValue x <  FP.bfValue y)
```
robdockins, line 327: Looks like `<` should be `>` here

yav: Thanks at @robdockins!.    Hey could you have a look at module `Cryptol.Eval.What4.SFloat`?  I tried to keep it independent of Cryptol, so that we can move it to `what4` eventually.  I tried to mirror the design of `SWord`, but let me know if you think it should be done in another way.

robdockins:
Overall, I think the design of the `SFloat` module looks pretty good; it should be easy to push it down into `what4` when the time comes.  I'd like to see documentation filled out and some commentary about what are the legal settings for the float sizes, but the basic setup looks good.  Obviously, there are a bunch of additional primitives to add, but that should be pretty straightforward work.

src/Cryptol/TypeCheck/Solver/Class.hs
```diff
@@ -348,5 +436,97 @@ solveLiteralInst val ty
       TVar _ -> Unsolved
 
       _ -> Unsolvable $ TCErrorMessage $ show
-         $ "Type" <+> quotes (pp ty) <+> "does not support literals."
+         $ "Type" <+> quotes (pp ty) <+> "does not support integer literals."
+
+
+-- | Add propositions that are implied by the given one.
```
robdockins, line 442: We don't actually use this function right? I think we can remove it.

yav: Hm, I thought that might be part of your super-classing stuff, but maybe I messed something up while I was merging?  If you don't recognize it we should probably get rid of it.

robdockins: I removed it at some point because it's unused.  Probably it got restored when you did a rebase.

yav: OK, I deleted it again

PR #767, opened 2020-06-17
Superclass inference

robdockins:
This PR augments the Cryptol typechecker with superclass information.  Inference and typechecking should both use this information to simplify goals.


yav:
That's pretty much what I was thinking we should do.  I just had a quick look but it looks good to me.

PR #788, opened 2020-06-30
Prover improvements

robdockins:
Fix a number of minor issues with prover selection, and bring What4 mostly up to feature parity with SBV by implementing smoke tests and portfolio solving.

### DIFF
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -43,6 +43,7 @@ library
   Default-language:
     Haskell2010
   Build-depends:       base              >= 4.8 && < 5,
+                       async             >= 2.2 && < 2.3,
                        base-compat       >= 0.6 && < 0.12,
                        bv-sized          >= 1.0 && < 1.1,
                        bytestring        >= 0.10,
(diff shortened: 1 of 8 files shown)


robdockins:
@jared-w, the SBV 8.7 bump I pulled in for a bug fix seems like it requires newer CVC4 versions than we have available.

cryptol.cabal
```diff
@@ -63,7 +64,7 @@ library
                        pretty            >= 1.1,
                        process           >= 1.2,
                        random            >= 1.0.1,
-                       sbv               >= 8.1,
+                       sbv               >= 8.7,
```
jared-w, line 67: Are we also missing a modification [here](https://github.com/GaloisInc/cryptol/blob/62a014384b3528d869c1310b5c13bcfeaf2cdeef/cryptol.cabal#L287)?

robdockins: Oh, I didn't see that one. *shrug* I usually don't like to put bounds on test/benchmark suites for packages that are already dependencies of the main library because they're redundant (as far as I know).  Probably we should just remove it.

jared-w: That's probably what I'd do, especially given that none of the other packages there have constraints.

robdockins:
@atomb, there are a couple of CI issues that are preventing this from merging just yet, but I think the code is ready to review.

PR #789, opened 2020-06-30
Remove GHC 8.4 from CI

jared-w:
This change removes 8.4 from the CI in accordance with supporting the "three most recent" GHC versions.


yav:
Is there any reason to not merge this?

atomb:
Huh, it looks like some of the builds are failing. Do you know what's going on with those, @jared-w?

jared-w:
Not quite sure. It seems to be related to submodules a bit, but that should've been fixed now. I'll re-run it off of the latest master just for kicks and see if that fixes things. If not, I'll probably roll these changes in anyway since they're not causing any of the other breakages.

jared-w:
I have everything rebased and building off of the latest master except one remaining issue: windows + ghc 8.6.5 + hsc2hs (introduced by libBF, I believe)

hsc2hs is used by libBF as a build-tool-depend. Consequently, it gets pulled into the solver and so on windows on GHC 8.6.5 there's an `if os(windows)` check that raises the minimum version of process from 1.1 to 1.6.8 which is higher than the 1.6.5 that our GHC 8.6.5 cabal freeze file pins the process library to.

At that point, I looked into moving hsc2hs down to 68.6 instead of 68.7 (which keeps the lower version bounds) but this version of hsc2hs appears to have some bugs severe enough that it's not recommended to be used on any OS currently. So... I gave up at that point. Not quite sure where to go from there. I suspect leaving hsc2hs out of the freeze file might work?

Issue #787, opened 2020-06-27
Multithreaded support


weaversa:
Is there anything holding back a multithreaded Cryptol interpreter? I think about this every time I write a `map`.

robdockins:
I guess it wouldn't actually be that hard to add some multithreaded support to the interpreter.  I'm slightly worried that we might discover some race conditions in the interpreter, but given the pure nature of Cryptol itself, we might actually be in good shape.

I think the bigger question is: will there be enough inherent parallelism in the algorithms we care about to make it worthwhile? Research in automatic parallelism for functional languages has been fairly disappointing, as I recall.  We might be able to add a parallel map primitive to let the user be more explicit about where to fork parallel tasks... for purely data-parallel applications this might be worthwhile.

weaversa:
I do have something specific in mind ---bBuilding the Merkle tree for PQ Merkle signature schemes is intense, but trivially parallelizable with the parallel map primitive you suggest.

robdockins:
I was curious what it would take to do this; turns out it isn't super hard (at least for the concrete evaluator).  Some very initial, but encouraging results:

```
$ time $CRYPTOL examples/AES.cry -c 'sum (map (\x -> (aesEncrypt (x, zero))) [ 0 .. 100 ])'
Loading module Cryptol
Loading module AES
0x8362b33433e62b81f9a5217bbbb486f9

real    0m14.903s
user    0m14.754s
sys     0m0.136s

$ time $CRYPTOL examples/AES.cry -c 'sum (parmap (\x -> (aesEncrypt (x, zero))) [ 0 .. 100 ])' +RTS -N1
Loading module Cryptol
Loading module AES
0x8362b33433e62b81f9a5217bbbb486f9

real    0m15.926s
user    0m15.774s
sys     0m0.139s

$ time $CRYPTOL examples/AES.cry -c 'sum (parmap (\x -> (aesEncrypt (x, zero))) [ 0 .. 100 ])' +RTS -N4
Loading module Cryptol
Loading module AES
0x8362b33433e62b81f9a5217bbbb486f9

real    0m4.537s
user    0m16.920s
sys     0m0.278s

$ time $CRYPTOL examples/AES.cry -c 'sum (parmap (\x -> (aesEncrypt (x, zero))) [ 0 .. 100 ])' +RTS -N8
Loading module Cryptol
Loading module AES
0x8362b33433e62b81f9a5217bbbb486f9

real    0m4.064s
user    0m28.482s
[...]

PR #792, opened 2020-07-01
Implement an experimental `parmap` operation

robdockins:
This is semantically the same as `map` on finite sequences,
but forks a separate Haskell thread for each element of the sequence.
When the GHC runtime is instructed to use multiple OS threads,
data-parallel execution can be achieved.

CF #787


robdockins:
This turned out to be easier to implement, and more reliable to use than I initially expected. I think we should include it before the next release as an experimental feature.  What do you think @atomb?

atomb:
I'd be in favor of that. What does "experimental" mean in this case, though? Undocumented? Enabled with a particular flag? We've implemented a specific mechanism for marking things as experimental in SAW, but not (yet) in Cryptol.

robdockins:
I'm open to ideas, but for now I just mean "experimental" in the sense that it is a feature that might substantially change or be removed in the future and that users should not rely heavily on it. I guess it's mostly an exercise in expectation management.

We could certainly implement some way to gate it with a flag or something; I don't know how difficult that would be.

atomb:
That sounds fine to me. I just wanted to clarify. I think labeling it as experimental in the release notes will be a start.

PR #793, opened 2020-07-01
Parameterize module system over filesystem access

david-christiansen:
This PR adds the ability for clients of Cryptol-the-library to provide their own means of accessing files on disk, as a sort of lightweight VFS. This will be used for caching multiple concurrent states in the SAW RPC interface.

### DIFF
--- a/src/Cryptol/Eval/Reference.lhs
+++ b/src/Cryptol/Eval/Reference.lhs
@@ -1645,7 +1645,7 @@ This module implements the core functionality of the `:eval
 running the reference evaluator on an expression.
 
 > evaluate :: Expr -> M.ModuleCmd Value
-> evaluate expr (_,modEnv) = return (Right (evalExpr env expr, modEnv), [])
+> evaluate expr (_, _, modEnv) = return (Right (evalExpr env expr, modEnv), [])
 >   where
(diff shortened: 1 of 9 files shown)


robdockins:
I took the liberty of rebasing this on top of current master.

robdockins:
@david-christiansen, is it important to you to get this in before the 2.9 release?

david-christiansen:
Thanks for the rebase! It's not massively important to have it in the release, but OTOH a merge would be nice to prevent bitrot.

What about a merge just after the release, to avoid last-minute pain for library clients?

robdockins:
Makes sense to me

yav:
Is the SAW RPC interface part of the next SAW release. If so, it'd make sense to merge that in before the release?

david-christiansen:
AFAIK the RPC system is not scheduled for a concrete SAW release yet. But @atomb is the one to check with on that.

robdockins:
Well, it's not a difficult change, and downstream PRs already exist to handle it.  I think we should probably just go ahead and merge it.

robdockins:
Internal CI isn't entirely green on this branch, but it all seems to be for other reasons. I'll merge so we can get this into the release.

PR #799, opened 2020-07-02
Actions/fix cvc4 install

jared-w:
(This is a separate PR just to test things out without clobbering over Rob's work)


robdockins:
Relevant fixes were cherry-picked.  Thanks @jared-w!

PR #804, opened 2020-07-06
Holey

robdockins:
See `HoleStatus.txt` for description and status.


robdockins:
I believe we are unlikely to make progress on this in the short term.

PR #816, opened 2020-07-13
Build system and documentation cleanup

atomb:
* Remove `Makefile`
* Remove Travis and AppVeyor configurations
* Improve portability of the `cry` script
* Fix Docker builds
* Update documentation to remove references to `make`
* Update copyright dates


CONTRIBUTING.md
```diff
@@ -257,37 +235,20 @@ with a [GPG key](http://www.cryptol.net/files/Galois.asc).
 
 ## Cutting releases
```
kquick, line 236: Also note that we now have https://gitlab-ext.galois.com/program-analysis/guidance/-/blob/master/HackageRelease.org and while this isn't a Hackage release, we might want to extend that documentation.  It's also not visible from outside the Galois network.

atomb: Yeah, that's a good idea. My focus for these changes is on removing the `make`-based build system, rather than fully fleshing out the contribution documentation, but I think we should do a more thorough job of that soon, too.

PR #821, opened 2020-07-15
Improve release ux

jared-w:


.github/workflows/build.yml
```diff
@@ -90,7 +88,7 @@ jobs:
         run: .github/ci.sh test_dist
 
       - if: >-
-          (startsWith(github.ref, 'refs/tags/v')
+          (startsWith(github.head_ref, 'releases-')
```
atomb, line 91: Should this be `release` instead of `releases`?

jared-w: Yup! good catch. Fixed

PR #822, opened 2020-07-16
Add support for emitting definitional equations during symbolic execution

yav:
…tion.

We use this in the definition of `fpToRational`.  It doesn't seem like
Z3 can do much with these at the moment, but maybe in the future.

### DIFF
--- a/src/Cryptol/Eval/What4/Value.hs
+++ b/src/Cryptol/Eval/What4/Value.hs
@@ -14,7 +14,7 @@ module Cryptol.Eval.What4.Value where
 
 
 import qualified Control.Exception as X
-import           Control.Monad (foldM,ap)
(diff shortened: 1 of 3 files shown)


src/Cryptol/Eval/What4/Value.hs
```diff
@@ -100,7 +155,7 @@ instance W4.IsExprBuilder sym => Backend (What4 sym) where
 
   assertSideCondition _ cond err
     | Just False <- W4.asConstantPred cond = W4Eval (\_ -> pure (W4Error err))
```
robdockins, line 157: I'm concerned here that we might be throwing away important definitional equations in error cases.

robdockins:
Looks pretty good to me.

PR #830, opened 2020-07-22
Remove What4 submodule and depend on hackage-relased What4 instead

robdockins:


robdockins:
Agreed that it doesn't make sense to drop the What4 submodules for the rest of the SAW ecosystem.

For the Cryptol release, we want to depend only on packages that are on Hackage.  Whether that means we should remove the submodule from the Cryptol repository is a separate question, I guess.  Perhaps this PR really only makes sense for the release branch.

atomb:
My understanding (but correct me if I'm wrong, @robdockins) is that Cryptol uses What4 in relatively straightforward and stable ways, and that we aren't planning to have it expand that use in major ways. If that's true, using the Hackage package shouldn't be an issue. And, yes, we want to upload the `cryptol` package to Hackage, so all its dependencies (at the point of release) need to be there, too. My hope is that we could upload new What4 releases often enough that it wouldn't be an issue for Cryptol (even if it is for other tools that use What4).

yav:
I think the important change is the update to the cabal file and that cryptol does work with the hackage version.  Where we get the source (hackage or a submodule) seems less important, although I think it makes sense to remove the submodule if we didn't need it.

After the release it might be worth to split the Cryptol library into multiple ones: front and back-end. Only the back-end needs to depend on what4, and I think SAW only needs to depend on the front end, which should lead to fewer opportunities for conflict.  We may also want to have multiple back-ends, although I am not sure how easy/important that is.

robdockins:
Well, in the meantime, what should we do about the what4 submodule? We should clearly merge the `cryptol.cabal` package bounds.  I don't have a strong opinion about the submodule.

yav:
I think we should remove it.  You probably know better, but I haven't heard of any plans for changes in `what4` that would matter to how we use it in `cryptol`, so I imagine we are going to be fine with the `hackage` version for a while.  It should also help with other projects that depend on `cryptol` as it would decrease the number of nested sub-modules we need to deal with.

robdockins:
Yeah, I don’t think we’ll need any cutting-edge What4 features in Cryptol for awhile.  And if we find we want the submodule again later, it’s easy enough to put it back.

yav:
Right.  The main thing that we may want to use from "future" `what4` is concrete evaluation on `floats` that I guess we may want to add at some point, but indeed it is easy enough to add the submodule if we need it.

PR #867, opened 2020-08-21
Fix race conditions in thunk evaluation

robdockins:
With the addition of `parmap`, it became possible to observe race conditions in the thunk evaluation lifecycle.  This PR reworks the evaluation monad so that threads will properly block and wait if they attempt to force a thunk that is already under evaluation by a different thread and uses transactional variables to ensure proper synchronization.


robdockins:
@jared-w, the failures on Ubuntu seem related to the CVC4 installation.  Did something on those configurations change recently?

PR #869, opened 2020-08-22
Small bug fixes

robdockins:


src/Cryptol/REPL/Command.hs
```diff
@@ -1068,10 +1073,17 @@ loadHelper how =
...
      setDynEnv mempty
 
+versionCmd :: REPL ()
+versionCmd = rPutStrLn versionText
+  where
+  hashText | commitShortHash == "UNKNOWN" = ""
+           | otherwise = " (" ++ commitShortHash ++
+                                 (if commitDirty then ", modified)" else ")")
+  versionText = "version " ++ showVersion version ++ hashText
+
```
yav, line 1083: Perhaps we should use the same version reporting format for this and the command line option?   The code for the command-line option is in `cryptol/Main.hs` function `displayVersion`.   I think it might make sense to make a function `cryptolPrettyVersion` (maybe it should go in `Cryptol.Version`) and use that in both cases.    

I don't think that's a big deal, so I am also fine with leaving it as is.

robdockins: That's a good point.  I modeled this after what appears in the logo, but the command-line printing is probably better.

PR #871, opened 2020-08-27
CVC4 linux fixes

jared-w:


Dockerfile
```diff
@@ -53,13 +55,14 @@ ENV LANG=C.UTF-8 \
     LC_ALL=C.UTF-8
 COPY cabal.GHC-8.8.3.config cabal.project.freeze
 RUN mkdir -p rootfs/usr/local/bin
-RUN cabal v2-update
-RUN cabal v2-install --install-method=copy --installdir=rootfs/usr/local/bin exe:cryptol
-RUN cabal v2-install --install-method=copy --installdir=bin test-lib
+RUN cabal v2-update && \
```
jared-w, line 58: This is the only meaningful change in the dockerfile. As each RUN command is in its own layer, putting cabal v2-update in a separate layer from the  build command does not always do what one wants (particularly when it comes to rebuilding the container with a cache)

PR #878, opened 2020-09-04
Accelerated AES and SHA2 implementations

robdockins:
This is a bit of a work in progress, but the initial results are promising.  Even a fairly naive Haskell implementation runs about 100x faster than interpreting a reference implementation.

Some questions:

1. What underlying implementation(s) should we use? This one was simply the most convenient to get going.
2. What should we do about the symbolic simulator(s)? I'm inclined to find a way to use either defined or uninterpreted functions, provided the underlying solver supports them, but it's a bit tricky to do generically.


robdockins:
@weaversa, I'm interested in your thoughts regarding the Cryptol API exposed via the new `SuiteB` module.  Should there be additions or changes with respect to AES and SHA?  Is there any appetite for, e.g., the other SHA512 variants?

robdockins:
This branch now contains built-in implementations of the following primitives: `AES-128`, `AES-192`, `AES-256`, `SHA-224`, `SHA-256`, `SHA-384`, and `SHA-512`.  They pass a battery of test vectors taken from the NIST Cryptographic algorithm validation program website (https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program), and agree with the reference implementations on randomly selected tests, so I'm reasonably confident they are correct.  These implementations run about 100x faster than directly interpreting the reference implementations.  At a later time, we may wish to replace these primitives with higher performance versions, and/or verified implementations.

Currently, these primitives are not implemented in the symbolic backends.  Attempts to call `:sat` or `:prove` will result in an "unimplemented primitive" error message.  I consider this acceptable for the time being, until we figure out what we want to do.

atomb (approved):
I think this all looks good as it is, but it makes me think of two meta topics.

First, we have `:dumptests` command, and I wonder whether we should have some sort of way of reading and running tests in the same format. That would allow the AES and SHA test vectors to be purely declarative, rather than being interspersed with code.

Second, as of recently we now depend on the `cryptonite` library for part of the RPC API. Would it make sense, given that, to use it for these crypto primitives? Does it expose a useful API? I'm not sure it's worth the effort to change, given that we'll eventually be swapping in C code, but I thought I might as well mention it.

robdockins:
I haven't looked much at the `:dumptests` format, but that might be a good idea... I'm not really sure.

As to `cryptonite`, it seems to have the same problem a lot of the C libraries I looked at for this purpose have; which is that they really want to package up programmer-oriented APIs for larger schemes/modes, whereas we want pretty low-level access to the raw crypto primitives.  The SHA algorithms, in particular, are all seem to be byte-oriented, whereas we want to expose (or be able to expose) the bit-oriented versions.

atomb:
Ah, yeah, I guess it makes sense that `cryptonite` doesn't expose the low-level interface. Anyway, those are both future questions. Let this be merged!

PR #907, opened 2020-09-23
Primitives for ECC on prime fields

robdockins:
This PR adds a new built-in module `PrimeEC` that exposes primitives for elliptic curve cryptography on curves over GF(p) where p is a prime greater than 3.  These routines are taken directly from the primitives of the same name appearing in the ECDSA spec from `cryptol-specs`.

Currently, the primitives use a quite naive modular multiplication, and `ec_twin_mult` is, for now, just a stand-in that performs to separate scalar multiplies and an ECC addition.  Nonetheless, the ECDSA signature scheme runs about 60x faster using the new primitives than when using interpreted ECC routines.  I have performed some basic validation of a few test vectors and comparisons against the reference ECDSA, so I'm somewhat confident of the correctness; however more validation is required.

This builds on PR #882


lib/PrimeEC.cry
```diff
@@ -1,50 +1,145 @@
...
+ */
+type ProjectivePoint p =
   { x : Z p
   , y : Z p
+  , z : Z p
   }
 
+/**
+ * 'ec_is_point_affine b S' checks that the supposed affine elliptic curve
+ * point 'S' fact lies on the curve defined by the curve parameter 'b'.  Here,
```
weaversa, line 41: fact?

robdockins: good catch.  Should be "in fact"

lib/PrimeEC.cry
```diff
@@ -1,50 +1,145 @@
...
+ * Given two projective points 'S' and 'T' where neither is the identity,
+ * compute 'S+T'. If the points are not know to be distinct from the point
+ * at infinity, use 'ec_add' instead.
+ */
+primitive ec_add_nonzero : {p} (prime p, p > 3) =>
+  ProjectivePoint p -> ProjectivePoint p -> ProjectivePoint p
 
-primitive ec_add : {p} (prime p, p > 3) => ProjectivePoint p -> ProjectivePoint p -> ProjectivePoint p
+/**
+ * Given a projective point 'S', compute it's negation, '-S'
```
weaversa, line 117: it's -> its

robdockins:
I think this PR is ready now to merge.

Some basic performance testing on my machine indicates we can do an P-192 ECDSA sign and verify in about 1ms using the new primitives.  This is in contrast to about 147ms required for the interpreted algorithm, and puts us in the neighborhood of 150x speedup.  For another comparison, one P-521 ECDSA sign and verify takes about 4.6ms using the primitives and 410ms when interpreted.  This yields a rather more modest, but still significant, speedup of about 90x.  I think this is unsurprising, as I expect more time to be consumed by the underlying modular arithmetic as the magnitude of the values involved increases.

In both cases, using the primitives also significantly decreases the memory usage and garbage collection time.  For both curves, using the interpreted algorithm, the program gets about 60% productivity, which means 40% of the program runtime is spend in the garbage collector.  In contrast, the version using primitives has about 99% productivity.

robdockins:
@jared-w seems like we've got some (independent?) strange situations going on with the CI jobs.

The Ubuntu jobs are crashing in the test suite; seems to be some change in the CVC4 install?
The Windows jobs are failing before the build even starts, I don't know what to make of that.

atomb:
I have a local fix for the CVC4 issue that I'll push shortly.

atomb:
Err, never mind. That fix is for CVC4 in Docker, not in the main Ubuntu builds.

atomb:
Now there should be a fix on `master` for the CVC4 issue. Try merging it in and see what happens? I'm wondering if the Windows issue is a cache problem.

atomb:
Hmm. Doesn't seem to fix it...

atomb:
I think the current failures are spurious, and something to do with file I/O problems.

robdockins:
Yeah, the situation is very odd, but seems to be unrelated to these changes.  I'm inclined to merge anyway.

Issue #826, opened 2020-07-18
Simple mistake leads to moral crushing error message


weaversa:
So, we're teaching a class on Cryptol. One of the students writes the following:

```
WStep:
    {n}
    (fin n, n >= 2) ->
    ([128] -> [128]) -> [n][64] -> [64] -> [n][64]
WStep CIPHk ([A] # Rs) t = [A'] # Rs'
  where
    [MSB, LSB] = split (CIPHk (A # head Rs))
    A'         = MSB ^ t
    Rs'        = tail Rs # [LSB]
```

Which produces this when read into Cryptol:

```
Loading module Cryptol
Loading module Main
[warning] at bugs.cry:7:36--7:40:
  Defaulting type argument 'n' of 'head' to 0

[error] at bugs.cry:1:1--5:22:
  Type mismatch:
    Expected type: [2][n`910][64]
    Inferred type: [128] -> [128]
  where
  n`910 is signature variable 'n' at bugs.cry:2:6--2:7
[error] at bugs.cry:3:6--3:11:
  Incorrect type form.
    Expected: a value type
    Inferred: a constraint
[error] at bugs.cry:3:13--3:19:
  Incorrect type form.
    Expected: a value type
    Inferred: a constraint
[error] at bugs.cry:5:1--9:33:
  Failed to validate user-specified signature.
    in the definition of 'WStep', at bugs.cry:5:1--5:6,
    we need to show that
      for any type n
      the following constraints hold:
	• fin n
            arising from
            use of expression (#)
            at bugs.cry:7:32--7:43
[error] at bugs.cry:5:28--5:38:
  Type mismatch:
    Expected type: [64] -> [n`910][64]
    Inferred type: [2][n`910][64]
  where
  n`910 is signature variable 'n' at bugs.cry:2:6--2:7
[error] at bugs.cry:7:25--7:30:
  Type mismatch:
    Expected type: [2 * n`910][64] -> [2 * n`910][64]
    Inferred type: (?err, ?err)
  where
  ?err is type error place-holder at bugs.cry:3:6--3:11
  ?err is type error place-holder at bugs.cry:3:13--3:19
  n`910 is signature variable 'n' at bugs.cry:2:6--2:7
```

I'm pretty sure this is the error message that most closely relates to the actual bug:
```
  Incorrect type form.
    Expected: a value type
    Inferred: a constraint
```
But even that doesn't tell the student what he did wrong.

robdockins:
[...]

PR #908, opened 2020-09-23
Typechecker error message improvements

yav:
This addresses issues #746 and #826


brianhuffman (approved):
It looks good to me. My only question is, do the regression tests demonstrate all the possible new error messages? It looks like the issue382 test covers quite a few. If there are any missing, then we should add more tests; otherwise I think it's good.

yav:
Good point, we didn't have these before but I just added some tests which demonstrate all the errors I could make happen.  There were a few that don't seem to fire so I made a ticket to look into that.

Issue #922, opened 2020-09-30
Mechanism for primitives with reference implementations (starting with pmod)


atomb:
The `pmod` operation is one of many examples that can easily be defined in Cryptol but could benefit from being a special case (especially for performance during concrete execution).

It would be useful to add support for such functions. They'd include Cryptol definitions that could be used in any case where special treatment isn't useful (such as in symbolic reasoning backends) but _could_ be treated as if they were primitives when useful (such as for concrete execution).

This would also be beneficial for some of the new crypto primitives.

robdockins:
This is now down for polynomial arithmetic.

robdockins:
There are still some questions to be answered about how to handle the SuiteB and PrimeEC primitives, but I don't think we will get to answering those in time for 2.10.

PR #923, opened 2020-10-01
Fast F2 polynomial arithmetic

robdockins:
Add new primitives for `pmult`, `pdiv` and `pmod`.  Currently, these are exposed in the prelude via new symbols named `fast_pmult`, `fast_pdiv` and `fast_pmod`.  The plan is to relocate the current in-Cryptol implementations to a reference implementation module and continue to use these definitions for the symbolic backends, along the lines suggested in #922. 

These still require additional testing, but seem to work correctly on a limited selection of tests.

### DIFF
--- a/src/Cryptol/Symbolic/What4.hs
+++ b/src/Cryptol/Symbolic/What4.hs
@@ -262,7 +262,7 @@ prepareQuery sym ProverCommand { .. } =
     do let lPutStrLn = M.withLogger logPutStrLn
        when pcVerbose (lPutStrLn "Simulating...")
 
-       ds <- do (_mp, m) <- M.loadModuleFrom (M.FromModule preludeReferenceName)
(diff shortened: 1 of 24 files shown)


src/Cryptol/Symbolic/What4.hs
```diff
@@ -257,12 +260,22 @@ prepareQuery sym ProverCommand { .. } =
   simulate args =
     do let lPutStrLn = M.withLogger logPutStrLn
        when pcVerbose (lPutStrLn "Simulating...")
+
```
robdockins, line 263: This following sequence feels like a bit of a hack.  Binding the top-level decls of the module using `EWhere` does seem to work, but feels pretty dirty.  Any ideas how to handle this better, @brianhuffman, @yav?

robdockins:
I haven't thought of any especially better way to handle the primitive redirection.  We'll just have to be sure we don't rely too much on CAFs being evaluated only once.

PR #927, opened 2020-10-07
Add `cryptol-remote-api`

atomb:
This commit moves the `cryptol-remote-api` package from the `argo`
repository to the `cryptol` repository.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -42,17 +42,6 @@ setup_external_tools() {
   cabal v2-install --install-method=copy --installdir="$BIN" test-lib
 }
 
-setup_venv() {
-  if $IS_WIN ; then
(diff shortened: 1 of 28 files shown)


robdockins:
This module doesn't contain the Python client library, right?  I'm not sure it makes sense to pull over the Python testing scripts in that case.

atomb:
The Python library is still in the `argo` submodule, so running `cabal v2-test cryptol-remote-api` works (given the update to `cabal.project` that I just pushed). Ultimately, I _think_ we should rearrange them so that the Cryptol portions of the client are in the `cryptol` repository, but there are some technical problems to resolve before we can do that.

robdockins (approved):
Seems OK to me; CI is passing currently.

PR #934, opened 2020-10-16
Exposes the E monad, which is needed for C code generation work.

benjaminselfridge:
The C code generation I'm working on calls the reference evaluator, and relies on the `E` type being exposed so we can handle the error conditions.

### DIFF
--- a/src/Cryptol/Eval/Reference.lhs
+++ b/src/Cryptol/Eval/Reference.lhs
@@ -13,6 +13,7 @@
 >
 > module Cryptol.Eval.Reference
 >   ( Value(..)
+>   , E(..)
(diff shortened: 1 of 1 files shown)


yav:
Why does code generation need the `E` monad?

benjaminselfridge:
Because it needs to call `evalExpr` (which is exposed), and do a case split on the result of the computation. I can give more detail if you want, but I think since `evalExpr` is already exposed and returns an `E Value`, it makes sense to also expose `E`, or at least some functions that can convert `E` into something else.

yav (approved):
I see.  Makes sense since it is in the type of the evaluation function.

PR #943, opened 2020-10-28
chore: bump argo submod, tests for BitVector cryptol args

pnwamk:
Need to merge https://github.com/GaloisInc/argo/pull/111 and re-direct submodule bump to that commit first.

### DIFF
--- a/cryptol-remote-api/test-scripts/cryptol-api_test.py
+++ b/cryptol-remote-api/test-scripts/cryptol-api_test.py
@@ -26,9 +26,6 @@ cryptol.add_cryptol_module('Foo', c)
 from Foo import *
 assert add(b'\2', 2) == b'\4'
 
-bv0 = BitVector( intVal = 0, size = 8 )
-bv1 = BitVector( intVal = 1, size = 8 )
-bv2 = BitVector( intVal = 2, size = 8 )
-bv3 = BitVector( intVal = 2, size = 8 )
-
-assert add(bv1, bv2) == add(bv0, bv3)
+assert add(BitVector( intVal = 0, size = 8 ), BitVector( intVal = 1, size = 8 )) == bytes.fromhex('01')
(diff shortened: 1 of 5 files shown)


cryptol-remote-api/test-scripts/cryptol-api_test.py
```diff
@@ -21,3 +22,10 @@
 cryptol.add_cryptol_module('Foo', c)
 from Foo import *
 assert add(b'\2', 2) == b'\4'
+
+bv0 = BitVector( intVal = 0, size = 8 )
+bv1 = BitVector( intVal = 1, size = 8 )
+bv2 = BitVector( intVal = 2, size = 8 )
+bv3 = BitVector( intVal = 2, size = 8 )
+
+assert add(bv1, bv2) == add(bv0, bv3)
```
david-christiansen, line 31: It seems to me that this should not pass, as 1 + 2 does not equal 0 + 2.

Does it pass due to this bug https://github.com/GaloisInc/argo/issues/110 ?

david-christiansen (changes requested):
We should fix the underlying bug along with this, right?

cryptol-remote-api/test-scripts/cryptol-api_test.py
pnwamk: It passes because the test is garbage and is never sanity checking that we actually _do the right thing_ with the BitVector values. (In this case it turns out all were being translated to `0`... oops). Fixed test and impl coming soon.

pnwamk: Okay, with 13210d3 we're not comparing against a non BitVector value so we can't get tripped up there, and everything seems good-to-go with the changes from https://github.com/GaloisInc/argo/pull/111/commits/90df6373fef435c9328e928abb89a2dd26e81907

PR #944, opened 2020-10-28
fix: cryptol-remote-api allow arbitrary expression as operator in call

pnwamk:
Fixes https://github.com/GaloisInc/argo/issues/109


david-christiansen (approved):
The code looks good.

This is a change to the protocol. Will there be a corresponding PR in https://github.com/galoisinc/argo that updates it, and perhaps also marks the `"call"` method as deprecated in favor of `"evaluate expression"`?

PR #952, opened 2020-11-09
REPL defaulting

robdockins:
Add rules to default types at the command line for type variables not constrained by literal constraints.


yav (approved):
Seems reasonable.

PR #968, opened 2020-11-16
cry: propagate command failure

### LINKED ISSUE #953: CI masks build failure
In a recent PR build, the CI task crashed in the middle of a build for some reason.

https://github.com/GaloisInc/cryptol/runs/1381112825?check_suite_focus=true#step:8:93

This is annoying, but more critically, the CI job masked this failure and continued on to the testing phase, and only failed because a test designed to exercise the new behavior failed.

LisannaAtGalois:
Closes #953

### DIFF
--- a/cry
+++ b/cry
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 BIN=bin
 
 QUICKTESTS="tests/issues tests/modsys tests/mono-binds tests/parser tests/regression tests/renamer"


LisannaAtGalois:
The original issue was a spurious crash, likely a problem with the VM. I reproduced the masking behavior by passing an invalid target to `cry build` and observed bash continuing past it: https://github.com/GaloisInc/cryptol/runs/1408891733#step:8:65

The fix here is to make sure that the `cry` script correctly propagates build failures. I tested this fix with my original contrived reproducer and the build correctly fails at that step (after trying it 3 times): https://github.com/GaloisInc/cryptol/runs/1408908452#step:8:67

cry
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
jared-w, line 3: Bit embarrassing that I forgot this in the first place... Good catch

jared-w (approved):

kquick (approved):

robdockins (approved):

PR #979, opened 2020-11-21
Fix compatibility with SBV 8.8 & 8.9

felixonmars:
Tested with SBV 8.9 here and it built fine.


felixonmars:
The test failure looks irrelevant.

robdockins:
Incorporated into #989.  Thanks @felixonmars!

PR #989, opened 2020-11-30
SBV fix

robdockins:
incorporates/rebases the SBV version fixes from #979


robdockins:
I did local builds/tests with SBV versions 8.6, 8.7, 8.8 and 8.9.  I think our version bounds are accurate.

Issue #639, opened 2019-08-15
`:prove`, `:check`, and `:exhaust` error on out-of-scope `import` list exclusion


WeeknightMVP:
```Cryptol
/* M.cry */
module M where
  f (a:Bit) = a
  property p a = f a == a
```

```Cryptol
/* C.cry */
module C where
  import M (f)
```

```
$ cryptol C.cry
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻  
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃  
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 2.7.1 (1e11ff6)

Loading module Cryptol
Loading module M
Loading module C
C> :check
property M::p 
[error] at <interactive>:1:1--1:5 Value not in scope: M::p
C> :exhaust
property M::p 
[error] at <interactive>:1:1--1:5 Value not in scope: M::p
C> :prove
:prove M::p
	
[error] at <interactive>:1:1--1:5 Value not in scope: M::p
```

yav:
I believe the correct behavior here should be the commands check all properties that are *in scope* (either qualified or not).   The current behavior is clearly wrong.

PR #995, opened 2020-12-03
Call stacks

robdockins:
Builds on PR #985 

This PR adds call stack handling to the interpreter, and prints backtraces when evaluation errors occur.  Getting this to work is a bit fiddly, but this seems to be operating more or less the way I expect.

### DIFF
--- a/src/Cryptol/Backend/Concrete.hs
+++ b/src/Cryptol/Backend/Concrete.hs
@@ -137,10 +137,12 @@ instance Backend Concrete where
   type SFloat Concrete = FP.BF
   type SEval Concrete = Eval
 
-  raiseError _ err = io (X.throwIO err)
+  raiseError _ rng err =
+    do stk <- getCallStack
+       io (X.throwIO (EvalErrorEx rng stk err))
 
-  assertSideCondition _ True _ = return ()
-  assertSideCondition _ False err = io (X.throwIO err)
+  assertSideCondition _ True _ _ = return ()
+  assertSideCondition sym False rng err = raiseError sym rng err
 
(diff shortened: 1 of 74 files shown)


src/Cryptol/Backend/Concrete.hs
```diff
@@ -157,9 +159,11 @@ instance Backend Concrete where
        y <- my
        f c x y
 
-  sDeclareHole _ = blackhole
+  sDeclareHole _ rng = blackhole rng
```
robdockins, line 162: humm...

src/Cryptol/REPL/Command.hs
```diff
@@ -412,19 +415,21 @@ data QCMode = QCRandom | QCExhaust deriving (Eq, Show)
...
+qcCmd :: QCMode -> String -> (Int,Int) -> Maybe FilePath -> REPL ()
+qcCmd qcMode "" pos fnm =
   do (xs,disp) <- getPropertyNames
      let nameStr x = show (fixNameDisp disp (pp x))
      if null xs
-        then rPutStrLn "There are no properties in scope." *> return []
-        else concat <$> (forM xs $ \x ->
+        then rPutStrLn "There are no properties in scope."
+        else forM_ xs $ \x ->
+               -- TOOD, this is pretty bogus.  Should find a way to do
```
robdockins, line 425: Note to self: should open a ticket about this

robdockins: Actually, I think #639 is caused by this.  We just take the name of the property and parse it in the current scope, which might fail.

PR #1007, opened 2020-12-10
First cut at adding real numbers to Cryptol

robdockins:
There are some tricky unanswered questions about what
to do with models that return algebraic reals.  I'm almost
inclined to make real numbers explicitly compute on the
algebraic reals, but that would be a farily large undertaking.

In addition, we should probably pay more attention to pretty printing
for real and rational values and give more options for decimal
printing, approximation, precision, etc.


yav:
I am closing this as we don't have a use case for this at the moment, and it has gotten quite out of date.

PR #1010, opened 2020-12-10
Interactive batch

### LINKED ISSUE #1006: Add command line option to treat batch files as individual lines in interactive mode
This would add an additional flag for the `-b FILE` batch mode for cryptol, which would have the effect of reverting to the old behavior. That behavior includes both:
  * using `<interactive>` as the file name in error messages
  * not tracking line numbers in error messages (so any error is reported as occurring on "line 1", regardless of where it appeared in the file)

The reason I need this is to keep the error messages in Programming Cryptol consistent with what the actual error message output would be.

As an example, suppose we have a file `in.cry` containing the following:
```
  True
  false
  False : Bit
  if True && False then 0x3 else 0x4
  False || True
  (True && False) ^ True
  ~False
  ~(False || True)
```

When I run `cryptol -b in.cry`, I get the following output:
```
True
[error] at in.icry:2:1--2:6 Value not in scope: false
False
0x4
True
True
True
False
```

I'd like this flag to have the effect of changing the above output to:

```
True
[error] at <interactive>:1:1--1:6 Value not in scope: false
False
0x4
True
True
True
False
```

robdockins:
Fixes #1006


cryptol/REPL/Haskeline.hs
```diff
@@ -67,12 +75,18 @@ crySession mbBatch stopOnError =
            | all (all isSpace) ls -> loop (lineNum + length ls)
            | otherwise            -> doCommand lineNum ls
 
+  run lineNum cmd =
+    case replMode of
+      InteractiveRepl    -> runCommand lineNum Nothing cmd
+      Batch path         -> runCommand lineNum (Just path) cmd
+      InteractiveBatch _ -> runCommand 1 Nothing cmd
```
robdockins, line 82: Do we really want every line to have line number 1 when running with `InteractiveBatch`?  It would be easy to count up inside a single interactive batch file, which is how the system would actually behave if you typed in the lines sequentially at the prompt...

benjaminselfridge: When I type in sequentially at the prompt, I get the line number as `1` every time.
```
Cryptol> 1
  * Using 'Integer' for the type of '(expression)'
1
Cryptol> a

[error] at <interactive>:1:1--1:2 Value not in scope: a
```

This is the behavior I'm trying to mimic. Not sure if it's going to be useful for anything other than the exercise checker, but that's the functionality I'd need for it to be used for that. 

If this change doesn't make sense for cryptol itself, I could do something else -- namely, have the checker itself ignore those line numbers. But this feature of cryptol would make it a lot easier.

robdockins: This is what you should see in current master:
```
Loading module Cryptol
Cryptol> 1
  * Using 'Integer' for the type of '(expression)'
1
Cryptol> a

[error] at <interactive>:2:1--2:2 Value not in scope: a
Cryptol> a

[error] at <interactive>:3:1--3:2 Value not in scope: a
```

benjaminselfridge: Nevermind -- that was my mistake. I was running an old version of the executable. Let's keep the line numbers incrementing.

PR #1011, opened 2020-12-11
Add check-exercises to CI

LisannaAtGalois:


cryptol/CheckExercises.hs
```diff
@@ -331,7 +331,7 @@ main = do
 
               writeFile outFile outText
 
-              let diffCmd = (P.shell ("diff " ++ outExpectedFile ++ " " ++ outFile))
+              let diffCmd = (P.shell ("diff -u " ++ outExpectedFile ++ " " ++ outFile))
```
LisannaAtGalois, line 334: This makes the diff output in the more familiar git style, if you prefer it the other way I can drop this change ^^

benjaminselfridge:
Looks good to me. @robdockins, can you merge this if you're cool with it?

robdockins (approved):
CI failure seems unrelated.  CVC4 version issue or something that cropped up since yesterday...

Looks OK to me, but let's clean up the git history via rebasing or squash merge.

Issue #870, opened 2020-08-25
Opaque types?


msaaltink:
I sometimes stub my toe by applying one of the built-ins, typically `==` or `+`, to something I shoudn't. The type-checker allows this because so many things satisfy Cmp (or is it Eq?) or Ring.  So it would occasionally be convenient to have a type constructor that is not a member of these classes.  For example, a constructor `opaque` with an interface like

    wrap: {t} t -> opaque t
    unwrap: {t} opaque t -> t

and where `opaque t` is in none of the type classes.  It would be quite convenient if we also had the ability to write things like

    f: opaque t -> ...
    f (wrap x) = ...

There is a sort of a way to fake this now, for example, with

    type opaque t =
      { payload: t
      , junk: [0]( Bit // eliminates Ring, Field, Round, Literal
                 , () -> () // eliminates Eq, Cmp
                 , Integer) // eliminats Logic
      }

    unwrap: {t} opaque t -> t
    unwrap x = x.payload

    wrap: {t} t -> opaque t
    wrap x = { payload = x, junk = []}

This does indeed prevent the inadvertent application of a built-in. It is not really satisfactory, though, as it makes `opaque t` not a testable or provable type.  For example, with type `opaque Bit`

    property tertium_non_datur x =
      unwrap x == True \/ unwrap x == False

in Cryptol we find

    Main> :l opaque.cry
    Loading module Cryptol
    Loading module Main
    Main> :check
    property tertium_non_datur 
    The expression is not of a testable type.
    Type: Main::opaque Bit -> Bit
    Main> :prove
    :prove tertium_non_datur
            Not a valid predicate type:
    Main::opaque Bit -> Bit

So, having a built-in opaque type constructor would be a win here. To be sure this is just a mild suggestion, as the situation comes up infrequently, but it seems like an easy thing to add.

Of course if you do add this feature, someone would be sure to ask why we could not now define a way to add custom definitions for some class, for example, adding my own functions to be `==` and `!=` for my opaque type.

robdockins:
I hesitate to even mention this because we are strongly considering removing this feature, but we actually have had for some time now an experimental `newtype` feature that does essentially what you request here.

```
newtype Box a = { contents : a }
```

```
Main> let x = Box { contents = 11 : Integer }
Main> :t x
x : Box Integer
Main> let y = Box { contents = 42 : Integer }
Main> y
{contents = 42}
Main> x+y

[error] at <interactive>:1:1--1:4:
[...]

PR #1015, opened 2020-12-12
Newtypes

robdockins:
We've had partial support for a while now for a `newtype` construct, but that support was never fully fleshed out.  This PR is for completing the job, writing associated documentation, test cases, etc.

CF #870


robdockins:
@pnwamk, @david-christiansen, There are some nontrivial questions about how to represent values of newtypes in the cryptol remote API, and I'd love you have your opinions about what we should do here.

robdockins:
@yav, @brianhuffman, aside from the questions about the remote API, I think this PR is ready to go.

david-christiansen:
Can we have a video chat about it in the first week of January? Just so there's time to go back and forth a bit about what the feature is used for and its history so we can pick the right trade-offs.

robdockins:
> Can we have a video chat about it in the first week of January? Just so there's time to go back and forth a bit about what the feature is used for and its history so we can pick the right trade-offs.

Sure, I'll set something up.

yav (approved):
I only looked at the diff on `github`, so not in a lot of detail, but it looks reasonable to me.

In the longer run we may want to represent newtypes with a single field a bit more efficiently, but let's worry about that if it every comes up.  I'd say merge it in, this way we'll get to test it by using it.

docs/ProgrammingCryptol/crashCourse/CrashCourse.tex
```diff
@@ -3399,6 +3399,116 @@ \section{Type constraint synonyms}
...
+      r = x.real + y.real
+      i = x.imag + y.imag
+
+  cplxMul : CplxQ -> CplxQ -> CplxQ
+  cplxMul x y = CplxQ { real = r, imag = i }
+    where
+      r = x.real * y.real - x.imag * y.imag
+      i = x.real * y.imag + x.imag * y.real
+
+  cplxEq : CplxQ -> CplxQ -> CplxQ
```
robdockins, line 3445: Note to self : should be `cplxEq : CplxQ -> CplxQ -> Bit`

src/Cryptol/Eval/Reference.lhs
```diff
@@ -506,7 +512,26 @@ the new bindings.
 >   case dDefinition d of
 >     DPrim   -> (dName d, pure (evalPrim (dName d)))
 >     DExpr e -> (dName d, evalExpr env e)
+>
+
+Newtypes
+--------
 
+At runtime, newtypes values are represented in exactly
+the same was as records.  The constructor function for
```
robdockins, line 521: was -> way

PR #1017, opened 2020-12-15
Add an evaluation-only version of cryptol-remote-api

### LINKED ISSUE #1009: Add flag to load Cryptol module immediately when starting cryptol-remote-api
This makes the most common stateful operation happen immediately on startup so we can treat it as essentially a stateless service (for certain use cases). We could also create a different executable that supports this and doesn't support some of the potentially stateful methods.

david-christiansen:
This version requires the file that's in scope to be provided as an
argument. It loads that file on startup, and then provides only
commands that don't change that module context. This means that it can
be used in a stateless load-balancing situation.

Fixes #1009.


weaversa:
With changes coming to the module system (to support some of the ideas such as having multiple modules in a file, or providing a way to instantiate modules during load), it might make sense to consider allowing the user to provide a module name that is in scope of `CRYPTOLPATH` rather than a file.

weaversa:
The change I suggest would also allow loading of the built-in modules such as the prelude, SuiteB, Float, etc.

david-christiansen:
My goal was to match the semantics of `cryptol`'s command line, which wants a file at the moment, or loads the prelude if none is given. But it can definitely be replaced by a mandatory choice between `--file FILE.cry` and `--module MODNAME`.

weaversa:
That works. Providing both options also allows Yahoo's to load files that
don't contain modules.

david-christiansen:
I could also have it attempt to load the string as a filename, and fall back to a module name, but that seems too error-prone for a server process.

david-christiansen:
OK, I think this is ready for review.

atomb (approved):
Looks good! I'll add integration into Docker as a separate PR.

Issue #1004, opened 2020-12-09
Remote API doesn't have a way to update evaluation options


robdockins:
In the following, hardcoded pretty printing options are always provided, and the new setting for enabling/disabling call stack tracking is hardcoded as well.

 https://github.com/GaloisInc/cryptol/blob/49d60aae2afdb37219bb5e05533b286ceb900485/cryptol-remote-api/src/CryptolServer.hs#L24

These should probably be tracked somewhere else (`ServerState`?), and some way provided to update at least the pretty printing options.  It doesn't make a lot of sense to turn call stack tracking off and on at runtime, but there should at least be some way to set it during startup.

david-christiansen:
Do you think this makes sense as global options for the server, or as local options for a particular message to the server? The latter seems to make the most sense to me.

This is because the options in question (pretty-printing and showing call stacks in case of errors) mostly govern the display or formatting of data. It seems to me that a given client may be interested in call stacks for one request, but not for another.

The way I see it, instead of adding a command-line option, we should have a standard way of writing these options down, and then make them into optional fields in `params` for each message. So I could e.g. write `"options": {"show call stack": true, "output": {"base": 16}` as part of the parameters to one request.

PR #1018, opened 2020-12-16
cryptol-remote-api: Add support for general options

david-christiansen:
Now, every method takes an optional parameter field that controls
Cryptol settings like the length of prefixes of infinite data to
display and whether to track call stacks.

RE #1004


david-christiansen:
I don't consider this a complete solution to #1004 because it does not include documentation of the JSON format for the options, but I'm out of hours and figured this is a good start.

atomb (approved):
Thanks! This looks really nice and clean.

david-christiansen:
I got a couple more hours to spend on the project, so I added docs :-)

PR #1023, opened 2020-12-21
Expose `qcExpr` in REPL.Command

abakst:
This change exposes `qcExpr` and `TestReport` in `REPL.Command` so that clients can easily quickcheck terms.


robdockins (approved):
Simple enough, but let's change the names of the `TestReport` fields to remove the leading underscores.  Those were only added to suppress warnings, and will not be necessary now that it is exported.

PR #1031, opened 2021-01-12
Fix helm testing and cvc4 url fix to missing locations

jared-w:


.github/workflows/nightly.yml
```diff
@@ -84,11 +84,11 @@ jobs:
           set -x
           kind create cluster --wait 10m
           kind load docker-image galoisinc/cryptol-remote-api:nightly
-          kind load docker-image test-cryptol-remote-api:latest
+          kind load docker-image galoisinc/cryptol:test-cryptol-remote-api
```
atomb, line 87: I think I'd prefer to have the Docker image containing the remote API still be called `cryptol-remote-api`. Maybe `galoisinc/cryptol-remote-api:test`?

atomb (changes requested):

.github/workflows/nightly.yml
jared-w: That particular image is never built/uploaded anywhere (and doesn't actually contain the remote api; it contains the python code in `cryptol-remote-api/test/test-cryptol-remote-api.py`).

The original reason for the change comes from the `repository` field in docker/build-push-action defaulting to `galoisinc/cryptol` and I think switching to have a consistent version of the action broke how the naming scheme was generated.

I can name it `galoisinc/cryptol-remote-api:test` though.

atomb: Ah, okay, I misunderstood how it was being used. Can't hurt to have that name, though. :)

atomb (approved):

PR #1038, opened 2021-01-15
cryptol-remote-api: submod bump and docs

pnwamk:
Submodule needs updated after https://github.com/GaloisInc/argo/pull/133 is merged - otherwise barring any test failures or similar I believe it's ready basically.

### DIFF
--- a/cryptol-remote-api/cryptol-eval-server/Main.hs
+++ b/cryptol-remote-api/cryptol-eval-server/Main.hs
@@ -1,11 +1,10 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 module Main (main) where
 
-import qualified Data.Aeson as JSON
 import Control.Lens hiding (argument)
-import Data.Text (Text)
 import System.Environment (lookupEnv)
 import System.Exit (exitFailure)
 import System.FilePath (splitSearchPath)
@@ -19,21 +18,24 @@ import Cryptol.TypeCheck.AST (mName)
 import Cryptol.Utils.Ident (ModName, modNameToText, textToModName, preludeName)
 import Cryptol.Utils.Logger (quietLogger)
 
-import Argo (MethodType(..), Method, mkApp)
+import Argo (MethodType(..), AppMethod, mkApp)
 import Argo.DefaultMain
+import qualified Argo.Doc as Doc
 
 
 import CryptolServer
+import CryptolServer.Call
 import CryptolServer.EvalExpr
 import CryptolServer.FocusedModule
 import CryptolServer.Names
 import CryptolServer.TypeCheck
+import CryptolServer.Sat
 
 main :: IO ()
-main = customMain initMod initMod initMod description buildApp
+main = customMain initMod initMod initMod initMod description buildApp
   where
     buildApp opts =
-      mkApp (startingState (userOptions opts)) cryptolEvalMethods
+      mkApp "Cryptol RPC Server" evalServerDocs (startingState (userOptions opts)) cryptolEvalMethods
 
     startingState (StartingFile file) reader =
       do paths <- getSearchPaths
@@ -61,6 +63,12 @@ main = customMain initMod initMod initMod description buildApp
                , evalPPOpts = defaultPPOpts
                }
 
+
+evalServerDocs :: [Doc.Block]
+evalServerDocs =
+  [ Doc.Paragraph [Doc.Text "A remote server for Cryptol that supports only type checking and evaluation of Cryptol code."]
+  ]
+
 description :: String
 description =
   "An RPC server for Cryptol that supports only type checking and evaluation of Cryptol code."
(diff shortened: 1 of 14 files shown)


pnwamk:
With these changes, `cryptol-remote-api doc` emits the following:

```
Cryptol RPC Server
==================

Fundamental Protocol
--------------------

This application is a `JSON-RPC <https://www.jsonrpc.org/specification>`_ server. Additionally, it maintains a persistent cache of application states and explicitly indicates the state in which each command is to be carried out.

Transport
~~~~~~~~~

The server supports three transport methods:


``stdio``
  in which the server communicates over``stdin``and``stdout``
  
  

Socket
  in which the server communicates over``stdin``and``stdout``
  
  

HTTP
  in which the server communicates over HTTP
  
  
In both``stdio``and socket mode, messages are delimited using `netstrings. <http://cr.yp.to/proto/netstrings.txt>`_


Application State
~~~~~~~~~~~~~~~~~

According to the JSON-RPC specification, the ``params`` field in a message object must be an array or object. In this protocol, it is always an object. While each message may specify its own arguments, every message has a parameter field named ``state``.

When the first message is sent from the client to the server, the ``state`` parameter should be initialized to the JSON null value ``null``. Replies from the server may contain a new state that should be used in subsequent requests, so that state changes executed by the request are visible. Prior versions of this protocol represented the initial state as the empty array ``[]``, but this is now deprecated and will be removed.

In particular, per JSON-RPC, non-error replies are always a JSON object that contains a ``result`` field. The result field always contains an ``answer`` field and a ``state`` field, as well as ``stdout`` and ``stderr``.


``answer``
  The value returned as a response to the request (the precise contents depend on which request was sent)
  
  

``state``
  The state, to be sent in subsequent requests. If the server did not modify its state in response to the command, then this state may be the same as the one sent by the client.
  
  

``stdout`` and ``stderr``
  These fields contain the contents of the Unix ``stdout`` and ``stderr`` file descriptors. They are intended as a stopgap measure for clients who are still in the process of obtaining structured information from the libraries on which they depend, so that information is not completely lost to users. However, the server may or may not cache this information and resend it. Applications are encouraged to used structured data and send it deliberately as the answer.
  
  
The precise structure of states is considered an implementation detail that could change at any time. Please treat them as opaque tokens that may be saved and re-used within a given server process, but not created by the client directly.



A remote server for Cryptol.

Methods
-------

change directory (command)
~~~~~~~~~~~~~~~~~~~~~~~~~~


``directory``
  The path to change the current directory.
  
  
Changes the server's working directory to the given path.


load module (command)
~~~~~~~~~~~~~~~~~~~~~


``module name``
  Name of module to load.
  
  
Load the specified module (by name).


load file (command)
~~~~~~~~~~~~~~~~~~~


``file``
  File path of the module to load.
  
  
Load the specified module (by file path).


focused module (query)
~~~~~~~~~~~~~~~~~~~~~~

No parameters

The 'current' module. Used to decide how to print names, for example.


evaluate expression (query)
~~~~~~~~~~~~~~~~~~~~~~~~~~~


``expression``
  The expression to evaluate.
  
  
Evaluate the Cryptol expression to a value.


call (query)
~~~~~~~~~~~~


``function``
  The function being called.
  
  

``arguments``
  The arguments the function is being applied to.
  
  
Evaluate the result of calling a Cryptol function on one or more parameters.


visible names (query)
~~~~~~~~~~~~~~~~~~~~~

No parameters

List the currently visible (i.e., in scope) names.


check type (query)
~~~~~~~~~~~~~~~~~~


``expression``
  Expression to type check.
  
  
Check and return the type of the given expression.


satisfy (query)
~~~~~~~~~~~~~~~


``prover``
  The SMT solver to use to check for satisfiability. I.e., one of the following: ``cvc4``, ``yices``, ``z3``, ``boolector``, ``mathsat``, ``abc``, ``offline``, ``any``, ``sbv-cvc4``, ``sbv-yices``, ``sbv-z3``, ``sbv-boolector``, ``sbv-mathsat``, ``sbv-abc``, ``sbv-offline``, ``sbv-any``, .
  
  

``expression``
  The predicate (i.e., function) to check for satisfiability; must be a monomorphic function type with return type Bit.
  
  

``result count``
  How many satisfying results to search for; either a positive integer or ``all``.
  
  
Find a value which satisfies the given predicate (i.e., a value which when passed to the argument produces true).
```

atomb:
This is excellent! I noticed a few bits of weirdness in the text that might be nice to fix, though. Some of the ` ``stdout`` ` and ` ``stderr`` ` instances are missing spaces around them. And the text "A remote server for Cryptol." seems like it should be either a heading or the only text under some other heading.

pnwamk:
New doc output:
```rst
Cryptol RPC Server
==================

Fundamental Protocol
--------------------

This application is a `JSON-RPC <https://www.jsonrpc.org/specification>`_ server. Additionally, it maintains a persistent cache of application states and explicitly indicates the state in which each command is to be carried out.

Transport
~~~~~~~~~

The server supports three transport methods:


``stdio``
  in which the server communicates over ``stdin`` and ``stdout``
  
  

Socket
  in which the server communicates over ``stdin`` and ``stdout``
  
  

HTTP
  in which the server communicates over HTTP
  
  
In both ``stdio`` and socket mode, messages are delimited using `netstrings. <http://cr.yp.to/proto/netstrings.txt>`_


Application State
~~~~~~~~~~~~~~~~~

According to the JSON-RPC specification, the ``params`` field in a message object must be an array or object. In this protocol, it is always an object. While each message may specify its own arguments, every message has a parameter field named ``state``.

When the first message is sent from the client to the server, the ``state`` parameter should be initialized to the JSON null value ``null``. Replies from the server may contain a new state that should be used in subsequent requests, so that state changes executed by the request are visible. Prior versions of this protocol represented the initial state as the empty array ``[]``, but this is now deprecated and will be removed.

In particular, per JSON-RPC, non-error replies are always a JSON object that contains a ``result`` field. The result field always contains an ``answer`` field and a ``state`` field, as well as ``stdout`` and ``stderr``.


``answer``
  The value returned as a response to the request (the precise contents depend on which request was sent)
  
  

``state``
  The state, to be sent in subsequent requests. If the server did not modify its state in response to the command, then this state may be the same as the one sent by the client.
  
  

``stdout`` and ``stderr``
  These fields contain the contents of the Unix ``stdout`` and ``stderr`` file descriptors. They are intended as a stopgap measure for clients who are still in the process of obtaining structured information from the libraries on which they depend, so that information is not completely lost to users. However, the server may or may not cache this information and resend it. Applications are encouraged to used structured data and send it deliberately as the answer.
  
  
The precise structure of states is considered an implementation detail that could change at any time. Please treat them as opaque tokens that may be saved and re-used within a given server process, but not created by the client directly.



Summary
-------

An RCP server for `Cryptol <https://https://cryptol.net/>`_ that supports type checking and evaluation of Cryptol code via the methods documented below.


Methods
-------

change directory (command)
~~~~~~~~~~~~~~~~~~~~~~~~~~


``directory``
  The path to change the current directory.
  
  
Changes the server's working directory to the given path.


load module (command)
~~~~~~~~~~~~~~~~~~~~~


``module name``
  Name of module to load.
  
  
Load the specified module (by name).


load file (command)
~~~~~~~~~~~~~~~~~~~


``file``
  File path of the module to load.
  
  
Load the specified module (by file path).


focused module (query)
~~~~~~~~~~~~~~~~~~~~~~

No parameters

The 'current' module. Used to decide how to print names, for example.


evaluate expression (query)
~~~~~~~~~~~~~~~~~~~~~~~~~~~


``expression``
  The expression to evaluate.
  
  
Evaluate the Cryptol expression to a value.


call (query)
~~~~~~~~~~~~


``function``
  The function being called.
  
  

``arguments``
  The arguments the function is being applied to.
  
  
Evaluate the result of calling a Cryptol function on one or more parameters.


visible names (query)
~~~~~~~~~~~~~~~~~~~~~

No parameters

List the currently visible (i.e., in scope) names.


check type (query)
~~~~~~~~~~~~~~~~~~


``expression``
  Expression to type check.
  
  
Check and return the type of the given expression.


satisfy (query)
~~~~~~~~~~~~~~~


``prover``
  The SMT solver to use to check for satisfiability. I.e., one of the following: ``cvc4``, ``yices``, ``z3``, ``boolector``, ``mathsat``, ``abc``, ``offline``, ``any``, ``sbv-cvc4``, ``sbv-yices``, ``sbv-z3``, ``sbv-boolector``, ``sbv-mathsat``, ``sbv-abc``, ``sbv-offline``, ``sbv-any``, .
  
  

``expression``
  The predicate (i.e., function) to check for satisfiability; must be a monomorphic function type with return type Bit.
  
  

``result count``
  How many satisfying results to search for; either a positive integer or ``all``.
  
  
Find a value which satisfies the given predicate (i.e., a value which when passed to the argument produces true).
```

atomb (approved):
Looks good!

PR #1046, opened 2021-01-25
Add remote API calls for for proving properties

atomb:
This currently depends on GaloisInc/argo#144, which we should merge first.


pnwamk (approved):
LGTM

PR #1065, opened 2021-02-09
Init ghcr.io/galoisinc/cryptol-remote-api:nightly-portable

LisannaAtGalois:
Builds cryptol-remote-api without eventfd and timerfd

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -79,7 +79,6 @@ jobs:
           repository: galoisinc/cryptol-remote-api
           push: false
           dockerfile: cryptol-remote-api/test/Dockerfile
-          build_args: GHCVER=8.10.3
       - name: Test cryptol-remote-api helm chart
         run: |
           set -x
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -91,6 +90,33 @@ jobs:
           kubectl run --rm --attach test-http --image=galoisinc/cryptol-remote-api:test --image-pull-policy=Never --restart=Never -- http cra-http-cryptol-remote-api 8080
           kubectl run --rm --attach test-socket --image=galoisinc/cryptol-remote-api:test --image-pull-policy=Never --restart=Never -- socket cra-socket-cryptol-remote-api 8080
 
+  docker-cryptol-remote-api-portable:
```
LisannaAtGalois, line 93: Building GHC with 2 cores takes an Extremely Long Time(tm), but its build can be cached against `ghcr.io/galoisinc/cryptol-remote-api:nightly-portable-cache`.

.github/workflows/nightly.yml
```diff
@@ -91,6 +90,33 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker/setup-buildx-action@v1
+        with:
+          install: true
+      - id: docker_meta
```
LisannaAtGalois, line 99: Using this pattern enables use of both registry caching from buildx and the git labels from `build-push-action@v1`

LisannaAtGalois:
Pre-loaded the cache with:

```bash
docker buildx build \
  --load \
  -f cryptol-remote-api/Dockerfile \
  --build-arg PORTABILITY=true \
  -t ghcr.io/galoisinc/cryptol-remote-api:nightly-portable \
  --cache-from type=registry,ref=ghcr.io/galoisinc/cryptol-remote-api:nightly-portable-cache \
  --cache-to type=registry,ref=ghcr.io/galoisinc/cryptol-remote-api:nightly-portable-cache,mode=max \
  https://github.com/GaloisInc/cryptol.git\#heads/lisanna/cra-portability
```

.github/workflows/nightly.yml
atomb: I'm not familiar with buildx specifically, but I'm definitely in favor of caching stuff between CI runs and artifact creation whenever we can reliably do so.

.github/workflows/nightly.yml
atomb: :+1:

atomb (approved):
This looks good to me. I don't know anything about buildx or much about how `ghcr.io` works, but from what I can tell it does what I'd hope. Thanks!

cryptol-remote-api/Dockerfile
```diff
@@ -43,18 +44,40 @@ RUN latest="$(curl -sSL 'http://cvc4.cs.stanford.edu/downloads/builds/x86_64-lin
...
     XDG_BIN_HOME=/usr/local/bin \
     XDG_DATA_HOME=/usr/local/share
 COPY --from=solvers /solvers/rootfs /
-RUN ghcup install ghc ${GHCVER} && ghcup set ghc ${GHCVER} && ghcup install cabal --set
+RUN ghcup install cabal --set
+ENV PATH /root/.cabal/bin:$PATH
+ADD ./cryptol-remote-api/ghc-portability.patch .
+ARG GHCVER
+ARG GHCVER_BOOTSTRAP
+RUN if ${PORTABILITY}; then \
```
jared-w, line 62: I think a lot of this ghc install logic can be largely replaced with `ghcup compile ghc -b ${GHCVER_BOOTSTRAP} -v ${GHCVER} -p /patches` (assuming the `ghc-portability.patch` is relocated to `/patches`).

But that's something to explore later rather than delay this.

jared-w (approved):
Nice!

cryptol-remote-api/Dockerfile
LisannaAtGalois: I don't think this is holding anything up, there's already a working version in the registry that we've sent them. Refactoring that now will be less expensive than doing it later. @atomb ?

atomb: Sure, if you're available and interested to simplify it now, go for it!

LisannaAtGalois: Looked at it and `ghcup compile` won't work here, since it compiles from source tarballs. The portability patch affects the generated `./configure` scripts via `configure.ac` and so must be applied before autoconf is run (`./boot`). Source tarballs already have the `./configure` files built so applying the patch there wouldn't make a difference.

jared-w: Awesome, good to know. There's a way to do it with ghcup; I think it might be possible to simply reconfigure it, but a custom bindist could certainly be used with the patch already applied; at that point, though, there's not a lot of ergonomics saved.

PR #1069, opened 2021-02-10
Float check

robdockins:
Implement `:exhaust` for floating point values, and improve the random generator to produce more representative samples.


brianhuffman (approved):
I agree that having `:exhaust` test multiple `NaN` values is not a problem: If the float type is small enough to be exhaustively tested, then you're going to be spending a relatively small amount of time checking `NaN` cases anyway.

For the random test generation, there seems to be a lot of arbitrary choices in there, and I'm wondering whether simply using `genBinary` by itself would suffice (with a small modification to replace `NaN` values with infinities half the time). In any case, the extra heuristics shouldn't hurt, so it probably doesn't matter much.

PR #1070, opened 2021-02-10
Fix printing of parse errors in enumeration sequences

### LINKED ISSUE #975: Raw type printing in some errors
```
Cryptol> [0, -1 .. 0 ]

Parse error at <interactive>:1:5--1:7
  The boundaries of .. sequences should be valid numeric types.
The expression `ELocated (ENeg (ELocated (ELocated (ELit (ECNum 1 (DecLit "1"))) (Range {from = Position {line = 1, col = 6}, to = Position {line = 1, col = 7}, source = "<interactive>"})) (Range {from = Position {line = 1, col = 6}, to = Position {line = 1, col = 7}, source = "<interactive>"}))) (Range {from = Position {line = 1, col = 5}, to = Position {line = 1, col = 7}, source = "<interactive>"})` is not.
```

It seems we are using `show` directly on type expressions instead of using the pretty printer in this instance, and perhaps in others.

robdockins:
Previously, raw AST would be printed instead of pretty-printed syntax
Also tweak the indentation for multiline parse errors.

Fixes #975


brianhuffman (approved):
Looks good to me.

PR #1072, opened 2021-02-11
Improved kind-checking for parameters

robdockins:
Restrict the kinds that are allowed for most flavors of parameters.


brianhuffman (approved):
Looks good to me.

PR #1075, opened 2021-02-12
Persist solver

robdockins:
Refactor the use of SMT solvers for typechecking to allow for the solver instances to be reused for a longer period.


brianhuffman (approved):
This looks like exactly what we want for saw-script.

PR #1088, opened 2021-02-23
chore: make missing records fields an error for cryptol-remote-api

pnwamk:


pnwamk:
Fix remote server w.r.t. changes in b3d142a76bd03a5073615c795124ac3156620394 and turn missing record fields into an error to avoid further changes sneaking by without notice.

PR #1097, opened 2021-03-03
test: script to build and test cryptol-remote-api docker image

pnwamk:


LisannaAtGalois (approved):
This looks good for now! It doesn't add a whole lot to the runtime. I can make an optimization pass on this later when I spend time looking at our docker build processes as a whole.

PR #1118, opened 2021-03-17
Make cryptol-remote-api mypy-clean

RyanGlScott:
A handful of tweaks needed to make `cryptol-remote-api`'s Python code `mypy`-clean:

* `__del__` needs an explicit `-> None` return type.
* `BitVector` has no type annotations, so `type: ignore` it.

While I was in town, I:

* Made sure that running `mypy` is a part of CI.
* Added a `pyproject.toml` file for running Python-related commands with `poetry`, much like we have in `saw-remote-api`.

### DIFF
--- a/cryptol-remote-api/run_rpc_tests.sh
+++ b/cryptol-remote-api/run_rpc_tests.sh
@@ -11,6 +11,13 @@ python3 -m venv virtenv
 . virtenv/bin/activate
 pip install -r requirements.txt
 
+echo "Typechecking code with mypy..."
+mypy cryptol/ tests/
+if [ $? -ne 0 ]; then
+  echo "Code failed to typecheck with mypy"
+  NUM_FAILS=$(($NUM_FAILS+1))
+fi
+
 export CRYPTOL_SERVER=$(cabal v2-exec which cryptol-remote-api)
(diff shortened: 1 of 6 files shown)


cryptol-remote-api/run_rpc_tests.sh
```diff
@@ -11,6 +11,12 @@ python3 -m venv virtenv
 . virtenv/bin/activate
 pip install -r requirements.txt
 
+mypy cryptol/ tests/
+if [ $? -ne 0 ]; then
+  echo "Code failed to typecheck with mypy"
+  exit 1
+fi
```
pnwamk, line 18: I would prefer we did not abort the tests entirely if typechecking fails, but rather increment the `NUM_FAILS` variable declared in that file. It is nice to get as much feedback as possible from a test run rather than abort early IMO

pnwamk (commented):
Looks great to me - just that one comment w.r.t. letting the tests do as much work as possible before reporting failure.

cryptol-remote-api/run_rpc_tests.sh
RyanGlScott: Good idea. I've done that in my latest push.

pnwamk (approved):

PR #1121, opened 2021-03-18
Update release notes for 2.11

atomb:


CHANGES.md
```diff
@@ -1,4 +1,11 @@
-# NEXT
+# 2.11.0
+
+## Language changes
+
+* The `newtype` construct, which has existed in the interpreter in an
```
robdockins, line 5: Perhaps worth mentioning that the `cryptol-remote-api` doesn't really support newtypes yet.  There's some questions to deal with regarding the generative nature of newtype that haven't fully been worked out yet, unless @pnwamk has done some work in that area I haven't noticed.

CHANGES.md
```diff
@@ -9,6 +16,36 @@
   runtime overhead, but may be disabled using the `--no-call-stacks`
   command-line option.
 
+* The `:exhaust` command now works for floating-point types.
+
+* The `:check` command now uses more representative sampling of random
+  input values to test.
```
robdockins, line 22: This is also WRT floating-point values, I think?

robdockins (approved):

CHANGES.md
atomb: Ah, good. Thanks for clarifying that. I'll update the text.

Issue #422, opened 2017-05-16
strictness of indexing


msaaltink:
Here's another place where the reference evaluator and REPL evaluator disagree:

    Cryptol> :eval [(),()] @ 3
    ()
    Cryptol> [(),()] @ 3
    
    invalid sequence index: 3

I think the reference evaluator actually *is* strict in the index, but the domain for () has no room for an error, as we see in

    Cryptol> (error "?"):()
    ()

We of course see the same issue with the other three indexing operators `!`, `@@`, and `!!`.

msaaltink:
This issue afflicts other primitives that take a strict number, in cases where the type of the result is a singleton.  For example

    Cryptol>  [(),()] << [True, error "x"]

    Run-time error: x
    Cryptol> :eval [(),()] << [True, error "x"]
    [(), ()]

or

    Cryptol> :eval [(),()] <<< [True, f 0x0] where f n = f(n+1)
    [(), ()]
    Cryptol> [(),()] <<< [True, f 0x0] where f n = f(n+1)

where the REPL gets in an infinite loop.

So the REPL does not appear to have the domains advertised in the Reference implementation, in particular we have extra error and bottom elements.

robdockins:
Indeed, these are pretty tricky issues.  We could solve them by doing the necessary eta-expansion at every indexing call, but I suspect that would result in an enormous performance degradation in the common cases.

I have some vague ideas in mind for how we could deal with these issues better.  The most fully formed is this: in the evaluator, always be evaluating in the context of a "demand shape", which identifies both the desired result type, and how much of the resulting shape the use site promises to examine.  This would essentially perform a runtime analogue of demand analysis.  Then, using that information, we would have enough context to know when we need to perform expensive eta-expansions, and when we can just go ahead and force things.  In many cases, I hypothesize, we would get the more efficient eager evaluation behavior, while still correctly handling the tricky corner cases.

robdockins:
The reference interpreter was updated in #866, which removes some, but not all, of these mismatches.

robdockins:
Fixed via #1136

PR #1136, opened 2021-03-26
Word eval refactor

robdockins:
Refactor various aspects of the evaluator, mostly dealing with the packed word representation, but also aiming to remove  the eta-delay mechanism, which should no longer be required with the current evaluation semantics.  The main trick, I think, is to find a way to implement sufficiently-lazy sequence primitives without also introducing unnecessary unpacking or thunking.


robdockins:
I think this refactoring has reached a pretty good point to merge.  The main goals of this were to factor out the `SeqMap` and `WordValue` types into separable modules with abstract APIs, remove the legacy eta-delay mechanism, and to fix such strictness bugs as I could without harming performance.  This should help enable the more aggressive performance-related refactoring stages I have in mind, as well as eventually allowing better reuse inside the SAWCore evaluator. As a side benefit, I was able to merge more of the primitive operations to use identical code paths; all the shift operations now use unified code paths.  Among the core operations, only sequence indexing and update have separate definitions anymore.

With the performance considerations in mind, this PR contains a lot of small, self-contained commits and I spent a lot of time performance testing after each change.  It should be relatively straightforward to bisect for performance problems inside this PR, as I believe each commit leaves the project in a testable state.  At the end, the benchmarks I tried ended up performance neutral, or slightly faster than they started, with the exception of the Karatsuba example, which slowed down somewhat due to a lazier `#` operator.

Along the way, I generalized the type of the `take` operation as discussed in issue #1064.  This required making `take` and `drop` primitive instead of `splitAt`, which will have minor downstream consequences for `cryptol-saw-core`.

The next stages of refactoring are quite a bit more speculative, but should hopefully be largely contained within the `SeqMap` and `WordValue` modules.

robdockins:
I think this last series of patches lets us finally close out the remaining known strictness bugs in the evaluator, #640, #619, and #422.

Issue #1081, opened 2021-02-16
cryptol-remote-api changes for the entire server and all states


pnwamk:
`cryptol-remote-api` uses [setCurrentDirectory](https://github.com/GaloisInc/cryptol/blob/6181f3213f8da30e2942c3dd99cfe53ac6ca063f/cryptol-remote-api/src/CryptolServer/ChangeDir.hs#L25) to support the `change directory` command. This invalidates any other state expecting a different working directory, including new connections which wish to start working from the initial state.

pnwamk:
Proposed solution: remove `change directory` from the Cryptol server, and instead rely on the `CRYPTOLPATH` for making loads more convenient/relative.

weaversa:
> Proposed solution: remove `change directory` from the Cryptol server, and instead rely on the `CRYPTOLPATH` for making loads more convenient/relative.

I second this solution.

PR #1137, opened 2021-03-26
feat(rpc): extend search path, remove cd

pnwamk:
Disable the ability for clients to change the current working directory of a running server - instead make it trivial to extend the search path dynamically. See https://github.com/GaloisInc/cryptol/issues/1081

- [x] add a test
- [x] verify the eval server is okay with this feature (it did not have the ability to change a working directory before... but extending the search path seems harmless and unlikely to violate eval server restrictions)
- [ ] investigate if `saw-remote-api` also needs this feature


pnwamk:
@atomb In this PR I've added the ability for clients to extend the `CRYPTOLPATH` the server uses to find modules for both `cryptol-remote-api` and `cryptol-eval-server`. I just want to double check this functionality doesn't conflict with the goals of the latter.

cryptol-remote-api/python/tests/cryptol_eval/test-files/test-subdir/Bar.cry
```diff
@@ -0,0 +1,7 @@
+module Bar where
```
atomb, line 1: It's a little unfortunate to need two copies of this file, but it's not really a big deal.

atomb (approved):
This looks good to me.

cryptol-remote-api/python/tests/cryptol_eval/test-files/test-subdir/Bar.cry
pnwamk: Yah, the way we have testing set up currently means it's quite convenient to have distinct directories for the standard and eval server... perhaps theres a set of "common test cases" that should work the same for both and we could reduce some duplication that way?

PR #1159, opened 2021-04-08
feat(rpc): check, AES example/test

pnwamk:
Essentially adds `:check` to the Python interface.

Partially addresses remaining issues here: https://github.com/GaloisInc/argo/issues/121

Adds an additional simple test for the Python interface (i.e., `test_AES.py`).


cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -126,6 +125,54 @@ def __init__(self, connection : HasProtocolState, fun : str, args : List[Any]) -
     def process_result(self, res : Any) -> Any:
         return from_cryptol_arg(res['value'])
 
+
+@dataclass
+class CheckReport:
```
atomb, line 130: I wonder whether it would be useful to include the values leading to failure in the case where check fails, in a form compatible with `prove`. In some sense, `check` and `prove` should be usable somewhat interchangeably, so making them compatible when possible seems useful.

atomb: Oh, never mind. It does. That's what `args` is. I guess the one additional possibility is to include names for those arguments, like I recently added to `prove` (and I think maybe should be added to `prove` and `sat` on the Cryptol side). But that could be in a separate PR later, I think.

atomb (approved):

atomb (approved):
Looks good!

Issue #1166, opened 2021-04-17
cryptol-remote-api + :safe


weaversa:
Are there plans to add `:safe` to the Python client for `cryptol-remote-api`? It would be a nice addition to a CI framework.

pnwamk:
Yes, I just started working on it actually. I’ll report back here once it
lands next week.

pnwamk:
Fixed via https://github.com/GaloisInc/cryptol/commit/966b3431c19a25d978901b58dfca9d2ec15a65fd

PR #1168, opened 2021-04-19
feat(rpc): safe for python api

pnwamk:
Add an equivalent to the REPL's `:safe` to the python API.

Addresses https://github.com/GaloisInc/cryptol/issues/1166

### DIFF
--- a/cryptol-remote-api/python/cryptol/__init__.py
+++ b/cryptol-remote-api/python/cryptol/__init__.py
@@ -206,7 +206,7 @@ class CryptolProveSat(argo.Command):
         if res['result'] == 'unsatisfiable':
             if self.qtype == SmtQueryType.SAT:
                 return False
-            elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAT:
+            elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAFE:
                 return True
             else:
                 raise ValueError("Unknown SMT query type: " + self.qtype)
--- a/cryptol-remote-api/python/cryptol/__init__.py
+++ b/cryptol-remote-api/python/cryptol/__init__.py
@@ -200,7 +200,7 @@ class CryptolProveSat(argo.Command):
         if res['result'] == 'unsatisfiable':
             if self.qtype == 'sat':
                 return False
-            elif self.qtype == 'prove':
+            elif self.qtype == 'prove' or self.qtype == 'safe':
                 return True
             else:
                 raise ValueError("Unknown prove/sat query type: " + self.qtype)
@@ -212,7 +212,7 @@ class CryptolProveSat(argo.Command):
                     for m in res['models']
                     for arg in m]
         else:
-            raise ValueError("Unknown sat result " + str(res))
+            raise ValueError("Unknown prove or sat result " + str(res))
 
 class CryptolProve(CryptolProveSat):
     def __init__(self, connection : HasProtocolState, expr : Any, solver : solver.Solver) -> None:
@@ -221,6 +221,10 @@ class CryptolProve(CryptolProveSat):
 class CryptolSat(CryptolProveSat):
     def __init__(self, connection : HasProtocolState, expr : Any, solver : solver.Solver, count : int) -> None:
         super(CryptolSat, self).__init__(connection, 'sat', expr, solver, count)
+
+class CryptolSafe(CryptolProveSat):
+    def __init__(self, connection : HasProtocolState, expr : Any, solver : solver.Solver) -> None:
+        super(CryptolSafe, self).__init__(connection, 'safe', expr, solver, 1)
 
 class CryptolNames(argo.Command):
     def __init__(self, connection : HasProtocolState) -> None:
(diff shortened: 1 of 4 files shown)


cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -200,7 +200,7 @@ def process_result(self, res : Any) -> Any:
         if res['result'] == 'unsatisfiable':
             if self.qtype == 'sat':
                 return False
-            elif self.qtype == 'prove':
+            elif self.qtype == 'prove' or self.qtype == 'safe':
```
RyanGlScott, line 203: This isn't the fault of your patch, but I'm somewhat leery about using strings here instead of an enumeration of some kind. Without looking, I wasn't sure whether `'prove'`, `'safe'`, and `'sat'` were the only possible options.

cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -212,7 +212,7 @@ def process_result(self, res : Any) -> Any:
                     for m in res['models']
                     for arg in m]
         else:
-            raise ValueError("Unknown sat result " + str(res))
+            raise ValueError("Unknown prove or sat result " + str(res))
```
RyanGlScott, line 215: Should this mention "`safe`" as well here?

pnwamk: Perhaps it should just say "Unknown SMT result"? I.e., something more general that describes the underlying dependency/failure instead of naming all of our API functions that utilize it...?

RyanGlScott: Sounds good to me. I especially like the idea of having fewer places to remember to keep in sync with the rest of the implementation.

cryptol-remote-api/python/cryptol/__init__.py
pnwamk: This is a good point -- I think some type annotations using `Union[Literal['prove'], Literal['safe'], Literal['sat']]` will allow us to use strings (which is convenient for JSON compat) and be clear/precise.

RyanGlScott: Does it even need to be that fancy? I was thinking of something like:

```py
class QueryType(str, Enum):
    PROVE = "prove"
    SAFE = "safe"
    SAT = "sat"
```

pnwamk: It might be hard to convince me that my solution is more fancy than the `str` mixin `Enum` solution =) I'll give the enum a try.

RyanGlScott: Admittedly, I'm not as familiar with `Literal`s as I am with `Enum`s. One of the reasons I would slightly prefer `Enum`s is that instead of writing `elif self.qtype == 'prove' or self.qtype == 'safe'`, you'd write `elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAT`. Although this is somewhat more verbose, it makes it clearer that the choices are coming from a fixed set of values rather than from arbitrary strings.

Ultimately, however, I'm not too opinionated on this. If you think the `Literal` approach would be better, I'll defer to your judgment.

cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -198,9 +204,9 @@ def __init__(self, connection : HasProtocolState, qtype : str, expr : Any, solve
...
     def process_result(self, res : Any) -> Any:
         if res['result'] == 'unsatisfiable':
-            if self.qtype == 'sat':
+            if self.qtype == SmtQueryType.SAT:
                 return False
-            elif self.qtype == 'prove':
+            elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAT:
                 return True
             else:
                 raise ValueError("Unknown prove/sat query type: " + self.qtype)
```
RyanGlScott, line 212:
```suggestion
                raise ValueError("Unknown SMT query type: " + self.qtype)
```

cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -198,9 +204,9 @@ def __init__(self, connection : HasProtocolState, qtype : str, expr : Any, solve
 
     def process_result(self, res : Any) -> Any:
         if res['result'] == 'unsatisfiable':
-            if self.qtype == 'sat':
+            if self.qtype == SmtQueryType.SAT:
                 return False
-            elif self.qtype == 'prove':
+            elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAT:
```
RyanGlScott, line 209:
```suggestion
            elif self.qtype == SmtQueryType.PROVE or self.qtype == SmtQueryType.SAFE:
```

cryptol-remote-api/python/cryptol/__init__.py
pnwamk: I agree with all of your intuitions/arguments/etc around enums here -- they do make things clearer in this case in many ways! =)

RyanGlScott (approved):

PR #1171, opened 2021-04-20
T1169

yav:
This fixes, hopefully, some incorrect behavior of import declarations


src/Cryptol/ModuleSystem/Renamer.hs
```diff
@@ -161,8 +161,10 @@ renameModule' thisNested env mpath m =
          allImps   = openLoop allNested env openDs imps
 
      (inScope,decls') <-
-        shadowNames allImps $
```
robdockins, line 164: This was basically the source of the spurious overlapping errors, right?

src/Cryptol/ModuleSystem/Renamer/Monad.hs
```diff
@@ -250,10 +250,16 @@ checkEnv check (NamingEnv lenv) r rw0
 
   where
   newEnv         = NamingEnv newMap
-  (rwFin,newMap) = Map.mapAccumWithKey doNS rw0 lenv
+  (rwFin,newMap) = Map.mapAccumWithKey doNS rw0 lenv  -- lenv 1 ns at a time
   doNS rw ns     = Map.mapAccumWithKey (step ns) rw
 
-  step ns acc k xs = (acc', [head xs])
+  -- namespace, current state, k : parse name, xs : possible entities for k
```
robdockins, line 256: what's this change about?

tests/modsys/nested/T15.icry.stdout
```diff
@@ -1,15 +1,6 @@
 Loading module Cryptol
 Loading module Cryptol
 Loading module T15
-[warning] at T15.cry:5:13--5:14
```
robdockins, line 4: I'm not sure how I feel about losing these warnings.  What do you think?

src/Cryptol/ModuleSystem/Renamer.hs
yav: In short yes, because we were checking that different imports don't define the same thing, but is actually OK if they do as long as we don't use.  Btw, I find the input argument to `shadowNames` to be quite confusing, and we may want to change in a future refactoring.

src/Cryptol/ModuleSystem/Renamer/Monad.hs
yav: Apart from the comment, there was another bug here.   This code basically does some checking and adds the names to the environment.   If the names come from a bunch of imports (the "NoCheck" case), it is possible that names are ambiguous, and we are not reporting an error.  In that case, we should not arbitrarily pick the first possible resolution, but rather keep all the names, so that we can report an ambiguity error if one of the names is actually used.

tests/modsys/nested/T15.icry.stdout
yav: I wondered about that too.  More generally, I wonder if we should issue a warning when a local name shadows an imported name.  This is what the comment on line 166 of `Renamer.hs` is about.

robdockins: I feel like it probably makes sense to do that, controlled by the `warnShadowing` option.

.github/ci.sh
```diff
@@ -39,7 +39,7 @@ retry() {
 
 setup_external_tools() {
   is_exe "$BIN" "test-runner" && return
-  cabal v2-install --install-method=copy --installdir="$BIN" test-lib
+  cabal v2-install --install-method=copy --installdir="$BIN" test-lib-0.3
```
ldettwy, line 42:
```suggestion
  cabal v2-install --install-method=copy --installdir="$BIN" test-lib
```

PR #1181, opened 2021-04-23
rpc: Move contents of __init__.py to new commands.py, connection.py files

m-yac:
This mirrors the organization of the SAW python bindings, and will enable us to export a new interface from `__init__.py` which depends on the old. The exports of `__init__.py` should stay the same, so no scripts should have to be updated.


cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -16,599 +16,17 @@
 from . import cryptoltypes
 from . import solver
 from .bitvector import BV
+from .commands import *
+from .connection import *
 
 
 __all__ = ['cryptoltypes', 'solver']
```
pnwamk, line 23: Should this include `commands` and `connection` as well now? (That may require additional imports such as `from . import commands` if so.)

cryptol-remote-api/python/cryptol/__init__.py
```diff
@@ -2,8 +2,17 @@
...
+import os
+from enum import Enum
+from dataclasses import dataclass
+from distutils.spawn import find_executable
+from typing import Any, List, NoReturn, Optional, Union
+from typing_extensions import Literal
+
+import argo_client.interaction as argo
+from argo_client.interaction import HasProtocolState
+from argo_client.connection import DynamicSocketProcess, ServerConnection, ServerProcess, StdIOProcess, HttpProcess
```
pnwamk: Does adding these `argo_client` imports and others like them back in actually do anything for users of this package? E.g., if someone has `import cryptol`, does that make any of these names in-scope and/or accessible?

m-yac: From what I can tell, yes. If you `import cryptol` then you can refer to `cryptol.find_executable`, for example, since the line `from distutils.spawn import find_executable` is in `__init__.py`.

Of course, it would be gross to actually do this – but I realized I couldn't claim "no scripts should have to be updated" unless I kept these lines. If we're fine with the small chance of breaking someone's script, it would make more sense to delete them.

cryptol-remote-api/python/cryptol/__init__.py
m-yac: I suppose so? I don't know what the philosophy should be behind `__all__` – it already seemed strange to me that the behavior of `from cryptol import *` was to only import `cryptoltypes` and `solver`, and not `connect`, etc.

pnwamk: Yah, I was just going off what's already there and skimming this: https://docs.python.org/3/tutorial/modules.html

Probably worth adding them if for no other reason than consistency...?

pnwamk (approved):
I would add those two names to `__all__` for consistency and then I think this is good if everything works unmodified (which it looks like it does).

cryptol-remote-api/python/cryptol/__init__.py
m-yac: Ah, that link is very helpful - thanks!

atomb (approved):
Looks good to me, too. (I'm assuming that the large removal and addition chunks are just to move code, rather than change it, so I haven't looked at them carefully.)

PR #1184, opened 2021-04-27
Functors

yav:


Riib11:
Is it correct behavior both `test1` and `test2` to work here? Or should only `test2` work?
```
test1 = hash
import Hash::SHA256 (hash)
test2 = hash
```

Riib11:
> Should it be possible to declare a submodule within a private block? For example
> 
> ```
> module A where
> 
> private
> 
> submodule B where
>   ...
> 
> ...
> ```

There is currently a bug that comes up relating to this:

`A.cry`:
```
module A where

import A::B
import submodule C
```

`A/B.cry`:
```
module A::B where

private

submodule C where
  f x = x
```

If I run `A.cry` like so:
```shell
> ./cry run A.cry
```
then I get the following error
```
  Revision:  a0b43eef33b6863e0bc3ec68d685a5c8ba873073
  Branch:    functors (uncommited files present)
  Location:  nameModPath
  Message:   Not a top-level name: 
             Name {nUnique = 4673, nInfo = LocalName NSModule (Ident False "C"), nFixity = Nothing, nLoc = Range {from = Position {line = 1, col = 1}, to = Position {line = 1, col = 1}, source = ""}}
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-2.12.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/ModuleSystem/Name.hs:249:16 in cryptol-2.12.0.99-inplace:Cryptol.ModuleSystem.Name
```

However, I am able to run `A/B.cry` just fine like so:
```shell
> ./cry run A/B.cry
```
Instead of crashing, it should just give an error right?

PR #1195, opened 2021-05-07
feat(rpc): make python client compat with python 3.7

pnwamk:


cryptol-remote-api/run_rpc_tests.sh
```diff
@@ -10,27 +10,22 @@ cabal v2-build exe:cryptol-eval-server
...
+function run_test {
+    "$@"
+    if (( $? != 0 )); then NUM_FAILS=$(($NUM_FAILS+1)); fi
+}
 
 echo "Setting up python environment for remote server clients..."
-python3 -m venv virtenv
-. virtenv/bin/activate
-pip install -r requirements.txt
+poetry install
```
RyanGlScott, line 19: CI is failing due to not being able to `poetry`, so you'll most likely need to add a GitHub Actions step to install it along [these lines](https://github.com/GaloisInc/saw-script/blob/2d92f2a9ef91d65f58ecb6679f34a5528b160b4e/.github/workflows/ci.yml#L208-L210).

cryptol-remote-api/python/cryptol/bitvector.py
```diff
@@ -178,6 +178,10 @@ def __index__(self) -> int:
         """Equivalent to ``self.value()``."""
         return self.__value
 
+    def __int__(self) -> int:
```
RyanGlScott, line 181: Is this needed to make this work with Python 3.7? Or is this a new piece of functionality? I'm not necessary opposed to adding it, but if this is a new feature, perhaps it should belong in its own commit.

RyanGlScott: Ah, that makes sense. Thanks for the clarification!

RyanGlScott (approved):

PR #1199, opened 2021-05-25
Add a field order option for the pretty-printer

dmwit:
We're working on a cryptol-to-C compiler, and one way we've been validating that it's working right is to use the cryptol repl's `:dumptests` feature and a small C wrapper that treats them as golden tests. I'd like to keep the C wrapper's parser simple, and if records get printed with a predictable field order, that simplifies the parser a lot.

This pull request adds a feature for this: a new setting, `fieldOrder`/`field-order` in the REPL, which can be set to `canonical` or `display`, and which controls the pretty-printer for records. It defaults to `display` for backwards compatibility.

(Actually, I'm going to use a different branch than `master` for the C compiler, but I'd like to minimize drift between them if possible.)


yav (approved):
Seems reasonable to me

robdockins (approved):

PR #1200, opened 2021-05-25
initial newtype support for rpc server

pnwamk:


cryptol-remote-api/python/cryptol/cryptoltypes.py
```diff
@@ -450,6 +457,8 @@ def to_type(t : Any) -> CryptolType:
         return Tuple(*map(to_type, t['contents']))
     elif t['type'] == 'record':
         return Record({k : to_type(t['fields'][k]) for k in t['fields']})
+    # elif t['type'] == 'newtype':
+    #     return Newtype(t['name'], t['uid'], map(to_type, t['arguments']))
```
atomb, line 461: IIUC, now it's only possible to write newtypes as concrete syntax, and that's why this alternative is no longer relevant. Is that right? And, if so, should we just remove it rather than commenting it out?

atomb (approved):
This looks good to me! There probably are bits of it that would be useful on the SAW side, and perhaps we should eventually refactor to make them usable by both servers, but I think getting this working now is more important, and we can do any refactoring in a later PR (or PRs).

cryptol-remote-api/python/cryptol/cryptoltypes.py
pnwamk: Yes, I think that's basically it --  it's either concrete syntax from the client which get's it's meaning/type/etc there or it's an opaque value sent from the server which already has a known type (and whose known type may not be representable syntactically for the same reasons we've introduced the framework for sending back opaque newtype values to clients). I've deleted the commented out code as you suggested 👍

PR #1203, opened 2021-05-27
Support TLS in cryptol-remote-api

### LINKED ISSUE #1008: Enable cryptol-remote-api over HTTPS
What's the best way to enable this? What are the options?

We may be able to use nginx in front on the server side, and the Python HTTP library may already support it, but we haven't tested it.

ldettwy:
Use scotty-tls to add TLS support to cryptol-remote-api.

- [x] Builds completely
- [x] https://github.com/GaloisInc/argo/pull/165
- [x] ~~scotty-tls updated in hackage~~ AMK: we basically inlined the one function needed to enable TLS in argo, so this isn't an issue anymore
  - [x] https://github.com/scotty-web/scotty-tls/pull/9
  - [x] N/A ~~https://hackage.haskell.org/package/scotty-tls-0.4.2~~
- [x] demo with example certificates from https://github.com/scotty-web/scotty-tls#scotty-tls
- [x] Choose between http and https via env var or from command line
- [x] Basic CI test for HTTPS functionality
- [x] `curl -k` equivalent in python: https://github.com/GaloisInc/argo/issues/168 @pnwamk 

Closes #1008


ldettwy:
This isn't building right now because there's a compatibility issue with the scotty-tls in hackage and warp: https://github.com/scotty-web/scotty-tls/pull/8

https://gitpod.io#snapshot/d3211af8-74e6-496a-adbb-8bf4a3d67660

ldettwy:
TLS usage:

```bash
# Generate a certificate. Self-signed example:
openssl req -nodes -newkey rsa:2048 -keyout server.key -out server.csr \
    -subj "/C=GB/ST=London/L=London/O=Acme Widgets/OU=IT Department/CN=localhost"
openssl x509 -req -days 365 -in example.csr -signkey server.key -out server.crt

# Change owner uid of the certificate and key to `1000` to match what's in the container image:
chown 1000 server.crt server.key

# Start server:
docker run --rm -ti --detach \         
  -v $PWD/server.crt:/home/cryptol/server.crt \
  -v $PWD/server.key:/home/cryptol/server.key \
  -e TLS_ENABLE=1 \
  -p 443:8080 \
  ghcr.io/galoisinc/cache-cryptol-remote-api:portable-1203-merge

# Connect to server
curl -k \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -XPOST \
  -d '{"jsonrpc":"2.0","id":1,"method":"load module","params":{"module name":"Cryptol","state":null}}' \
  https://localhost/
```

Result:
```
{
  "result": {
    "state": "81ecbea3-6a4c-49a3-abe3-853b90cb18cb",
    "stdout": "",
    "stderr": "",
    "answer": []
  },
  "jsonrpc": "2.0",
  "id": 1
}
```
```

pnwamk:
https://github.com/GaloisInc/argo/pull/165 has been merged - of note:
+ the `TLS_ENABLE` env var can be set to a nonzero, nonempty string value to enable TLS for the argo server run in `http` mode
+ The underlying argo python client now has a [verify](https://github.com/GaloisInc/argo/blob/cdf2d30effbd18b7b082352abb30379ccf8c2de0/python/argo_client/connection.py#L284) keyword argument that can be used to disable certificate verification for TLS connections.
+ argo now depends on `scotty-tls` as a submodule until `v0.4.2` is published to hackage (do we need it to be a submodule for `cryptol` as well...?)

Issue #1206, opened 2021-06-01
cryptol-remote-api: unify manual and automatically generated server docs


RyanGlScott:
Currently, the `cryptol-remote-api` documentation in `cryptol-remote-api/docs/Cryptol.rst` are written by hand. We should instead strive to make the documentation completely autogenerated by using `cryptol-remote-api doc`. This way, the documentation will stay in sync with future additions to `cryptol-remote-api`.

This is the `cryptol-remote-api` equivalent of [this issue](https://github.com/GaloisInc/saw-script/issues/1121).

pnwamk:
This may have already been done - I'll look into this and confirm.

Issue #1196, opened 2021-05-14
Not in scope: type constructor or class `Doc.DescribedParams'


weaversa:
I'm attempting to build `cryptol-remote-api` on CentOS 7 and am having trouble. The dockerfile is attached.
[cryptol-solvers-centos7-nightly.Dockerfile.txt](https://github.com/GaloisInc/cryptol/files/6476027/cryptol-solvers-centos7-nightly.Dockerfile.txt)
Here is the error:
Any help would be appreciated.

```
[ 3 of 15] Compiling CryptolServer.Options ( src/CryptolServer/Options.hs, /workdir/dist-newstyle/build/x86_64-linux/ghc-8.10.2/cryptol-remote-api-0.1.0.0/build/CryptolServer/Options.o, /workdir/dist-newstyle/build/x86_64-linux/ghc-8.10.2/cryptol-remote-api-0.1.0.0/build/CryptolServer/Options.dyn_o )

src/CryptolServer/Options.hs:57:26: error:
    Not in scope: type constructor or class `Doc.DescribedParams'
    Module `Argo.Doc' does not export `DescribedParams'.
   |
57 | instance forall params . Doc.DescribedParams params => Doc.DescribedParams (WithOptions params) where
   |                          ^^^^^^^^^^^^^^^^^^^

src/CryptolServer/Options.hs:57:56: error:
    Not in scope: type constructor or class `Doc.DescribedParams'
    Module `Argo.Doc' does not export `DescribedParams'.
   |
57 | instance forall params . Doc.DescribedParams params => Doc.DescribedParams (WithOptions params) where
   |                                                        ^^^^^^^^^^^^^^^^^^^
cabal: Failed to build cryptol-remote-api-0.1.0.0 (which is required by
exe:cryptol-remote-api from cryptol-remote-api-0.1.0.0 and
exe:cryptol-eval-server from cryptol-remote-api-0.1.0.0).

The command '/bin/sh -c ./cry build cryptol-remote-api' returned a non-zero code: 1
```

pnwamk:
How odd. Is it possible the argo submodule is out of date? That would be my
guess from the error contents anyway.

I'm messing around with the Dockerfile myself now... will report if I
figure something out.

pnwamk:
Aha! the `--remote` flag on the line in the Dockerfile
```
RUN git clone https://github.com/GaloisInc/cryptol.git . && git submodule update --init --remote
```
[...]

PR #1207, opened 2021-06-01
Bump argo submodule to include docs for method results

RyanGlScott:
This follows up on a promise I made in https://github.com/GaloisInc/cryptol/issues/1196#issuecomment-841179362 to bump the `argo` submodule to include `argo` commit https://github.com/GaloisInc/argo/commit/b9b3edd992e57ca31224530f763f313fcf24bab7, which adds the ability to document the results of methods (in addition to their parameters). In addition to dealing with the breaking API changes from that commit, this adds missing documentation for each method's results.

Along the way, I realized that `cryptol-remote-api`'s documentation is hand-written rather than autogenerated (see #1206). In pursuit of eventually fixing that issue, the CI now makes sure that any changes to the autogenerated documentation are checked in.


cryptol-remote-api/check_docs.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+cd $DIR/docs
+
+export CRYPTOL_SERVER=$(which cryptol-remote-api)
```
pnwamk, line 7: The other existing RPC script uses a [slightly different method](https://github.com/GaloisInc/cryptol/blob/1d8a2fdb527bc85b0aa7acbf8a0d819a26262184/cryptol-remote-api/run_rpc_tests.sh#L24) to find the server:
```
export CRYPTOL_SERVER=$(cabal v2-exec which cryptol-remote-api)
```
Should these be the same and/or more similar? (I'm just imagining one failing and the other not and that being kind of funny/frustrating.) I think personally I have usually relied on `cabal run` or `cabal exec`, but I certainly don't know what method for finding the executable is the "best".

RyanGlScott: I'll be honest—I cargo-culted much of this script from [`saw-script`'s counterpart](https://github.com/GaloisInc/saw-script/blob/5535b03531d0790cddb01d732be316c96838a8db/saw-remote-api/scripts/check_docs.sh#L9) without thinking too deeply about it. But yes, we should be consistent with the conventions already established in `cryptol`. Fixed.

RyanGlScott: And as luck would have it, that's exactly the change I needed to make anyway in order to fix CI :)

pnwamk (approved):
LGTM!

While perusing the docs I discovered some issues but I think they're probably just better addressed separately instead of holding up this PR.

In particular, some of `Argo`'s docs need updating: https://github.com/GaloisInc/argo/pull/166

And perhaps it's worth considering generating something more polished from the `.rst`: https://github.com/GaloisInc/cryptol/issues/1208

PR #1224, opened 2021-06-29
feat: support for all solver backends for cryptol server/client

pnwamk:
Makes available both SBV and What4 based solver backends in the RPC server and Python client.

Addresses https://github.com/GaloisInc/argo/issues/173

### DIFF
--- a/cryptol-remote-api/python/cryptol/solver.py
+++ b/cryptol-remote-api/python/cryptol/solver.py
@@ -55,6 +55,3 @@ W4_Z3: Solver        = Solver("w4-z3")
 W4_BOOLECTOR: Solver = Solver("w4-boolector")
 W4_OFFLINE: Solver   = Solver("w4-offline")
 W4_ABC: Solver       = Solver("w4-any")
-
-
-
--- a/cryptol-remote-api/python/cryptol/solver.py
+++ b/cryptol-remote-api/python/cryptol/solver.py
@@ -1,7 +1,35 @@
 """Cryptol solver-related definitions"""
 from typing import NewType
 
-Solver = NewType('Solver', str)
+from dataclasses import dataclass
+
+@dataclass
+class OfflineSmtQuery():
+  """An SMTLIB2 script -- produced when an `offline` prover is used."""
+  content: str
+
+class Solver():
+  """An SMT solver mode selectable for Cryptol. Compare with the `:set prover` options in
+  the Cryptol REPL."""
+  __name: str
+  __hash_consing: bool
+
+  def __init__(self, name:str, hash_consing:bool=True) -> None:
+    self.__name = name
+    self.__hash_consing = hash_consing
+
+  def name(self) -> str:
+    """Returns a string describing the solver setup which Cryptol will recognize -- i.e.,
+    see the options available via `:set prover = NAME`."""
+    return self.__name
+
+  def hash_consing(self) -> bool:
+    """Returns whether hash consing is enabled (if applicable)."""
+    return self.__hash_consing
+
+  def without_hash_consing(self) -> 'Solver':
+    """Returns an identical solver but with hash consing disabled (if applicable)."""
+    return Solver(name=self.__name, hash_consing=False)
 
 # Cryptol-supported SMT configurations/solvers
 # (see Cryptol.Symbolic.SBV Haskell module)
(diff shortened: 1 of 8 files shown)


cryptol-remote-api/src/CryptolServer/Sat.hs
```diff
@@ -124,25 +215,25 @@ instance ToJSON ProveSatResult where
                   | res <- xs
                   ]
                 ]
+  toJSON (OfflineSMTQuery smtlib) =
+    JSON.object [ "result" .= ("offline" :: Text)
```
RyanGlScott, line 219: Don't forget to update the `resultFieldDescription` with this new information.

cryptol-remote-api/python/cryptol/solver.py
```diff
@@ -25,5 +53,8 @@
 W4_YICES: Solver     = Solver("w4-yices")
 W4_Z3: Solver        = Solver("w4-z3")
 W4_BOOLECTOR: Solver = Solver("w4-boolector")
-W4_MATHSAT: Solver   = Solver("w4-offline")
+W4_OFFLINE: Solver   = Solver("w4-offline")
 W4_ABC: Solver       = Solver("w4-any")
+
```
RyanGlScott, line 58: Trailing whitespace

cryptol-remote-api/src/CryptolServer/Sat.hs
pnwamk: Thanks!

cryptol-remote-api/src/CryptolServer/Sat.hs
```diff
@@ -27,76 +32,162 @@ import Cryptol.ModuleSystem (checkExpr)
...
+        Right smtlib -> pure $ OfflineSMTQuery smtlib
+    Right w4Cfg -> do
+      smtlibRef <- liftIO $ newIORef ""
+      result <- liftModuleCmd $
+        W4.satProveOffline w4Cfg hConsing False cmd $ \f -> do
+          withRWTempFile "smtOutput.tmp" $ \h -> do
+            f h
+            hSeek h AbsoluteSeek 0
+            contents <- TIO.hGetContents h
+            writeIORef smtlibRef $ T.unpack contents
```
atomb, line 133: Maybe don't unpack and make the field of `OfflineSMTQuery` a `Text` value instead? The contents of these files can get large sometimes.

cryptol-remote-api/src/CryptolServer/Sat.hs
```diff
@@ -27,76 +32,162 @@ import Cryptol.ModuleSystem (checkExpr)
...
+        Nothing -> do
+          smtlib <- liftIO $ readIORef smtlibRef
+          pure $ OfflineSMTQuery smtlib
   where
-    satResult :: [(TValue, Expr, Value)] -> CryptolCommand [(JSONType, Expression)]
-    satResult es = traverse result es
+    withRWTempFile name k =
+      X.bracket
+        (do tmp <- getTemporaryDirectory
+            let esc c = if isAscii c && isAlphaNum c then c else '_'
```
atomb, line 143: Well done. We should have similar escaping mechanisms in other places.

cryptol-remote-api/src/CryptolServer/Sat.hs
```diff
@@ -27,76 +32,162 @@ import Cryptol.ModuleSystem (checkExpr)
...
+          me <- readBack t v
+          case me of
+            Nothing -> error $ "type is not convertable: " ++ (show t)
+            Just e -> pure (JSONType mempty (tValTy t), e)
 
 data ProveSatResult
   = Unsatisfiable
   | Invalid CounterExampleType [(JSONType, Expression)]
   | Satisfied [[(JSONType, Expression)]]
+  | OfflineSMTQuery String
```
atomb, line 190: See comment about `T.unpack` above.

cryptol-remote-api/src/CryptolServer/Sat.hs
```diff
@@ -124,25 +215,25 @@ instance ToJSON ProveSatResult where
                   | res <- xs
                   ]
                 ]
+  toJSON (OfflineSMTQuery smtlib) =
+    JSON.object [ "result" .= ("offline" :: Text)
+                , "query" .= (T.pack smtlib)
```
atomb, line 220: Won't need to do `T.pack` if the file contents are already `Text`.

atomb (changes requested):
I think the offline SMT-Lib contents should probably be `Text` instead of `String`, but otherwise this is great. Thanks for updating the docs along with the code. :)

cryptol-remote-api/src/CryptolServer/Sat.hs
pnwamk: I remembered I borrowed it from the REPL code -- we now import that to reuse it. We can import it wherever else we want.

pnwamk:
Comments addressed and passing all tests (the two failures are unrelated). Planning on merging soon.

RyanGlScott (approved):

PR #1227, opened 2021-06-29
Explicit stride

robdockins:
This PR adds new syntactic forms for enumerations with explicit stride values (CF #1087).

The currently-implemented syntax steals the identifiers `by` and `down` as new keywords; we should discuss if this is acceptable, or if new syntax should be designed.

### DIFF
--- a/lib/Cryptol.cry
+++ b/lib/Cryptol.cry
@@ -195,19 +195,66 @@ length _ = `n
 /**
  * A finite sequence counting up from 'first' to 'last'.
  *
- * '[a..b]' is syntactic sugar for 'fromTo`{first=a,last=b}'.
+ * '[x .. y]' is syntactic sugar for 'fromTo`{first=x,last=y}'.
  */
-primitive fromTo : {first, last, a} (fin last, last >= first,
-                                    Literal first a, Literal last a) =>
-                                    [1 + (last - first)]a
+primitive fromTo : {first, last, a}
+  (fin last, last >= first, Literal last a) =>
+  [1 + (last - first)]a
+
+/**
+ * A finite sequence counting up from 'first' to 'last' by 'stride'
+ *
+ * '[x .. y by n]` is syntactic sugar for 'fromToBy`{first=x,last=y,stride=n}'.
+ */
+primitive fromToBy : {first, last, stride, a}
+  (fin last, fin stride, stride >= 1, last >= first, Literal last a) =>
+  [1 + (last - first)/stride]a
 
 /**
  * A possibly infinite sequence counting up from 'first' up to (but not including) 'bound'.
  *
- * Note that if 'first' = 'bound' then the sequence will be empty.
+ * '[ x ..< y ]' is syntacic sugar for 'fromToLessThan`{first=x,bound=y}'.
+ *
+ * Note that if 'first' = 'bound' then the sequence will be empty.  If 'bound = inf'
+ * then the sequence will be infinite, and will eventually wrap around for bounded types.
  */
 primitive fromToLessThan :
-  {first, bound, a} (fin first, bound >= first, LiteralLessThan bound a) => [bound - first]a
+  {first, bound, a} (fin first, bound >= first, LiteralLessThan bound a) =>
+  [bound - first]a
+
+/**
+ * A finite sequence counting from 'first' up to (but not including) 'bound'
+ * by 'stride'.  Note that if 'first = bound' then the sequence will
+ * be empty.  If 'bound = inf' then the sequence will be infinite, and will
+ * eventually wrap around for bounded types.
+ *
+ * '[x ..< y by n]' is syntactic sugar for 'fromToByLessThan`{first=a,bound=b,stride=n}'.
+ */
+primitive fromToByLessThan : {first, bound, stride, a}
+  (fin first, fin stride, stride >= 1, bound >= first, LiteralLessThan bound a) =>
+  [(bound - first)/^stride]a
+
+/**
+ * A finite sequence counting from 'first' down to 'last' by 'stride'.
+ *
+ * '[x .. y down by n]` is syntactic sugar for 'fromToDownBy`{first=x,last=y,stride=n}'.
+ */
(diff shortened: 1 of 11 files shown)


robdockins:
Documentation updates are still needed, but I wanted to make sure we are happy with the syntax choices before doing that.

yav:
@robdockins the notation makes sense to me.   The only improvement I can think of is to clarify in the comment-docs for the "inclusive constructs (the non `<` ones) that the end point might not actually be in the final result (as in `[ 4 .. 1 down by 2]` is `[4,2]`, which didn't get to the 1)

src/Cryptol/TypeCheck/Solver/InfNat.hs
```diff
@@ -122,23 +122,22 @@ nMod (Nat x) (Nat y)  = Just (Nat (mod x y))
...
--- could be relaxed later.
+-- @msgLen <= blockSize * n@. It is undefined when @blockSize = 0@,
+-- or when @blockSize = inf@. @inf@ divided by any positive
+-- finite value is @inf@.
 nCeilDiv :: Nat' -> Nat' -> Maybe Nat'
+nCeilDiv _       Inf      = Nothing
 nCeilDiv _       (Nat 0)  = Nothing
-nCeilDiv Inf     _        = Nothing
-nCeilDiv (Nat _) Inf      = Nothing
+nCeilDiv Inf     (Nat _)  = Just Inf
```
yav, line 131: You probably need to update `lib/CryptolTC.z3` to match the new behavior

src/Cryptol/TypeCheck/Solver/InfNat.hs
```diff
@@ -122,23 +122,22 @@ nMod (Nat x) (Nat y)  = Just (Nat (mod x y))
...
--- It is also undefined when either input is infinite; perhaps this
--- could be relaxed later.
+-- @blockSize@ divides @msgLen + k@. It is undefined when @blockSize = 0@
+-- or @blockSize = inf@.  @inf@ modulus any positive finite value is @0@.
 nCeilMod :: Nat' -> Nat' -> Maybe Nat'
+nCeilMod _       Inf      = Nothing
 nCeilMod _       (Nat 0)  = Nothing
-nCeilMod Inf     _        = Nothing
-nCeilMod (Nat _) Inf      = Nothing
+nCeilMod Inf     (Nat _)  = Just (Nat 0)
```
yav, line 140: You probably need to update lib/CryptolTC.z3 to match the new behavior

src/Cryptol/TypeCheck/Solver/InfNat.hs
yav: Also, update `TypeCheck/Solver/Numeric/Fin.hs` as it is currently assume that the result is always `fin`

yav: Any (maybe) `src/Cryptol/TypeCheck/Solver/Numeric/Interval.hs`

src/Cryptol/TypeCheck/Solver/InfNat.hs
yav: Maybe we need to update `src/Cryptol/TypeCheck/Solver/Numeric/Interval.hs`

yav (changes requested):
I think there are a few more changes that we need for the solver, to match the more relaxed constraints on `/^` and `/%`.

Otherwise it looks good to me, with the caveat that I didn't go over the evaluation semantics in detail

src/Cryptol/TypeCheck/Solver/InfNat.hs
robdockins: Do you have a sense about how and if the interval arithmetic needs to change?

yav: Not sure, but I was thinking that since previously the answer was always finite, and now it can be infinite sometimes, we may have to tweak something to make a bit larger

robdockins: The interval computations fall back on `nCeilDiv` for the new cases that matter.  After staring at the interval code for awhile, I'm pretty confident they don't need changing.

robdockins:
AFAICT, the remaining failures are not related to this PR.

robdockins:
@yav, @brianhuffman, any additional thoughts about this?

lib/Cryptol.cry
```diff
@@ -195,20 +195,70 @@ length _ = `n
...
+ *
+ * '[x .. y by n]` is syntactic sugar for 'fromToBy`{first=x,last=y,stride=n}'.
+ */
+primitive fromToBy : {first, last, stride, a}
+  (fin last, fin stride, stride >= 1, last >= first, Literal last a) =>
+  [1 + (last - first)/stride]a
+
+/**
+ * A finite sequence counting from 'first' up to (but not including) 'bound'
+ * by 'stride'.  Note that if 'first = bound' then the sequence will
```
yav, line 229: Should this be `first + stride <= bound`?

lib/Cryptol.cry
```diff
@@ -195,20 +195,70 @@ length _ = `n
...
+  [1 + (first - last)/stride]a
+
+/**
+ * A finite sequence counting from 'first' down to (but not including)
+ * 'bound' by 'stride'.
+ *
+ * '[x ..> y down by n]` is syntactic sugar for
+ * 'fromToDownByGreaterThan`{first=x,bound=y,stride=n}'.
+ *
+ * Note that if 'first = bound' the sequence will be empty.
```
yav, line 257: Same as above, maybe we should mention the stride?

lib/Cryptol.cry
robdockins: I'm not sure what you mean... `first` will definitely be included in the enumeration as long as `first < bound`, regardless of the stride.

yav (approved):
Looks good to me.

lib/Cryptol.cry
yav: Oh, right, I was confused, never mind.

PR #1228, opened 2021-07-06
Build distribution tarballs with solvers

atomb:


.github/ci.sh
```diff
@@ -163,6 +163,13 @@ zip_dist() {
   name="${name:-"cryptol-$VERSION-$RUNNER_OS-x86_64"}"
   cp -r dist "$name"
   tar -cvzf "$name".tar.gz "$name"
+  sname="${name}-with-solvers"
```
RyanGlScott, line 166: For the sake of separating concerns, I think it would be better to split `zip_dist` into two functions, one for creating the vanilla bindist and another for creating the `with-solvers` bindist. This has the benefit of making it easier to create and test each bindist locally without having to create the other, and it more closely mirrors the structure of `ci.yml`, which signs and uploads each bindist separately.

RyanGlScott (approved):

.github/ci.sh
atomb: Yeah, that's a good thought. I'll change it.

PR #1230, opened 2021-07-07
Attempt to fix CVC4 and Docker upload issues

atomb:
We will need a better solution to getting CVC4 eventually, but unbreaking the builds, even temporarily, is high priority. And the right solution for CVC4, with the recent transition to CVC5, is somewhat tricky.

I believe the one test failure is an allowed failure and can be ignored for now.


kquick (approved):
I agree: this should re-enable useful build results and we can address CVC updates later.

PR #1231, opened 2021-07-07
feat: portable rpc server builds use fork/threading less

pnwamk:


cryptol-remote-api/server-not-threaded.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/cryptol-remote-api/cryptol-remote-api.cabal b/cryptol-remote-api/cryptol-remote-api.cabal
```
RyanGlScott, line 1: Do you need this `.patch` file if the use of `-threaded` is controlled by a `cabal` flag?

pnwamk: whoops - thanks! =)

atomb (approved):
These changes look good to me. Thanks for looking into the issue!

PR #1232, opened 2021-07-08
Attempt to fix allowed test failure on Windows

atomb:


kquick (approved):
It's not clear what's wrong with the windows CI builds, but it also doesn't seem to be related to the code or the CI configs, so I agree that disabling them for now to get green CI is more useful.

Looks like the actual test failure is probably a transient network issue.  

I'm not sure why the windows test is marked as "skipped" as opposed  to just being gone; I'm slightly concerned that will result in yellow status instead of a green checkmark but maybe it's just a sync comparison between this PR and master that will go away when the merge happens.

atomb:
Yeah, I just re-started the jobs in the hop that the failed test will work this time. 🤞

PR #1246, opened 2021-07-22
Temporarily disable uploading artifacts for Windows.

yav:
This seems to break the CI, we need to investigate why

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,13 +181,15 @@ jobs:
           path: bin
           name: ${{ runner.os }}-bin
 
-      - uses: actions/upload-artifact@v2
-        if: runner.os == 'Windows'
-        with:
-          name: ${{ env.NAME }}
-          path: "cryptol.msi*"
-          if-no-files-found: error
-          retention-days: ${{ needs.config.outputs.retention-days }}
+# This is commented out because it appears to fail uploading
+# and the entire build fails.  XXX: investigate why.
+#      - uses: actions/upload-artifact@v2
+#        if: runner.os == 'Windows'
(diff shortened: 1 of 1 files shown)


atomb:
Although disabling Windows temporarily seems reasonable (though I'd like it if we can resolve it soon), the portable Docker container is something we're critically depending on. I believe the failure you saw was spurious, and would be resolved by restarting the build, given that it's been succeeding in other runs. Would you be up for reverting that part of the PR?

yav:
@atomb I am fine with re-enabling it after it is fixed.  I don't think building *ghc* on every commit is useful, as the task takes hours to finish, and we should not be doing that on every single commit.

atomb:
Just to double-check: by "fixed" do you mean "completing at all" or "completing quickly"? I ask because I think the former is already true: it should work if we just re-run it (though if that's not true, we should certainly address that).

I think it could be reasonable to tweak that workflow to run only when requested (`workflow_dispatch`), or perhaps nightly, instead of on every PR. It would also be a major improvement to cache the results of the GHC build and reuse them, rather than building it every time, but I'd prefer to have it re-enabled (at least on-demand) and make that improvement to the performance of the build afterward.

yav:
@atomb if you know how to make it cache things, or schedule it to run nightly, please do enable it.  My goal is to have the CI detect errors in commits, reliably in under 1 hour.    I don't think that the Cryptol CI is the right place to be building GHC though, at least not on every commit.

atomb:
I'd forgotten that I already had a branch set up to fix the Windows build. I'll extend it to make the portable Docker build run only on nightly builds.

PR #1247, opened 2021-07-22
Update Syntax.md

weaversa:
> changed to <

### DIFF
--- a/docs/Syntax.md
+++ b/docs/Syntax.md
@@ -444,7 +444,7 @@ an infinite stream of bits.
     [t1 .. t2 by n]       // Enumeration (stride)
     [t1 .. <t2 by n]      // Enumeration (stride, ex. bound)
     [t1 .. t2 down by n]  // Enumeration (downward stride)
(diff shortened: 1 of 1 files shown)


docs/Syntax.md
```diff
@@ -444,7 +444,7 @@ an infinite stream of bits.
     [t1 .. t2 by n]       // Enumeration (stride)
     [t1 .. <t2 by n]      // Enumeration (stride, ex. bound)
     [t1 .. t2 down by n]  // Enumeration (downward stride)
-    [t1 .. >t2 down by n] // Enumeration (downward stride, ex. bound)
+    [t1 .. <t2 down by n] // Enumeration (downward stride, ex. bound)
```
yav, line 447: I think the downward one is supposed to be using `>`?  As in:
```
f = [ 3 .. > 1 down by 1]

[3, 2]
```

weaversa: Yes, you're right. I tried this on a docker nightly version of Cryptol from 2 weeks ago that did not yet have this feature and the interpreter complained so I though it was wrong. After updating to the nightly just now I see that this works.

Issue #1250, opened 2021-07-22
Cryptol RPC server sometimes hangs


atomb:
We've seen mention of cases where the Cryptol RPC server can hang when loading Cryptol modules, but haven't been able to reproduce it locally yet.

@pnwamk, do you have some more detail you can fill in?

pnwamk:
As it was described to me, the client hangs on `c.load_module()` after having successfully loading _many_ previous modules in a script such as the following while the server is running in a docker container:

```
import cryptol
import os

# Steps to clone example repo and set up server:

# git clone https://github.com/GaloisInc/cryptol-specs.git
# git checkout rpc-load-module-test

# docker run --rm -ti \
# -v $PWD/Common:/cryptol/Common \
# -v $PWD/Primitive:/cryptol/Primitive \
# -v $PWD/McEliece_KEM:/cryptol/McEliece_KEM \
# -p 8080:8080 \
# -e CRYPTOLPATH='/cryptol' \
# ghcr.io/galoisinc/cryptol-remote-api:nightly


# After server is running, simply execute this script, e.g.:
# python3 load.py

c = cryptol.connect(url="http://localhost:8080", reset_server=True)

for root, dir, files in os.walk("."):
    for name in files:
        if name.endswith(".cry"):
            new_file = os.path.join(root, name)[2:-4].replace("/", "::")
            print("Loading ", new_file, flush=True)
            c.load_module(new_file).result() # added .result()
            c.reset_server()
```

N.B., the above script and it's accompanying instructions don't _quite_ work as is, as the [cryptol-specs](https://github.com/galoisinc/cryptol-specs) repository has some modules that fail to load at present.

pnwamk:
This may be reproducing the error we've had reported. At least observationally it appears similar:

```
# Steps to clone example repo and set up server:

# git clone https://github.com/GaloisInc/BLST-Verification.git
# cd BLST-Verification
# git submodule update --init

# docker run --rm -ti \
# -v $PWD/spec:/cryptol/spec \
# -v $PWD/cryptol-specs:/cryptol/cryptol-specs \
# -p 8080:8080 \
# -e CRYPTOLPATH='/cryptol/spec:/cryptol/cryptol-specs' \
# ghcr.io/galoisinc/cryptol-remote-api:nightly


# After server is running, simply execute the following script from the `spec` directory, e.g.:
# cd spec
[...]

PR #1258, opened 2021-07-28
fix(rpc): reset TC solver on file/module load

pnwamk:
Some difficult to reproduce bugs regarding the RPC server hanging
(https://github.com/GaloisInc/cryptol/issues/1250) seemed to be
related to the type checker solver (Z3) hanging after periods of
extended use. I.e., a module would load/error quickly normally,
but if preceded by enough prior solver/server activity, Z3
would consistently hang. To avoid this, we can simply reset the
solver and reload the type checker prelude when the server is asked
to load a file or module. This solves the Z3 hang issues in observed
cases and requires no additional forking (which we're trying to
minimize when possible).

### DIFF
--- a/cryptol-remote-api/src/CryptolServer.hs
+++ b/cryptol-remote-api/src/CryptolServer.hs
@@ -85,6 +85,9 @@ modifyModuleEnv f =
 getTCSolver :: CryptolCommand SMT.Solver
 getTCSolver = CryptolCommand $ const $ view tcSolver <$> Argo.getState
 
+getTCSolverConfig :: CryptolCommand SMT.SolverConfig
+getTCSolverConfig = CryptolCommand $ const $ solverConfig <$> Argo.getState
+
 liftModuleCmd :: ModuleCmd a -> CryptolCommand a
 liftModuleCmd cmd =
     do Options callStacks evOpts <- getOptions
@@ -123,6 +126,7 @@ data ServerState =
   ServerState { _loadedModule :: Maybe LoadedModule
               , _moduleEnv :: ModuleEnv
               , _tcSolver :: SMT.Solver
+              , solverConfig :: SMT.SolverConfig
               }
 
 loadedModule :: Lens' ServerState (Maybe LoadedModule)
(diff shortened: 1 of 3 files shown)


robdockins:
Interesting... do you find that the Z3 version matters?

Either way, I wonder if this is worth trying to reduce to an issue we can file upstream?

pnwamk:
It was with Z3 version 4.8.8 inside the docker container produced by [this Dockerfile](https://github.com/GaloisInc/cryptol/blob/1b52a351f7f1b549a6fac0fb1beb63dac7af617a/cryptol-remote-api/Dockerfile). The steps in this issue reproduce it: https://github.com/GaloisInc/cryptol/issues/1250#issuecomment-885782292

I can try bumping the version of Z3 in the Dockerfile to see if that impacts things and we can go from there w.r.t. a future issue.

robdockins:
Yeah, I'd see if, e.g., Z3 4.8.11 makes a difference.

weaversa:
We tried 4.8.13 today and it resolved our observable issues. It's possible that upgrading just stalls the problem. That is, imagine that the problem is that some series of push/pop is leaving too much in the z3 state, and when z3 is presented with a new type-checking problem, it becomes bogged down by previous state. Perhaps newer versions of z3 are simply faster, but may get bogged down in the future as the state continues to grow. The reset problem would fix any such leakage, if it exists.

Of course, it could simply be a bug in and old z3 and the new version just fixes it. I don't know.

robdockins:
Interesting. I know they have been doing a lot of work on a new real/integer arithmetic solver over the last several versions, so that might be related.

I feel like it's a bit of a shame to reset the solver so often because then we have to reload the type-checking prelude... but if it prevents problems, perhaps it is worth it.

pnwamk:
Chatted briefly with @atomb and I think these changes all things considered are worth merging: we're still miles ahead of where we used to be---i.e., we used to reinvoke the solver from scratch constantly and now use a single instance, [thanks](https://github.com/GaloisInc/cryptol/commit/6626c4d33e8a9470845c8613a83bbc91f16eae4a#diff-11c812f40d1c42d97abec1a0b704134767926169723f0952a06798bb9b63485f) @robdockins!---, and so while this is a slight step back in performance for some queries, it seems unlikely to be a bottle neck given that we're just doing it when a _new_ module/file is loaded explicitly by an RPC client, and that's going to potentially invoke quite a bit of cryptol/server computation anyway.

PR #1267, opened 2021-08-20
Demote compare

robdockins:
Add primitives for computing value-level results from comparing number types.


lib/Cryptol.cry
```diff
@@ -192,6 +192,24 @@ demote = number`{val}
...
+ */
+primitive compareLt : { m : #, n : # } Bit
+
+/**
+ * Compare the given numeric types, returning
+ * 'True' when 'm <= n' and 'False' otherwise.
+ */
+primitive compareLeq : { m : #, n : # } Bit
+
 /**
```
yav, line 213: In the interest of keeping the number of primitives down, I wonder if it might be better to add a single `demoteNumeric` primitve?   Perhaps something like this:
```
demoteNumeric : { m : # } => (Bool,Integer)
```
The first component indicates if the number is infinite, so finite numbers would be of the form `(false,n)` and infinity would be `(true,0)`.   I guess it would be prettier to expose `InfNat`, but I am not sure it is worth the extra work.

weaversa: I was envisioning a `demoteConstraint` that takes logic about `inf` into account.

weaversa: Also, I was able to get `isPrefix` to work, so perhaps we only need a narrow solution for cases like `coerceSize`?

```
isPrefix : {m, n, a} (fin n, Eq a) => [m]a -> [n]a -> Bit
isPrefix xs x =
    if `(min n m) < `n then False
    else and [ xi == xsi | xi <- x | xsi <- xs ]
```

robdockins: Oh, that trick with `min` is interesting.  We can probably reduce everything we want to do down to just the `compareEq` primitive with clever uses of `min` and `max`.  Or, Iavor's idea would also work.

yav: Hm, `demoteConstraint` as a primitive seems tricky, as it is not clear what would be its type.  I think for now we should go with a simpler solution, and the "proper" fix would be to add a GADT style language construct for examining numeric types.

yav:
Closing this for now, seems we can do a bunch of this stuff with prop guards now.

PR #1268, opened 2021-08-20
Add arrayCopy, arraySet, arrayRangeEqual array primitives.

andreistefanescu:
Add arrayRangeLookup and arrayRangeUpdate helpers.

### DIFF
--- a/lib/Array.cry
+++ b/lib/Array.cry
@@ -11,3 +11,47 @@ primitive arrayConstant : {a, b} b -> (Array a b)
 primitive arrayLookup : {a, b} (Array a b) -> a -> b
 primitive arrayUpdate : {a, b} (Array a b) -> a -> b -> (Array a b)
 
+/**
+ * Copy elements from the source array to the destination array.
+ *
+ * 'arrayCopy dest_arr dest_idx src_arr src_idx len' copies the
+ * elements from 'src_arr' at indices '[src_idx .. (src_idx + len)]' into
+ * 'dest_arr' at indices '[dest_idx .. (dest_idx + len)]'.
+ *
+ * The result is undefined if either 'dest_idx + len' or 'src_idx + len'
+ * wraps around.
+ */
+primitive arrayCopy : {n, a} (Array [n] a) -> [n] -> (Array [n] a) -> [n] -> [n] -> (Array [n] a)
+/**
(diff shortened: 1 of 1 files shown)


lib/Array.cry
```diff
@@ -11,3 +11,47 @@ primitive arrayConstant : {a, b} b -> (Array a b)
...
+
+arrayRangeLookup : {a, b, n} (Integral a, fin n, n >= 1, LiteralLessThan n a) => (Array a b) -> a -> [n]b
+arrayRangeLookup arr idx = res
+  where
+    res @ i = arrayLookup arr (idx + i)
+
+arrayRangeUpdate : {a, b, n} (Integral a, fin n, n >= 1, Literal (n - 1) a) => (Array a b) -> a -> [n]b -> (Array a b)
+arrayRangeUpdate arr idx vals = arrs ! 0
+  where
+    arrs = [arr] # [ arrayUpdate acc (idx + i) val | acc <- arrs | i <- [0 .. (n - 1)] | val <- vals ]
```
robdockins, line 56: It's a small thing, but I think you could avoid the `n >= 1` constraint if you used the `i <- [ 0 ..< n ]` enumeration form instead.

I think the `n >= 1` constraint is also unnecessary for `arrayRangeLookup`.

lib/Array.cry
```diff
@@ -11,3 +11,47 @@ primitive arrayConstant : {a, b} b -> (Array a b)
 primitive arrayLookup : {a, b} (Array a b) -> a -> b
 primitive arrayUpdate : {a, b} (Array a b) -> a -> b -> (Array a b)
 
+/**
+ * Copy elements from the source array to the destination array.
+ *
+ * 'arrayCopy dest_arr dest_idx src_arr src_idx len' copies the
+ * elements from 'src_arr' at indices '[src_idx .. (src_idx + len)]' into
```
robdockins, line 18: Sorry to nitpick some more, but I think essentially all these ranges are intended to be half-open and we should probably use the Cryptol-style notation for that, which would make this `[src_idx ..< (src_idx + len)]`, and similar below.

robdockins (approved):
Let's fix up the failing regression test case and address the comment nitpicks. Then I think this is good to go.

PR #1282, opened 2021-09-11
feat: timeout for cryptol python client methods

pnwamk:


pnwamk:
This PR should probably be rebased one https://github.com/GaloisInc/cryptol/pull/1272 is merged and the timeouts can be added to that cleaner interface as well.

pnwamk:
Also, if using the single implicit connection interface (https://github.com/GaloisInc/cryptol/pull/1285), it would be very natural to add an interrupt handler to send an interrupt notification to the server upon receiving a SIGINT interrupt from the user (i.e., `ctl+c`).

PR #1284, opened 2021-09-14
Use prebuilt solver binaries from what4-solvers

atomb:


kquick (approved):
Looks good!

.github/ci.sh
```diff
@@ -9,6 +9,14 @@ mkdir -p "$BIN"
 
 is_exe() { [[ -x "$1/$2$EXT" ]] || command -v "$2" > /dev/null 2>&1; }
 
+deps() {
+  case "$RUNNER_OS" in
+    Linux) ldd $1 || true ;;
+    macOS) otool -L $1 || true ;;
+    Windows) ldd $1 || true ;;
```
kquick, line 16: What's the intent  of the `|| true` for these?  It will always allow these to succeed, so is this just for getting logging output to show which binary is running and the validity of  the available shared libs?

atomb: At least `ldd` (and maybe `otool`?) return non-zero for statically-linked executables, and some of the executables we generate are expected to be statically linked. And, ultimately, yeah, the purpose of this is to help with debugging.

PR #1285, opened 2021-09-14
[RPC] Single-connection interface

m-yac:
This PR adds a single-connection interface for the python Cryptol RPC bindings based on the existing synchronous, typed interface. Since the argo server's default setting is to only allow a single connection at a time, this interface is a better fit for the expected use case of these bindings. The existing interfaces, where multiple `Connection` objects can be created, still exist and are unchanged.

Compared to the synchronous interface, this change lets us get rid of the "`c.`"s in front of every command. Compared to the old interface, this change also lets us get rid of the `.result()` calls and adds more types. For example, what was before:
```python
c.call("f", x).result()   # type: Any
c.prove("prop").result()  # type: Any
```
is now:
```python
call("f", x)   # type: CryptolValue
prove("prop")  # type: Union[Qed, Counterexample]
```

Link for comparing this interface to the old interface: https://github.com/galoisInc/cryptol/compare/5a668a4594e386a084f6f0ceb231c69946971a50...rpc/single-conn-interface (e.g. check out the test cases)


atomb (approved):
I think this looks really nice, and should make simple scripts less noisy. The one CI failure is, I believe, and issue with the CVC4 binaries we're using. I'm looking into fixing it separately.

atomb:
Perhaps try rebasing/merging with `master` and see if CI succeeds now. I think it should.

pnwamk:
This PR should probably bump the client version and document what is added briefly in the CHANGELOG.md file.

m-yac:
This should be ready to merge, the RPC bits of the CI have passed.

The changes made in this latest batch of commits are:
- Update the synchronous and single-connection interfaces with `timeout` parameters
- Add a `timeout` property to the synchronous interface to match the `timeout` member field in the old interface
- Add `get_default_timeout`/`set_default_timeout` methods to all three interfaces (since the single-connection interface can't have a `timeout` member field or property)
- Update the changelog and bump the version number to 2.11.7

cryptol-remote-api/python/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
 
+## 2.11.7 -- 2021-09-22
+
+* Add a synchronous, type-annotated interface (`synchronous.py`). To use this
+  interface, connect using `c = cryptol.sync.connect()` and remove all
+  `.result()` calls.
+* Add a single-connection, synchronous, type-annotated interface based on the
+  above. To use this interface, add `from cryptol.single_connection import *`,
+  connect using `connect()`, replace `c.eval(...).result()` with
+  `eval_cry(...)`, remove all `c.` prefixes, and remove all `.result()` calls.
```
pnwamk, line 11: `eval_cry` => `cry_eval` right?

pnwamk (approved):

m-yac:
The CI passed for the second most recent commit, and the most recent commit just fixes a typo in `CHANGELOG.md` - so I'm going to merge this

PR #1288, opened 2021-09-21
Update test_error_recovery.py

pnwamk:
disable certificate verification since test suite uses
self signed certs


atomb (approved):
The simplicity of this makes me happy. :)

PR #1303, opened 2021-11-12
Primitive scanl

robdockins:
This PR changes `scanl` to a primitive instead of a defined value in the Prelude.  In addition, a few other definitions in the Prelude were reimplemented to use `foldl` or `scanl` instead of being directly recursive.

These changes are primarily intended to support the Cryptol->Coq extraction use case in SAW.  That extraction currently has no way to translate recursive Cryptol programs, so it is useful to avoid them for basic functionality exposed via the Prelude.  With this PR the `sortBy` function is the only remaining recursive definition in the Prelude.

### DIFF
--- a/lib/Cryptol.cry
+++ b/lib/Cryptol.cry
@@ -1157,27 +1157,22 @@ foldr' f acc xs = foldl' g acc (reverse xs)
 sum : {n, a} (fin n, Eq a, Ring a) => [n]a -> a
 sum xs = foldl' (+) (fromInteger 0) xs
 
-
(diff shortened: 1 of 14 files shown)


brianhuffman (approved):
It would be good to have something in the test suite that checks, for both concrete and symbolic evaluation, that the implementations of `foldl` and `scanl` match the behavior specified by their original definitions. (Actually, we should have such tests for *all* primitives—see #48—but we can start by making tests for new primitives as we add them.)

PR #1307, opened 2021-11-20
[RPC] f-string-like Cryptol quasiquotation

m-yac:
This PR adds the function `cry_f` for quasiquoting cryptol using an f-string-like syntax. The example from GaloisInc/saw-script#1188 can now be written as:
```python
a = cry('g x')
b = cry_f('foo {a}')
```
This PR also adds `cry_eval_f` and `Connection.eval_f` which behave the same as composing `cry_eval` (or `Connection.eval`) with `cry_f`. See `tests/cryptol/test_quoting.py` for more examples.

Under the hood, `cry_f` uses Python's f-string parser and converts all of the bracketed arguments it is given into Cryptol syntax. This enables it to handle more complex expressions, like:
```python
cry_f('{ {"x": BV(size=8, value=1)} }')
cry_f('{{x = {BV(size=8, value=1)}}}')
```
both of which produce `cry('{x = 0x01}')`.

In making this PR I realized `cryptoltypes.py` needs a lot of cleaning up. I pulled out most of the changes I made to that file along the way and will be making a separate PR for that soon.

### DIFF
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Revision history for `cryptol` Python package
 
-## 2.12.2 -- YYYY-MM-DD
+## 2.12.2 -- 2021-11-19
 
-* NEW CHANGELOG ENTRIES SINCE 2.12.0 GO HERE
+* Add an interface for Cryptol quasiquotation using an f-string-like syntax,
+  see `tests/cryptol/test_quoting` for some examples.
 
 ## 2.12.0 -- 2021-11-19
 
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Revision history for `cryptol` Python package
 
-## 2.12.2 -- 2021-11-19
+## 2.12.2 -- YYYY-MM-DD
 
 * Add an interface for Cryptol quasiquotation using an f-string-like syntax,
   see `tests/cryptol/test_quoting` for some examples.
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -46,11 +46,30 @@ def cry_f(s : str) -> CryptolLiteral:
     """Embed a string of cryptol syntax as ``CryptolCode``, where the given
        string is parsed as an f-string, and the values within brackets are
        converted to cryptol syntax using ``to_cryptol_str``.
-       
-       Example:
-           >>> x = BV(size=7, value=1)
-           >>> y = cry_f('fun1 {x}')
-           >>> cry_f('fun2 {y}')
-           'fun2 (fun1 (1 : [7]))'
+
+       Like f-strings, values in brackets (``{``, ``}``) are parsed as python
+       expressions in the caller's context of local and global variables, and
+       to include a literal bracket in the final string, it must be doubled
+       (i.e. ``{{`` or ``}}``). The latter is needed when using explicit type
+       application or record syntax. For example, if ``x = [0,1]`` then
+       ``cry_f('length `{{2}} {x}')`` is equal to ``cry('length `{2} [0,1]')``
+       and ``cry_f('{{ x = {x} }}')`` is equal to ``cry('{ x = [0,1] }')``.
+
+       When formatting Cryptol, it is recomended to always use this function
+       and never any of python's built-in methods of string formatting (e.g.
+       f-strings, ``str.format``) as the latter will not always produce valid
+       cryptol syntax. Specifically, this function differs from these methods
+       in the cases of ``BV``s, string literals, function application (this
+       function will add parentheses as needed), and dicts. For example,
+       ``cry_f('{ {"x": 5, "y": 4} }')`` equals ``cry('{x = 5, y = 4}')``
+       but ``f'{ {"x": 5, "y": 4} }'`` equals ``'{"x": 5, "y": 4}'``. Only
+       the former is valid cryptol syntax for a record.
+
+       :example:
+
+       >>> x = BV(size=7, value=1)
+       >>> y = cry_f('fun1 {x}')
+       >>> cry_f('fun2 {y}')
+       'fun2 (fun1 (1 : [7]))'
     """
     return CryptolLiteral(func_customf(s, to_cryptol_str, frames=1, filename="<cry_f>"))
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -1,0 +1,56 @@
+"""Cryptol quasiquotation using an f-string-like syntax"""
+
+from typing import Any, Union
+
+from .bitvector import BV
+from .opaque import OpaqueValue
+from .commands import CryptolValue
+from .cryptoltypes import CryptolLiteral
+from .custom_fstring import *
+
+def to_cryptol_str(val : Union[CryptolValue, str, CryptolLiteral]) -> str:
+    """Converts a ``CryptolValue``, string literal, or ``CryptolLiteral`` into
+       a string of cryptol syntax."""
+    if isinstance(val, bool):
+        return 'True' if val else 'False'
+    elif isinstance(val, tuple):
+        return '(' + ', '.join(to_cryptol_str(x) for x in val) + ')'
+    elif isinstance(val, dict):
+        return '{' + ', '.join(f'{k} = {to_cryptol_str(v)}' for k,v in val.items()) + '}'
+    elif isinstance(val, int):
+        return str(val)
+    elif isinstance(val, list):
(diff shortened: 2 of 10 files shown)


RyanGlScott:
Fantastic work, @m-yac! I feel like my ability to review this will be hampered by my limited understanding of the underlying Python tricks, but here are some questions/comments I have after skimming through the patch:

* If I understand correctly, the main advantage of using `cry_f`/`eval_f` over Python's usual string formatting is that the former will automatically parenthesize quasiquoted arguments. If so, where is that bit implemented? Are there other advantages to this approach that I've overlooked?
* Speaking of which, it seems like combining Python's default string formatting with Cryptol should be discouraged, given the perils. It would be worth calling out this point a bit more loudly in the documentation, I think, as that is a footgun that is easy to overlook. I'm not sure _where_ exactly would be the best place to call this out, but somewhere.
* Since we are in the business of creating our own quasiquoter, what do you think about picking a different syntax from `{...}`? The reason I bring this up is because Cryptol already uses braces to denote explicit type applications (e.g., ``length`{42}``), so Python users will have to remember to type ``length`{{42}}`` (with double braces) to avoid confusion with the quasiquoting syntax.

  I'm not sure how difficult this would be to implement. If it proves impractical to do so, then we should at least document this as another potential pitfall.

m-yac:
Thanks @RyanGlScott! These is super helpful questions and pieces of feedback.

> * If I understand correctly, the main advantage of using `cry_f`/`eval_f` over Python's usual string formatting is that the former will automatically parenthesize quasiquoted arguments. If so, where is that bit implemented? Are there other advantages to this approach that I've overlooked?

Each of the quoted arguments in a `cry_f` are passed to [`to_cryptol_str`](https://github.com/GaloisInc/cryptol/blob/f46dfa5937d4d3fd3de9664378dc31b460e1b9ad/cryptol-remote-api/python/cryptol/quoting.py#L11) to convert them to strings, then after all the strings are put together, [the result is wrapped in a `CryptolLiteral`](https://github.com/GaloisInc/cryptol/blob/f46dfa5937d4d3fd3de9664378dc31b460e1b9ad/cryptol-remote-api/python/cryptol/quoting.py#L56).

So, for your first question, the parentheses are added in the [`CryptolLiteral` case of `to_cryptol_str`](https://github.com/GaloisInc/cryptol/blob/f46dfa5937d4d3fd3de9664378dc31b460e1b9ad/cryptol-remote-api/python/cryptol/quoting.py#L37).

For your second question, the other big advantage is that `to_cryptol_str` will also ensure other python values are represented as valid Cryptol syntax. Specifically dicts, BVs, and strings do not give valid Cryptol syntax when passed to `str()`, so they have special cases in `to_cryptol_str`. For example:
```python
>>> f'{ {"x": BV(size=8, value=1)} }' == '{"x": BV(8, 0x01)}' # not valid Cryptol syntax
True
>>> cry_f('{ {"x": BV(size=8, value=1)} }') == cry('{x = 0x01}') # valid Cryptol syntax
True
```

> * Speaking of which, it seems like combining Python's default string formatting with Cryptol should be discouraged, given the perils. It would be worth calling out this point a bit more loudly in the documentation, I think, as that is a footgun that is easy to overlook. I'm not sure _where_ exactly would be the best place to call this out, but somewhere.

Ah, by this do you mean specifically saying to always using `cry_f`-strings and never f-strings when formatting Cryptol? In which case I agree. As for where to put it, definitely in the docstring for `cry_f`, but perhaps also it should also go in the README? If this feature is going to end up getting used a lot it might be worth highlighting it and discussing pitfalls.

> * Since we are in the business of creating our own quasiquoter, what do you think about picking a different syntax from `{...}`? The reason I bring this up is because Cryptol already uses braces to denote explicit type applications (e.g., `` length`{42} ``), so Python users will have to remember to type `` length`{{42}} `` (with double braces) to avoid confusion with the quasiquoting syntax.
>
>   I'm not sure how difficult this would be to implement. If it proves impractical to do so, then we should at least document this as another potential pitfall.

Dang it, good point! Unfortunately, since I'm using reflection to make this work, I can't think of an easy way to change what characters delimit f-string arguments. Perhaps we could do a find-and-replace swapping braces and some other pair of characters before the reflection, then swap again after? But what character pair would we use to replace braces? – square brackets and parentheses have even more conflicts in cryptol syntax.

Regardless, I agree that this should be mentioned as a pitfall. If we're assuming our users are already fairly familiar with Python, then they should be used to adding double-braces in the context of f-strings, but I have a feeling that's not a valid assumption.

RyanGlScott:
> So, for your first question, the parentheses are added in the [`CryptolLiteral` case of `to_cryptol_str`](https://github.com/GaloisInc/cryptol/blob/f46dfa5937d4d3fd3de9664378dc31b460e1b9ad/cryptol-remote-api/python/cryptol/quoting.py#L37).

Got it, that clarifies things.

> Ah, by this do you mean specifically saying to always using `cry_f`-strings and never f-strings when formatting Cryptol?

Correct.

> As for where to put it, definitely in the docstring for `cry_f`, but perhaps also it should also go in the README? If this feature is going to end up getting used a lot it might be worth highlighting it and discussing pitfalls.

If we had some kind of high-level tutorial for the Cryptol Python bindings, that would be a good place to mention it. From a quick search over the various READMEs in the repo, I'm not sure if such a tutorial exists. In any case, I'd be content with mentioning this in the docstring for `cry_f`.

> Unfortunately, since I'm using reflection to make this work, I can't think of an easy way to change what characters delimit f-string arguments. [...] Regardless, I agree that this should be mentioned as a pitfall.

Alright, I'd be fine with simply mentioning this point in `cry_f`'s docstring.

m-yac:
Ok @RyanGlScott I incorporated your suggestions into the docstring for `cry_f`, let me know what you think!

Note that I also improved the way `cry_f`/`cry_eval_f` handles conversions (e.g. `{...!r}`) and format specs (e.g. `{...:+.2f}`) to ensure that the output is always valid cryptol.

cryptol-remote-api/python/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 # Revision history for `cryptol` Python package
 
-## 2.12.2 -- YYYY-MM-DD
+## 2.12.2 -- 2021-11-19
```
RyanGlScott, line 3: I wouldn't bother putting a date here just yet, as we won't know when 2.12.2's release date will be until it actually happens.

cryptol-remote-api/python/cryptol/quoting.py
```diff
@@ -0,0 +1,84 @@
...
+                                           filename=filename)
+
+def cry(s : str) -> CryptolLiteral:
+    """Embed a string of cryptol syntax as ``CryptolCode``"""
+    return CryptolLiteral(s)
+
+def cry_f(s : str) -> CryptolLiteral:
+    """Embed a string of cryptol syntax as ``CryptolCode``, where the given
+       string is parsed as an f-string, and the values within brackets are
+       converted to cryptol syntax using ``to_cryptol_str``.
```
RyanGlScott, line 51: This docstring is a bit inconsistent about whether it capitalizes the C in Cryptol or not in English prose. I'd prefer using capital-C Cryptol whenever it's being used in a context where snake case isn't being used (i.e., not in a Python function name). That is,

```suggestion
    """Embed a string of Cryptol syntax as ``CryptolCode``, where the given
       string is parsed as an f-string, and the values within brackets are
       converted to Cryptol syntax using ``to_cryptol_str``.
```

And similarly for the rest of the docstring.

RyanGlScott (approved):
Fantastic stuff. I have a fewer minor comments about documentation, but other than that, this LGTM.

m-yac:
I realized I should probably not overload `__str__`, so all this latest commit adds is a function `__to_cryptol_str__` to `CryptolJSON` which mirrors `__to_cryptol__` (as well as a better type to `__to_cryptol__` to help make the difference between the two functions clearer).

pnwamk:
This is awesome -- nice job!

The nasty example I was thinking of being nice to clean up some day is the [salsa20 example](https://github.com/GaloisInc/saw-script/blob/8abb655a69ffb80661f7e0de0eb73ca35028e27f/saw-remote-api/python/tests/saw/test_salsa20.py#L58-L62) from saw-script, i.e. lines like these:
```
res = cryptol("quarterround")([y0, y1, y2, y3])
self.points_to(y0_p, cryptol("(@)")(res, cryptol("0")))
```

Am I correct in thinking these can be rewritten as follows with these changes?
```
res = cry_f('quarterround {[y0, y1, y2, y3]}')
self.points_to(y0_p, cry_f('{res}@0'))
```

pnwamk (approved):

m-yac:
Thanks all! Yes, @pnwamk that's exactly right. Though the SAW bindings will need to define its own versions of `cry` and `cry_f` which wrap their respective results in [`CryptolTerm`](https://github.com/GaloisInc/saw-script/blob/8abb655a69ffb80661f7e0de0eb73ca35028e27f/saw-remote-api/python/saw_client/crucible.py#L62) (which is just `CryptolLiteral` + `SetupVal`) to get them to play nice with the rest of the SAW client. I'm working on adding those and updating the SAW tests right now.

PR #1308, opened 2021-11-24
`cryptol-remote-api`: Allow building with `aeson-2.0.*`

RyanGlScott:
This introduces some CPP, unfortunately, but this probably necessary to
continue allowing `cryptol-remote-api` to be buildable with old versions of
`aeson` until the ecosystem catches up with `aeson-2.0.*`.

### DIFF
--- a/cryptol-remote-api/cryptol-remote-api.cabal
+++ b/cryptol-remote-api/cryptol-remote-api.cabal
@@ -42,7 +42,7 @@ common deps
   build-depends:
     base                 >=4.11.1.0 && <4.15,
     argo,
-    aeson                >= 1.4.2 && < 2.0,
+    aeson                >= 1.4.2 && < 2.1,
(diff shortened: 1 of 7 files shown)


pnwamk (approved):
LGTM.

Out of curiosity, why is this an improvement over leaving the upper bound at `2.0` until `the ecosystem widely uses @aeson-2.0.0.0@ or later`?

RyanGlScott:
> Out of curiosity, why is this an improvement over leaving the upper bound at `2.0` until `the ecosystem widely uses @aeson-2.0.0.0@ or later`?

We could do that, I suppose. Supporting `aeson-2.0.*` is a prerequisite to being able to build things with GHC 9.2, however, and I'm not sure whether we'll support 9.2 before or after the ecosystem catches up to `aeson-2.0.*`. If I had to place a bet, I'd bet "before".

Issue #1310, opened 2021-12-08
abc not working in cryptol-remote-api nightly container


michaelabernardo:
Good morning! 

We noticed in the `nightly` cryptol-remote-api container from two days ago that abc does not seem to be working as it previously did when trying to prove a property. 

The error we get is `argo_client.interaction.ArgoException: Internal error ArgoThreadTimedOut` for W4_ABC and for abc,
```
argo_client.interaction.ArgoException: Internal error Failed to start the external solver: 
ArgoThreadTimedOut
Make sure you can start "/usr/local/bin/abc"
from the command line without issues.
```

michaelabernardo:
Also, is this right in [solver.py](https://github.com/GaloisInc/cryptol/blob/master/cryptol-remote-api/python/cryptol/solver.py)? `W4_ABC: OnlineSolver       = OnlineSolver("w4-any")`

weaversa:
> Also, is this right in [solver.py](https://github.com/GaloisInc/cryptol/blob/master/cryptol-remote-api/python/cryptol/solver.py)? `W4_ABC: OnlineSolver = OnlineSolver("w4-any")`

The commit for that is here: https://github.com/GaloisInc/cryptol/commit/5feef24c9a898cc5eb5df1a2af7f6aa968dcbad2

weaversa:
The commit that changed how the solvers were installed in the container is here: https://github.com/GaloisInc/cryptol/commit/c8c122fac61191162c5c5c106735496a317ecba0#diff-92e1e9b869d970ba6ca75eebada07c31dfbda3bf4995b0db2fe5331980912df9

pnwamk:
I'm not sure why things are timing out now where they did not do so before (this can happen sometimes with different solver versions of course, but I haven't specifically verified that's what's going on yet).

As a quick check, instead of using `W4_ABC` (which is incorrectly defined as you noted -- thanks! I've got a fix inbound for that), does using `OnlineSolver("w4-abc")` in its place fix the timeout issue by chance?

michaelabernardo:
> As a quick check, instead of using `W4_ABC` (which is incorrectly defined as you noted -- thanks! I've got a fix inbound for that), does using `OnlineSolver("w4-abc")` in its place fix the timeout issue by chance?

This didn't seem to help as far as I can tell. I mostly still got the timeout error, though in one place, I instead got: 
```
argo_client.interaction.ArgoException: Internal error Could not parse solver response:
[...]

PR #1311, opened 2021-12-08
fix what4-abc solver in python client, add more SMT tests

pnwamk:


pnwamk:
This may partially address https://github.com/GaloisInc/cryptol/issues/1310

Still looking into it... the failure explicitly noted in that issue may be a combination of SBV+ABC not playing well together or similar

cryptol-remote-api/python/tests/cryptol/test_smt.py
```diff
@@ -47,6 +48,38 @@ def test_SMT(self):
...
 
+    def test_each_online_solver(self):
+        # we test each solver that is packaged for use with what4
+        # via https://github.com/GaloisInc/what4-solvers
+        ex_true  = '\(x : [8]) -> x+x == x+x'
+        solvers = \
+            [solver.CVC4,
+             solver.YICES,
+             solver.Z3,
+             #solver.BOOLECTOR,
```
RyanGlScott, line 59: Are there issues with using these commented-out solvers, out of curiosity?

RyanGlScott (commented):
Good catch. Even if this doesn't fix #1310, this is definitely a worthwhile patch.

cryptol-remote-api/python/tests/cryptol/test_smt.py
pnwamk: As sneakily noted above these lines, I was trying to test only the solvers that are distributed in `what4-solvers`. The exception at this point being the SBV ones, but I've added those back into the list with a subsequent commit.

pnwamk: I added a little additional clarification to the comment regarding the intentionally commented out solvers.

RyanGlScott (approved):

PR #1312, opened 2021-12-09
[RPC] Clean up `to_cryptol` interface in cryptoltypes.py

m-yac:
Currently, `cryptoltypes.py` has the functions `to_cryptol`, `CryptolType.from_python`, and `CryptolType.convert`, all of which do approximately the same thing. Only the first is used in any current code in this repo or in saw-script, so this PR removes the latter two.

Currently, `__to_cryptol__` has a `CryptolType` parameter that does nothing, and `to_cryptol` has a `CryptolType` parameter that only matters in the following two cases:
```python
to_cryptol(x, Bitvector(Num(w))) == to_cryptol(BV(w, x)) # if x is an int
to_cryptol(x, Bitvector(Num(w))) == to_cryptol(BV.from_bytes(x, size=w)) # if x is a bytearray or bytes
```
Otherwise every Python value has a unique `CryptolType`, so the parameter is unused. These `CryptolType` parameters are never actually used in any current code in this repo or in saw-script, and I prefer the expressions on the RHS of the equalities above anyway, so this PR also removes these parameters.

NB: This could potentially break someone's code if they use one of the deleted functions or parameters.

### DIFF
--- a/cryptol-remote-api/python/cryptol/cryptoltypes.py
+++ b/cryptol-remote-api/python/cryptol/cryptoltypes.py
@@ -54,7 +54,7 @@ def parenthesize(s : str) -> str:
 JSON = Union[bool, int, float, str, Dict, typing.Tuple, List]
 
 class CryptolJSON(Protocol):
-    def __to_cryptol__(self, ty : CryptolType) -> JSON: ...
+    def __to_cryptol__(self) -> JSON: ...
     def __to_cryptol_str__(self) -> str: ...
 
 class CryptolCode(metaclass=ABCMeta):
@@ -65,7 +65,7 @@ class CryptolCode(metaclass=ABCMeta):
             raise ValueError("Argument to __call__ on CryptolCode is not CryptolJSON")
 
     @abstractmethod
-    def __to_cryptol__(self, ty : CryptolType) -> JSON: ...
+    def __to_cryptol__(self) -> JSON: ...
 
     @abstractmethod
     def __to_cryptol_str__(self) -> str: ...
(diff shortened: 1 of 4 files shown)


RyanGlScott:
> NB: This could potentially break someone's code if they use one of the deleted functions or parameters.

In that case, do mention these changes in [the `CHANGELOG`](https://github.com/GaloisInc/cryptol/blob/490e046b072e2600b4fe32eec258a3570956049f/cryptol-remote-api/python/CHANGELOG.md).

cryptol-remote-api/python/cryptol/cryptoltypes.py
```diff
@@ -185,62 +228,7 @@ def is_plausible_json(val : Any) -> bool:
     return False
 
 class CryptolType:
-    def from_python(self, val : Any) -> JSON:
```
pnwamk, line 188: It's possible there is not any code that explicitly depends on the methods on the parent class `CryptolType` that are deleted in this PR (`from_python` and `convert`)... but to be safe, could we keep those methods and just have them call the functions that now have all of that logic? (is that mostly just `to_cryptol` for both...?)

pnwamk (approved):
LGTM module the one comment regarding possibly keeping the methods on `CryptolType`

cryptol-remote-api/python/cryptol/cryptoltypes.py
m-yac: As per your later suggestion, I put both of these methods back but added errors to both saying they're deprecated.

PR #1317, opened 2021-12-21
[RPC] Finish cleaning cryptoltypes.py

m-yac:
This PR finishes cleaning up `cryptoltypes.py`, specifically:
- Finishes adding classes for all of the currently implemented Cryptol props (see [Cryptol.TypeCheck.TCon](https://github.com/GaloisInc/cryptol/blob/08cd609c4c58ebb5a4cd22ad053d8255ce8d1fc5/src/Cryptol/TypeCheck/TCon.hs#L200) and [CryptolServer.Data.Type](https://github.com/GaloisInc/cryptol/blob/08cd609c4c58ebb5a4cd22ad053d8255ce8d1fc5/cryptol-remote-api/src/CryptolServer/Data/Type.hs#L131))
- Adds class for `lengthFromTo` type operator
- Adds type annotations for all instance variables
- Finishes implementing `__str__` methods for all Cryptol types
- Uses `@dataclass` everywhere possible to avoid writing `__repr__` and `__eq__` methods
- Adds more docstrings
- Adds comments delineating sections

### DIFF
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Revision history for `cryptol` Python package
 
+## 2.12.4 -- YYYY-MM-DD
+
+* NEW CHANGELOG ENTRIES SINCE 2.12.2 GO HERE
+
 ## 2.12.2 -- 2021-12-21
 
 * Add an interface for Cryptol quasiquotation using an f-string-like syntax,
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Revision history for `cryptol` Python package
 
-## 2.12.2 -- YYYY-MM-DD
+## 2.12.2 -- 2021-12-21
 
 * Add an interface for Cryptol quasiquotation using an f-string-like syntax,
   see `tests/cryptol/test_quoting` for some examples.
--- a/cryptol-remote-api/python/cryptol/cryptoltypes.py
+++ b/cryptol-remote-api/python/cryptol/cryptoltypes.py
@@ -10,7 +10,7 @@ from .opaque import OpaqueValue
 
 import typing
 from typing import cast, Any, Dict, Iterable, List, NoReturn, Optional, TypeVar, Union
-from typing_extensions import Literal, Protocol
+from typing_extensions import Literal, Protocol, runtime_checkable
 
 A = TypeVar('A')
 
@@ -51,31 +51,31 @@ def parenthesize(s : str) -> str:
        ``is_parenthesized(s)`` is ``False``"""
     return s if is_parenthesized(s) else f'({s})'
 
-
 JSON = Union[bool, int, float, str, Dict, typing.Tuple, List]
 
+@runtime_checkable
 class CryptolJSON(Protocol):
+    """A ``Protocol`` for objects which can be converted to Cryptol JSON or
+       Cryptol strings."""
     def __to_cryptol__(self) -> JSON: ...
     def __to_cryptol_str__(self) -> str: ...
 
 class CryptolCode(metaclass=ABCMeta):
-    def __call__(self, *others : CryptolJSON) -> CryptolCode:
-        if all(hasattr(other, '__to_cryptol__') for other in others):
-            return CryptolApplication(self, *others)
-        else:
-            raise ValueError("Argument to __call__ on CryptolCode is not CryptolJSON")
-
+    """The base class for ``CryptolLiteral`` and ``CryptolApplication``."""
     @abstractmethod
     def __to_cryptol__(self) -> JSON: ...
-
     @abstractmethod
     def __to_cryptol_str__(self) -> str: ...
 
     def __str__(self) -> str:
         return self.__to_cryptol_str__()
 
+    def __call__(self, *others : CryptolJSON) -> CryptolCode:
+        return CryptolApplication(self, *others)
+
 @dataclass
 class CryptolLiteral(CryptolCode):
+    """A string of Cryptol syntax."""
     _code : str
 
     def __to_cryptol__(self) -> JSON:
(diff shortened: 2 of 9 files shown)


cryptol-remote-api/python/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Revision history for `cryptol` Python package
 
-## 2.12.2 -- YYYY-MM-DD
+## 2.12.4 -- YYYY-MM-DD
+
+* NEW CHANGELOG ENTRIES SINCE 2.12.2 GO HERE
+
+## 2.12.2 -- 2021-12-21
```
pnwamk, line 7: Let's make an explicit `2.12.2 release` commit, _then_ bump things past it

cryptol-remote-api/python/cryptol/cryptoltypes.py
```diff
@@ -50,49 +57,54 @@ def parenthesize(s : str) -> str:
...
+    _rands : typing.Sequence[CryptolJSON]
+
     def __init__(self, rator : CryptolJSON, *rands : CryptolJSON) -> None:
-        self._rator = rator
-        self._rands = rands
+        if all(isinstance(rand, CryptolJSON) for rand in rands):
+            self._rator = rator
+            self._rands = rands
+        else:
+            raise ValueError("Arguments given to CryptolApplication must be CryptolJSON")
```
pnwamk, line 104: Might be nice to include the arguments in the error string

cryptol-remote-api/python/cryptol/cryptoltypes.py
```diff
@@ -105,22 +117,19 @@ def __to_cryptol_str__(self) -> str:
...
+    domain : CryptolKind
+    range : CryptolKind
 
-    def __repr__(self) -> str:
-        return f"CryptolArrowKind({self.domain!r}, {self.range!r})"
-
-CryptolKind = Union[Literal['Type'], Literal['Num'], Literal['Prop'], CryptolArrowKind]
+CryptolKind = Union[typing_extensions.Literal['Type'],
+                    typing_extensions.Literal['Num'],
+                    typing_extensions.Literal['Prop'], CryptolArrowKind]
```
pnwamk, line 133: can `CryptolArrowKind` live with it's union-mates on it's own new line too? minor I know xD sorry

cryptol-remote-api/python/CHANGELOG.md
m-yac: Good suggestion! This is now ded73dc

cryptol-remote-api/python/cryptol/cryptoltypes.py
```diff
@@ -229,230 +304,183 @@ def is_plausible_json(val : Any) -> bool:
...
         self.types = types
 
     def __repr__(self) -> str:
-        return "Tuple(" + ", ".join(map(str, self.types)) + ")"
+        return "Tuple(" + ", ".join(map(repr, self.types)) + ")"
 
-class Record(CryptolType):
-    def __init__(self, fields : Dict[str, CryptolType]) -> None:
-        self.fields = fields
+    def __str__(self) -> str:
```
pnwamk, line 472: Is `Record` not supposed to have a `__repr__`? It's not obvious to me yet why some classes have `__repr__` and some don't. Ah, is it because `@dataclass` is giving us the one we want for some but not others?

pnwamk: Oops, pretend this comment is a few lines down where `Record` is defined xD

cryptol-remote-api/python/cryptol/opaque.py
```diff
@@ -1,5 +1,7 @@
 from typing import Any
+from dataclasses import dataclass
 
+@dataclass(frozen=True)
```
pnwamk, line 4: Can we get a comment why this is `frozen` but other `dataclass`es are not?

pnwamk (approved):
LGTM modulo minor comments.

yav:
I haven't followed closely what's been happening with the Python API, but looking at the files here I am a getting worried that we are duplicating a whole bunch of Cryptol in Python, which might lead to maintenance headaches.  I am thinking of some of the things in `cryptoltypes.py` specifically.  If we do need such code, I wonder if we could, perhaps, auto-generate it from Haskell, which is likely to be less error prone and then if we make changes in Cryptol, they'll just get reflected in the python client.   Thoughts?

m-yac:
> I haven't followed closely what's been happening with the Python API, but looking at the files here I am a getting worried that we are duplicating a whole bunch of Cryptol in Python, which might lead to maintenance headaches. I am thinking of some of the things in cryptoltypes.py specifically. If we do need such code, I wonder if we could, perhaps, auto-generate it from Haskell, which is likely to be less error prone and then if we make changes in Cryptol, they'll just get reflected in the python client. Thoughts?

This is an excellent idea! I have been uneasy about the fact that we're duplicating code here for a while, but I didn't get to the point where I realized we could just generate it. I'm going to experiment with this.

pnwamk:
> I haven't followed closely what's been happening with the Python API, but looking at the files here I am a getting worried that we are duplicating a whole bunch of Cryptol in Python, which might lead to maintenance headaches. I am thinking of some of the things in `cryptoltypes.py` specifically. If we do need such code, I wonder if we could, perhaps, auto-generate it from Haskell, which is likely to be less error prone and then if we make changes in Cryptol, they'll just get reflected in the python client. Thoughts?

I empathize with the overall concerns you describe.

This PR is specifically aimed at tidying up `cryptoltypes.py`. Perhaps while this is fresh in our minds, we could add some comments in this PR describing not only the "what" but the "why" for the code in here (specifically, why does the python client need to do these things, instead of delegating the work to the server? Some parts might be obvious, but I suspect there are a few design decisions/justifications which are more subtle we might forget from time to time).

As for auto generating the `cryptoltypes.py` file... it's not clear to me how much of a win/pain that would be. What kinds of changes in Cryptol would actually "break" the code in that file? How often are those likely to happen? Aren't we (or shouldn't we) _already_ be testing for those exact kinds of things in our CI? I'm happy to dive into this further in more detail elsewhere; at the moment in my head it feels like a premature optimization/abstraction given the size and content, I dunno ¯\\\_(ツ)\_/¯

cryptol-remote-api/python/cryptol/cryptoltypes.py
m-yac: Added a comment saying why `Tuple` has an `__init__` and `__repr__` while all the others don't:
```python
    # we override the @dataclass __init__ and __repr__ because we want the
    #  syntax of variable numbers of arguments
```
Does that make sense?

cryptol-remote-api/python/cryptol/opaque.py
m-yac: Are you satisfied by the comment I added?

m-yac:
Inspired by a conversation between me and @pnwamk, I've taken the opposite approach and have deleted as many of these classes as I could. Specifically, I added the `TypeOp` subclass of `CryptolType`, which incorporates all of the type level operators (previously `Plus`, `Max`, etc.), so if one were to be removed/changed/added, we wouldn't have to update the Python interface. I also did the same for `CryptolProp`s.

I also cleaned up `CryptolTypeSchema` and used it in the `check_type` command – so now `check_type` will work correctly on functions.

Finally, I updated `test_types.py` to reflect these changes.

cryptol-remote-api/python/cryptol/cryptoltypes.py
pnwamk: Yes

PR #1358, opened 2022-05-23
Only include bindist-related files in `cryptol.msi`

### LINKED ISSUE #977: Windows installer includes too many files?
The Windows `.msi` installer we produce for releases is 4x the size of the corresponding tarball. It may partly be using less powerful compression, but I'm suspicious that it might include multiple copies of the binaries, in different locations.

RyanGlScott:
Previously, we would include every file in the current `cryptol` checkout in `cryptol.msi`, which would result in horribly bloated installer files. We now only include those files under the `dist` directory (created by `.github/ci.sh bundle_files`), which curates only those files we want to include in a binary distribution.

While I was in town, I modified the conventioned used in `win32/cryptol.wxs` so that we no longer need to reference `dist`, which does not show up in `.tar.gz`-based bindists.

Fixes #977.


kquick (approved):
This seems very reasonable.  I have no Windows system to verify this with, but I don't anticipate any particular issues.

PR #1362, opened 2022-06-02
Always embed `git` commit into `--version` output

### LINKED ISSUE #1353: `cryptol --version` doesn't show git commit
If I run `cryptol --version` using Cryptol 2.13.0, it produces:

```
$ ~/Software/cryptol-2.13.0/bin/cryptol --version
Cryptol 2.13.0
Git commit UNKNOWN
    branch UNKNOWN
```

Note the `UNKNOWN`s. Something similar happens with Cryptol 2.12.0.

I'm not entirely sure why this happens off the top of my head. I would imagine that GitHub Actions (which builds the binary distributions) has `git` installed, which is a prerequisite to compiling this information directly into the binary.

RyanGlScott:
Previously, we attempted to scrub out `git` commits from release binaries using a `sed` invocation in the `cry` script, but as was observed in #1353, this ended up scrubbing out `git` commits on all builds, not just release ones. We have decided that we should just include `git` commits on all builds, so we can fix the issue by removing the buggy use of `sed`.

Unfortunately, `cabal`'s recompilation logic is not smart enough to always rebuild `GitRev.hs` on every new commit, even though it should. For this reason, we make the `./cry build` command temporarily dirty the `GitRev.hs` file so that it is forcibly rebuilt, then revert the change to `GitRev.hs` and rebuild. It's rather heavy-handed, but it works.

Fixes #1353.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,10 @@ jobs:
       release: ${{ steps.config.outputs.release }}
       retention-days: ${{ steps.config.outputs.retention-days }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: config
         id: config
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Alright, I think I'm happy with this approach now. I've verified that the `--version` output with the binary artifacts from this PR looks like what you would expect:

```
$ ./cryptol-3.0.0.99/bin/cryptol --version
Cryptol 3.0.0.99
Git commit defd7d4016b971884b133f626870b05d0ed08c62
    branch HEAD
FFI enabled
```

I had to tweak the `checkout` action's behavior regarding whether to use a merge commit in CI (the default) or whether to use the HEAD commit from the branch during CI. The default was not ideal since it would cause `--version` to print a merge commit that is only created for CI purposes and doesn't actually point to anything in the `cryptol` repo itself.

PR #1363, opened 2022-06-13
New module system

yav:
Cryptol's new module system rebased on top of `master`

## TODO

- [X] #1366
- [X] #1368
- [X] #1371
- [X] #1372
- [x] #1373
- [ ] #1375
- [x] Make sure browsing modules on the REPL still works
- [ ] Make sure the [cryptol-specs](https://github.com/GaloisInc/cryptol-specs) repo still builds


tests/modsys/T5.icry.stdout
```diff
@@ -1,7 +1,12 @@
...
 Loading module Cryptol
-Loading module T5::A
-Loading module T5::B
-Loading module T5::Main
-main : [8]
-0x11
+Loading module T5::Main$argument
+
+Parse error at ./T5/B.cry:4:3--4:4
+  Module argument may not be a functor
```
m-yac, line 6: Is this test supposed to be failing? If it is:
1. I think the error message should be improved, it doesn't seem like "Module argument may not be a functor" is applicable here.
2. We should probably add a line to the docs stating that partially instantiating parameterized modules is no longer supported.

This same error happens with `T6.icry` and `T12.icry` (which have this error committed) and `tests/issues/issue796.icry` (which does not have it committed).

yav: This fails because anonymous argument module to `A` (i..e., the `where` block) is a functor, but it is not allowed to be.  Perhaps we should rephrase the error, please just change it if you have ideas, the message is in `ParserUtils` I think.

Partial module instantiations never worked, but this sort of pattern used to mean that you are defining a functor.  We decided that this notation is confusing (because the `parameters` are for `B` but they appear in the argument to `A`), so we should not support it.  I am not sure how official this feature ever was, but feel free to add a note to the docs to clarify that this does not work anymore.

Riib11:
Why does this give me a dependency error?
```
module A

interface submodule I where 
  type a : *
  
import interface submodule I
```
Causes this error:
```
[error] Invalid recursive dependency:
    • module A
    • submodule A::I, defined at 3:21--3:22
```

robdockins:
That would make the submodule `I` into a dependency of `A` (as it defines the parameters of `A`), but `I` is defined in `A`. @yav and I talked about this situation, and in general it seems very difficult to support.

robdockins:
> > That would make the submodule `I` into a dependency of `A` (as it defines the parameters of `A`), but `I` is defined in `A`. @yav and I talked about this situation, and in general it seems very difficult to support.
> 
> It's interesting because this is how I imagine anonymous interfaces working implicitly i.e.
> 
> ```
> parameter 
>   a : *
> ```
> 
> But I guess not quite

Indeed, the desugaring has to play some tricks to make sure that the anonymous interface module is defined "outside" the module that uses it.

Riib11:
Why does this give me a `Module not in scope: Lib1_Inst` error while loading `Lib1`
```
module Lib1 where

parameter
  type a : *


module Lib2 where

parameter 
  type b : *

submodule Lib1_Inst = Lib1
  where 
  type a = b

import submodule Lib1_Inst
```

Whereas this does not given any errors?
```
module Lib1 where

parameter
  type a : *


module Lib2 where

submodule Lib1_Inst = Lib1
  where 
  type a = [8]

import submodule Lib1_Inst
```

Is it for a similar reason as we just discussed with the previous example of importing a local interface?

Riib11:
> @Riib11 I think this is due to the bug discussed yesterday on the channel---currently the desugaring for `parameter` blocks makes an interface to be placed outside the functor. At the moment it also adds all imports from the module to the new interface. This however is wrong, because some of the imports might refer to names defined in the module, and so they don't make sense outside.
> 
> It is hard to know what's a good solution here, but the current idea is that we'll only move _top-level_ imports to the new signature. @m-yac is working on implementing this I believe.
> 
> Concretely, in your example, the new interface incorrectly gets the `import submodule Lib1_Inst` which leads to the error you are seeing.

That all makes sense. 

Basically what I'm trying to accomplish is "functor eta-expansion" in order to allow a parametrized module B to be imported by another module A that is parametrized by the same (of perhaps some subset of) parameters. And even when this bug is fixed, it's still a little clunky to do this.

m-yac:
> Why does this give me a `Module not in scope: Lib1_Inst` error while loading `Lib2`
> 
> ```
> module Lib1 where
> 
> parameter
>   type a : *
> 
> 
> module Lib2 where
> 
> parameter 
>   type b : *
> 
> submodule Lib1_Inst = Lib1
>   where 
>   type a = b
> 
> import submodule Lib1_Inst
> ```

@Riib11 I just pushed the change @yav suggested - this example should work now. I also got a bunch more of the tests working. I think the only one that's still failing is `tests/regression/repl-decls.icry` because for some reason:
```
let x -<- y = x - y; infixl 5 -<-; let (-<-) : Integer -> Integer -> Integer
let x ->- y = x - y; infixr 5 ->-; let (->-) : Integer -> Integer -> Integer

42 -<- 10 -<- 100
42 ->- 10 ->- 100
42 ->- 10 -<- 100
```
outputs
```
> -68
> -68
> -68
```
instead of what is used to, which was:
```
-68
132

[error] at repl-decls.icry:6:4--6:7 and repl-decls.icry:6:11--6:14
    The fixities of
      • (->-) (precedence 5, right-associative)
      • (-<-) (precedence 5, left-associative)
    are not compatible.
    You may use explicit parentheses to disambiguate.
```
Any idea why this PR changes precedence behavior? There aren't even any modules or imports in this file.

Riib11:
Unfortunately I'm having a different problem now. When I run `Lib2`, I'm told `Value not in scope: x`.
```
module Lib1 where
parameter
  type a : *
  x : a


module Lib2 where
parameter
  type b : *
  y : b

submodule Lib1_Inst = Lib1
  where
  type a = b
  x = y

import submodule Lib1_Inst
z : b
z = x
```

yav:
Can you make tickets if you find issues, so we can keep track of the various things that need fixing.   Let's try to keep the PR comments for specific issues related to the code.

tests/modsys/T5.icry.stdout
m-yac: In 70d96006a5526a94dc72d5ee3bc84096770b7e3b I changed this message to "Instantiation of a parameterized module may not itself be parameterized" since this error is only raised when desugaring an anonymous instantiation of a parameterized module.

PR #1376, opened 2022-06-25
Foreign Function Interface

qsctr:
This PR adds a foreign function interface (FFI) to Cryptol. This allows Cryptol to call functions written in C.

For instance, we can declare this foreign function in Cryptol
```
foreign add : [32] -> [32] -> [32]
```
and implement it in C
```c
uint32_t add(uint32_t x, uint32_t y) {
  return x + y;
}
```
and then call it like any other Cryptol function.
```
Main> add 1 2
0x00000003
```

See the [FFI docs](https://github.com/GaloisInc/cryptol/blob/ffi/docs/RefMan/FFI.rst) for more information.


cryptol.cabal
```diff
@@ -58,6 +58,7 @@ library
                        GraphSCC          >= 1.0.4,
                        heredoc           >= 0.2,
                        libBF             >= 0.6 && < 0.7,
+                       libffi,
```
RyanGlScott, line 61: It would be worth guarding this dependency behind a Cabal flag. The reason is that Windows doesn't come with `libffi` by default, so most Windows users who try to naïvely install this would run into a build failure. Guarding it behind a Cabal flag would ensure that the solver can detect this situation and avoid depending on `libffi` if the user hasn't installed it.

The flip side to doing this is that we'd need a dummy implementation of the FFI that always fails if the user doesn't have `libffi` installed.

src/Cryptol/Eval/FFI.hs
```diff
@@ -0,0 +1,55 @@
...
+import           Cryptol.ModuleSystem.Name
+import           Cryptol.TypeCheck.AST
+import           Cryptol.Utils.Ident
+import           Cryptol.Utils.Panic
+
+addForeignDecls :: ModulePath -> Module ->
+  EvalEnv -> Eval (Maybe ForeignSrc, EvalEnv)
+addForeignDecls path m env = io do
+  foreignSrc <- newIORef Nothing
+  let addForeignDeclGroup e (Recursive ds)   = foldlM addForeignDecl e ds
```
robdockins, line 29: Probably should just refuse to consider recursive groups containing foreign declarations, as this shouldn't happen.

src/Cryptol/Eval/FFI.hs
```diff
@@ -0,0 +1,55 @@
...
+          impl <- loadForeignImpl fsrc $ unpackIdent $ nameIdent $ dName d
+          pure $ bindVarDirect (dName d) (foreignPrim impl) e
+        _ -> pure e
+  env' <- foldlM addForeignDeclGroup env $ mDecls m
+  fsrc <- readIORef foreignSrc
+  pure (fsrc, env')
+
+foreignPrim :: ForeignImpl -> Prim Concrete
+foreignPrim impl = PStrict \case
+  VWord 64 wv -> PPrim $ fmap (VWord 64 . wordVal) $
```
robdockins, line 51: This will need to be done more carefully in general, but is probably OK to test with for now. In general, you may also have a `VSeq` value here representing an explicit sequence of bits, which will need to be converted to a "packed" word.

robdockins: Sorry... nevermind. I've misremembered the value invariants here.

src/Cryptol/Eval/FFI.hs
qsctr: Yeah, I was thinking that it could happen when we add the ability to provide backup Cryptol implementations of foreign functions, but I can remove it for now.

cryptol.cabal
```diff
@@ -264,6 +264,9 @@ executable cryptol
       ld-options: -static -pthread
       ghc-options: -optl-fuse-ld=bfd
 
+  if flag(ffi)
+      ld-options: -Wl,rpath,$ORIGIN
```
RyanGlScott, line 268: I think this should be `-rpath=$ORIGIN`. The `-Wl` part is a prefix required by C compilers that direct them to pass the things following it to the linker, but since `ld-options` passes the options directly to the linker, we don't need the `-Wl` part.

qsctr: From running cabal with `-v` it seems like what `ld-options` does is pass it prepended with `-optl` to GHC, which then passes the rest of it to the C compiler, so I think we still need the `-Wl,`. I did forget to put `-` before `rpath` though.

RyanGlScott: Ah, I see. My mistake!

.github/ci.sh
```diff
@@ -33,6 +33,16 @@ extract_exe() {
...
 }
 
+extract_libffi() {
+  libffi=("$(ghc --print-libdir)"/rts/libffi.*)
+  mkdir -p "$1"
+  cp "${libffi[@]}" "$1"
+}
+
+extract_libffi_for_bin() {
+  extract_libffi "$(dirname "$(cabal v2-exec which "$1$EXT")")/../lib"
```
RyanGlScott, line 43: One problem with putting the shared libraries under `../lib` is that we package up `dist/bin` as an artifact for use in later stages of the CI pipeline. (See [here](https://github.com/GaloisInc/cryptol/blob/9d68dc91ae7fe7fc5b7bb78484fe3909950afb25/.github/workflows/ci.yml#L206-L210).) As a result, this artifact will not include any of the shared libraries, since it doesn't include `dist/lib`. This doesn't break anything on CI currently, since none of these artifacts are used on OSes where the choice of `libffi` version matters, but this could cause problems down the road.

I think the simplest solution would be to copy these shared libraries to somewhere under `dist/bin` so that they are included in the artifacts.

cryptol-remote-api/cryptol-remote-api.cabal
```diff
@@ -100,6 +104,9 @@ executable cryptol-remote-api
   if os(linux) && flag(static)
       ld-options:      -static -pthread
 
+  if os(linux) && flag(ffi)
```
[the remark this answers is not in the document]
qsctr: Upon further investigation we discovered that GHC actually links `libffi` statically on macOS, so we don't have to do this.

src/Cryptol/TypeCheck/Solve.hs
```diff
@@ -287,15 +289,16 @@ proveImplication lnam as ps gs =
 
 
 proveImplicationIO :: Solver
+                   -> Bool     -- ^ Whether to remove duplicate goals in errors
```
yav, line 292: I wonder if really need the extra argument, isn't it reasonable to always remove duplicate goals?

src/Cryptol/Eval/FFI.hs
```diff
@@ -0,0 +1,289 @@
...
+    marshalArgs typesAndVals \inArgs -> do
+      tyArgs <- traverse marshalTyArg ffiTParams
+      let tyInArgs = tyArgs ++ inArgs
+      marshalRet ffiRetType GetRet
+        { getRetAsValue = callForeignImpl impl tyInArgs
+        , getRetAsOutArgs = callForeignImpl impl . (tyInArgs ++) }
+
+  -- Look up the value of a type parameter in the type environment and marshal
+  -- it.
+  marshalTyArg :: TParam -> Eval SomeFFIArg
```
yav, line 145: Perhaps it makes sense to use `evalFinType` here?

yav (approved):
Looks good to me.   I made a few suggestion for stylistic improvements.   On the topic of style, I really prefer the `case` expressions that do not use inverted indentation because it is quite a bit easier to see what is being matched where.  For example like this:

I like:
```
function = some long prefix thing
   case expr of
       A -> thing1
       B -> thing2
```
rather than:

```
function = some long prefix thing case expr of
  A -> thing1
  B -> thing2
```

src/Cryptol/TypeCheck/Solve.hs
qsctr: `Goal` has a weird `Eq` and `Ord` instance where it only compares the actual `Prop` and not where it came from, so I thought it might still be useful to report if you needed to satisfy the same `Prop` but in two different places. In the case of foreign declarations the source is always the same (it just says it's coming from that foreign declaration) so it's safe to deduplicate it by comparing `Prop`s.

I could have also deduplicated by comparing all the fields, but that would have required adding an `Eq` instance for `ConstraintSource` and then for `Expr`, which seemed like a lot of changes and I wasn't sure if `Expr` lacked an `Eq` instance for a reason.

docs/RefMan/FFI.rst
```diff
@@ -0,0 +1,270 @@
...
+==========================
+
+The foreign function interface (FFI) allows Cryptol to call functions written in
+C (or other languages that use the C calling convention).
+
+Platform support
+----------------
+
+The FFI is currently **not supported on Windows**, and only works on Unix-like
+systems (macOS and Linux).
```
Riib11, line 11: MacOS

PR #1379, opened 2022-07-08
`cryptol-remote-api`: Don't use verbose SBV settings

### LINKED ISSUE #1378: cryptol-remote-api dumping SBV debug info
During our CI runs, calls that involve SBV are dumping debug lines to the log, such as `[ISSUE] (define-fun s832 () (_BitVec 28) (concat s732 s831)` -- dumping SMT files and other large things. This is recent, and does not happen when calling solvers via W4. The `[ISSUE]` line shows up here: https://github.com/LeventErkok/sbv/blob/bbab5b9aa5ef63d393120cca2e93df3b703626a1/Data/SBV/SMT/SMT.hs#L864
We also see other debug commands such as `[SEND]` and `[RECV]`. Is there a way to quiet this noisiness?

RyanGlScott:
We should try to match Cryptol's default settings when invoking the remote API, but the `pcVerbose` value (`True`) didn't match the defaults. Changing this to `False` should fix #1378. There is a separate question of whether remote API users should be able to change this default, but for the time being, let's at least be consistent between Cryptol's settings and the remote API's settings.

### DIFF
--- a/cryptol-remote-api/src/CryptolServer/Sat.hs
+++ b/cryptol-remote-api/src/CryptolServer/Sat.hs
@@ -67,7 +67,7 @@ proveSat (ProveSatParams queryType (ProverName proverName) jsonExpr hConsing) = 
             ProverCommand
             { pcQueryType    = queryType
             , pcProverName   = proverName
-            , pcVerbose      = True
+            , pcVerbose      = False
             , pcProverStats  = timing
             , pcExtraDecls   = decls
             , pcSmtFile      = Nothing


RyanGlScott:
@weaversa, can you see if this issue fixes #1378 for you? You should be able to grab a binary distribution of `cryptol-remote-api` [here](https://github.com/GaloisInc/cryptol/actions/runs/2636566015)—let me know if you don't have the necessary permissions to view that.

weaversa:
> @weaversa, can you see if this issue fixes #1378 for you? You should be able to grab a binary distribution of `cryptol-remote-api` [here](https://github.com/GaloisInc/cryptol/actions/runs/2636566015)—let me know if you don't have the necessary permissions to view that.

It's not easy for me to test. I have to build from source and the process is in some automated flow. Would you be willing to merge and I will let you know afterwards?

RyanGlScott:
Ah, OK. In any case, I think this is a pretty low-risk change, so I'm fine with just merging it as-is.

PR #1380, opened 2022-07-08
Numeric Constraint Guards

Riib11:
This PR implements a new feature: _numeric constraint guards_.

_Numeric constraint guards_ are like guards (e.g. Haskell) except that the guarding condition is a numeric constraint. The following example includes the currently-implemented prototype syntax:
```
f : {n} [n] -> [8]
f x = propguards
  | (n == 0) => g1 x
  | (n <= 4) => g2 x
  | (5 <= n) => g3 x

g1 : [0] -> [8]
g1 x = 1

g2 : {n} (n <= 4) => [n] -> [8]
g2 x = 2

g3 : {n} (5 <= n) => [n] -> [8]
g3 x = 3
```

Without this feature, `f` would have to be written verbosely and awkwardly like so as `f'`:
```
f' : {n} fin n => [n] -> [8]
f' x =
  if n == 1 then g1 (take `{0} x) else
  if n <= 4 then g2 (take `{min 4 n} x)
            else g3 (pad `{5} x)
  where
    n = `n

pad : {m, n} (fin m, fin n) => [n] -> [max m n]
pad = ...
```

For more complicated constraints, this second kind of implementation becomes more and more cumbersome.
// TODO: more cumbersome example

// TODO: practical everyday examples

## Implementation

The significance of this feature is that the guarding numeric constraint is introduced as as assumption _locally_ to the guarded expression. 

Immediately after parsing (and the NoPat pass), each guarded expression is expanded into an auxiliary top-level declaration that has the same type signature as the declaration the constraint guards appeared in, but with the guarding constraints appended to its signature's constraint. The original guarded expression is replaced with an appropriate reference to this auxiliary definition.

During typechecking, the guarding constraints of each case are assumed locally before typechecking the guarded expression.

During evaluation, the constraint guards are folded over until a constraint evaluates to `True` (via `evalProp`), in which case only that constraints's guarded expression is evaluated as the result.

## TODO

- [ ] Exhaustivity checking for a declaration's set of guarding constraints
- [ ] Exclusivity checking for a declaration's set of guarding constraints
- [ ] What kinds of constraints are allowed in the guarding constraints? Currently allowed are:
  - [ ] equality of pair of numeric literals
  - [ ] non-equality of pair of numeric literals
  - [ ] inequality of pair of numeric literals
  - [ ] fin-ness of a numeric literal
  - [ ] prime-ness of numeric literal


yav:
I tried these two examples:
```len : {n,a} (fin n) => [n] a -> Integer
len x = propguards
  | (n == 0) => 0
  | (n > 1) => len (drop`{1} x)

len2 : {n,a} (fin n) => [2^^n] a -> Integer
len2 x = propguards
  | (n == 0)  => 1
  | (n > 1)   => (len2 a + len2 b)
                    where (a,b) = splitAt `{n/2} x
```
The first one type checked correctly, although I got a panic:
```
 Message:   cannot use this as a guarding constraint: 
             1 <= n`902
```
Is that because the evaluator is not yet implemented?

The second one also works correctly from the point of view of this extension, but it reveals some additional facts we need to
teach the solver.   I'll make separate tickets for them.

Riib11:
> I tried these two examples:
> 
> ```
> len x = propguards
>   | (n == 0) => 0
>   | (n > 1) => len (drop`{1} x)
> 
> len2 : {n,a} (fin n) => [2^^n] a -> Integer
> len2 x = propguards
>   | (n == 0)  => 1
>   | (n > 1)   => (len2 a + len2 b)
>                     where (a,b) = splitAt `{n/2} x
> ```
> 
> The first one type checked correctly, although I got a panic:
> 
> ```
>  Message:   cannot use this as a guarding constraint: 
>              1 <= n`902
> ```
> 
> Is that because the evaluator is not yet implemented?
> 
> The second one also works correctly from the point of view of this extension, but it reveals some additional facts we need to teach the solver. I'll make separate tickets for them.

The first example doesn't work because you didn't quantify `n` in a signature.

Riib11:
> @Riib11 I don't think we need the `propguards` keyword. I tried the following variation, and it seems to work OK:
> 
> ```
> len : {n,a} (fin n) => [n] a -> Integer
> len x 
>   | (n == 0) => 0
>   | (n > 1) => len (drop`{1} x)
> ```
> 
> This is all I changed:
> 
> ```
> -  | var apats_indices '=' propguards
> -                           { mkIndexedPropGuardsDecl $1 $2 $4 }
> +
> +  | var apats_indices '|' propguards_cases { mkIndexedPropGuardsDecl $1 $2 $4 }
> ```

Awesome! I had added that keyword to deal with an issue getting happy to accept my grammar and couldn't find a better way, but looks like you found one.

src/Cryptol/Parser.y
```diff
@@ -298,11 +300,27 @@ mbDoc                   :: { Maybe (Located Text) }
...
   | {- empty -}            { Nothing }
 
+propguards :: { [([Prop PName], Expr PName)] }
+  : 'propguards' '|' propguards_cases { $3 }
+
+propguards_cases :: { [([Prop PName], Expr PName)] }
+  : propguards_case '|' propguards_cases { $1 ++ $3 }
+  | propguards_case { $1 }
+
+propguards_case :: { [([Prop PName], Expr PName)] }
```
yav, line 310: It might be nicer if we make the `|' part of `propguards_case`, as each of them starts with a `|`.  Then `propguards_cases` would be just a list of 1 or more `propguards_case`, and we wouldn't need the starting `|` in the declaration.

Riib11: I'll see what I can do, this is another place where I ran into difficulties making happy accept the grammer

weaversa:
This is great work!

I have some questions about syntax -- I'm trying to read the semantics out-loud for the following:

```
len x 
  | (n == 0) => 0
  | (n > 1) => 1 + len (drop`{1} x)
```

I want to say something like "len x is a function that, when n equals 0, produces 0 and when n is greater than 1 produces 1 plus ... So, here's I'm interpreting `|` as when. Is there a better way to say that?

Also, we normally write a function with an `=`, but when using this new proposed feature, we see `=>`, which is notation that currently separates type constraints from the type of a function. What are you thoughts on using `=` instead?

```
len : {n,a} (fin n) => [n] a -> Integer
len x
  | (n == 0) = 0
  | (n > 1)    = 1 + len (drop`{1} x)
```

And, how does whitespace work. Can we do this?

```
len : {n,a} (fin n) => [n] a -> Integer
len x | (n == 0) => 0
         | (n > 1) => 1 + len (drop`{1} x)
```

How do `where` clauses scope? Can we do this:

```
fun : {n,a} (fin n) => [n] a -> Integer
fun x 
  | (n == 0) => z
  | (n > 1) => z + len (drop`{1} x)
  where z = 2
```

weaversa:
Here's a workup that follows Crypto's `if then else` shortcutting, and allows for a default case.

```
len x =
    case (n == 0) => 0
     |   (n == 1) => 1
    default       => 1 + len (drop`{1} x)
```

weaversa:
One benefit of the workup below is that there's no special syntax.

```
len : {n,a} (n == 0) => [n] a -> Integer
len x = 0

len : {n,a} (n > 1) => [n] a -> Integer
len x = 1 + len (drop`{1} x)
```

yav:
Thanks for the feedback, @weaversa!

On the question of how to read `| P => e`, I like your suggestion for "when `P` holds then the definition of `len` is `e`.

For the separator between the predicates and the definition, `=` would work just as well.  I don't have a strong opinion on if we should use `=` or `=>`, my guess would be that @Riib11 picked the `=>` as a hint that the predicates on the left of it are *type constraints* and not ordinary values.

There is no layout involved with the `|` so you don't need to line them all up (although it is probably a good idea to do so to help with readability).

The `where` in Cryptol is part of the expression language, which is different from Haskell where it is part of a declaration.  So writing a `where` at then end would simply make it part of the definition for the last alternative.

We considered something quite similar to your suggestion with the `case` syntax, but decided against it for the following reasons:
   1. one can't do type inference in the presence of local constraints (such as the predicates in each alternative), because most general types don't exist (i.e., it is not clear what type one should infer).   So it is important that the construct has a type signature associated with, and it seemed most natural to link it to a declaration, rather than an expression.
   2. The "default" construct is problematic because it involves negation (i.e., you are saying "if none of the above cases match, then do this").  Unfortunately, we can't negate all predicates in Cryptol, because this would result in constraint disjunctions, and these have a whole separate set of issues that would need to be solved.  So, for the time being, each guard is considered on its own (i.e., none of them have the implicit assumption that their predecessor didn't match).

We also considered the multi equation form, it's biggest benefit being that
it could support argument patterns that benefit from the additoinal constraints.
Once the basics are working, we may revisit this and add support for
multi-equation definitions, althouhg we'd need to figure out how to write the
constraint guards.  Having a separate full signature on each equation is not
great because:
  1. There is a lot of duplication---this makes it harder to write and maintain
     the code, and also adds additional burden on the implementation in that
     we have to check that the signatures are the same; we'd also have to
     decide exactly what it means to be "the same".
  2. The function being declared needs to have a single type, so that
     when you call it you need to know what to check.  This is also nice for
     the users of the function, in that they can easily see the type of the
     function.  After all, the fact that the definition is defined
     by cases is an implementation detail, and it doesn't need to be exposed to
     the callers.

Your intuition is spot on though---the implementation under the hood does
pretty much exactly what you suggest.  More concretely, the example:
```
len x : {n} (fin n) => [n]a -> Integer
len x
  | (n == 0) => 0
  | (n > 1) => 1 + len (drop`{1} x)
```
internally us represented like this:
```
len x : {n,a} (fin n) => [n]a -> Integer
len x = type_if
          | n == 0 then len_A`{n,a} x
          | n >  1 then len_B`{n,a} x

len_A : {n,a} (fin n, n == 0) => [n]a -> Integer
len_A x = 0

len_B : {n,a} (fin n, n > 1) => [n]a -> Integer
len_B x = 1 + len (drop`{1} x)
```
The `type_if` is the new (internal) Cryptol extension that does the magic
type-based dispatch.

weaversa:
Thanks @yav! @auricratio mentioned that the 'multi equation form' feels a lot like type classes. For example, two operators could be defined, but w/ the same name, one that operates on bit vectors and one on `Integer`s.

Riib11:
@weaversa I like your suggestion to add a default case. There is a `True` Cryptol `Prop` (i.e. type constraint), so the default case should simply have `True` as it's constraint. And I'm pretty sure using `default` as a keyword there works fine as well.

yav:
@Riib11 I think using up a whole new keyword like `default` just for this case is a bit of an overkill.   I suspect one could define the `true` constraint in existing Cryptol if they wanted to though, like this:
```
type constraint Default = ()

f : Default => Bit
f = False
```

Riib11:
> @Riib11 I think using up a whole new keyword like `default` just for this case is a bit of an overkill. I suspect one could define the `true` constraint in existing Cryptol if they wanted to though, like this:
> 
> ```
> type constraint Default = ()
> 
> f : Default => Bit
> f = False
> ```

That's reasonable. Although for some reason `Haskell` introduces a keyword `otherwise` when you can always just use `True` instead, not sure why.

yav:
In Haskell `otherwise` is not a keyword, it's just defined in the Prelude as `True`.  We could add something like `Default` above to the Cryptol prelude too...  I am hesitant to do so at the moment because my hunch is that with the current semantics (i.e., where the previous cases are not implicitly negated) having a fall through case would be more confusing than useful, but perhaps there's some examples?  I guess it could work fine if the fall through case is non recursive, and you don't care about the type of the argument (as is the case in the `len` example).

src/Cryptol/ModuleSystem/InstantiateModule.hs
```diff
@@ -217,6 +217,8 @@ instance Inst Expr where
         EProofApp e               -> EProofApp (go e)
         EWhere e ds               -> EWhere (go e) (inst env ds)
 
+        EPropGuards _guards       -> EPropGuards undefined -- TODO
+
```
yav, line 221: Don't worry about this---it doesn't exist in the new module system, and we'll have to implement it again there anyway.

src/Cryptol/Parser.hs
```diff
@@ -0,0 +1,3995 @@
+{-# OPTIONS_GHC -w #-}
```
yav, line 1: This file should not be committed to the repo, it is autogenerated from `Parser.y`

yav:
@Riib11 is PR up to date?  I just tried the length example an am still getting a panic:
```
len : {n,a} (fin n) => [n] a -> Integer
len x
  | (n == 0) => 0
  | (n > 0) => 1 + len (drop`{1} x)
```
```
  Location:  [Eval] evalProp
  Message:   cannot use this as a guarding constraint: 
             n`902 > 0
```

Riib11:
Ah I see what the problem is, I tested 
```
len : {n,a} (fin n) => [n] a -> Integer
len x
  | (n == 0) => 0
  | (n >= 0) => 1 + len (drop`{1} x)
```
which works because `>=` is a built-in numeric type constraint. But `>` is not built in, so I need to do a normalization pass that expands that definition.

Riib11:
Now type aliases are handled in prop guards. See `tests/constraint-guards/Len.cry` for an example of using `n > 0`, where `x > y` is a type alias for `x >= y + 1`.

yav (changes requested):
Hi, I added some smaller comments inline.  I haven't had a chance to look at the two main big changes though, namely Infer and ExpandPropGuards, but figured these would be good to get started with.

src/Cryptol/Parser.y
Riib11: I got frustrated and gave up on it for a while, but I just tried again and it seems to be fixed now! Weird, I'm not sure what I was missing before

Riib11:
@yav @robdockins So, it turns out that I actually was able to write an exhaustive checker after all that even handles conjunction prop. The way it works is that when it gets to a conjunction prop, it branches on the case of each conjunct individually being false. And since this works, it got me thinking that we could do the same thing for typechecking as well. For example, you currently need to write
```
tail : {n, a} (fin n) => [n]a -> [(max n 1) - 1]a
tail xs | n == 0 => xs
        | n >= 1 => drop `{1} xs
```
Because the typechecker doesnt know that in the second guard that `n /= 0` which implies `n >= 1`. If we did that same thing as the exhaustive checker, we could just write
```
tail : {n, a} (fin n) => [n]a -> [(max n 1) - 1]a
tail xs | n == 0 => xs
        | () => drop `{1} xs
```
This is more useful of course when you have more constraints going on e.g. you could write `merge` as just
```
merge : {m,n,a} Cmp a => [m]a -> [n]a -> [m+n]a
merge xs ys
  | m == 0  => ys
  | n == 0  => xs
  | () => if x <= y 
                        then [x] # merge restX ys
                        else [y] # merge xs restY
                      where
                      [x] # restX = xs
                      [y] # restY = ys
```
The reservation I can imagine is that this complicates typechecking, since subsequent guard constraints additionally have the negations of the previous guard constraints as assumptions, and if a previous guard constraint was a conjunction then this entails multiple typecheckings of the subsequent guards (each typecheck pass assuming one of the props in the guard is false).

yav:
I think that's a good idea, but the main issue is that it could lead to code duplication in some cases, probably quite a lot if you tried, although I am not sure how much in practice.   I am think of situations like that:
```
f : ...
f x
  | (P,Q) => e1
  | ()       => e2
```
under this new system would become:
```
f : ...
f x
  | (P,Q)            => e1
  | not P, Q       => e2
  | P, not Q       => e2
  | not P, not Q => e2
```
And these would multiply if there were multiple guards with conjunctions.   The reason you need the repeated expressions is that when you type check a source expression, the assumptions that are in scope might affect how the expression is elaborated in the IR.  

I'd say let's make a separate ticket for this to discuss various ways in which we could add some such feature, but let's work on it once we've merged the base functionality into `master` as we have quite a few PRs that need merging and I want to get them in without too much re-merging.

src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -1019,54 +1035,202 @@ checkSigB b (Forall as asmps0 t0, validSchema) = case thing (P.bDef b) of
...
+                    PGeq -> pure [TCon (PC PNeq) tys, TCon (PC PGeq) (reverse tys)]
+                    -- not fin x  <=>  x == Inf
+                    PFin | [ty] <- tys -> pure [TCon (PC PEqual) [ty, tInf]]
+                         | otherwise -> panicInvalid
+                    -- not True  <=>  0 == 1
+                    PTrue -> pure [TCon (PC PEqual) [tZero, tOne]]
+                    _ -> mempty
+                  TC _tc -> panicInvalid
+                  TF _tf -> panicInvalid
+                  TError _ki -> panicInvalid -- TODO: where does this come from?
```
yav, line 1113: `TError` results from an "impossible" term (e.g., `1-2`).  It acts kind of like an exception, and when you encounter it you should just propagate it (i.e., `pNotNumeric (TError x) = TError x`).

src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -1019,54 +1035,202 @@ checkSigB b (Forall as asmps0 t0, validSchema) = case thing (P.bDef b) of
...
+              -- Try to prove that the first guard will be satisfied. If cannot,
+              -- then assume it is false (via `negateSimpleNumProps`) and try to
+              -- prove that the second guard will be satisfied. If cannot, then
+              -- assume it is false, and so on. If the last guard cannot be
+              -- proven in this way, then issue a `NonExhaustivePropGuards`
+              -- warning. Note that when assuming that a conjunction of multiple
+              -- constraints is false, this results in a disjunction, and so
+              -- must check exhaustive under assumption that each disjunct is
+              -- false independently. 
+              --
```
yav, line 1146: This is the derivation I came up with.  For a declaration:
```
AS =>

  | P, Q => ..
  | R, S => ..
  | T, U => ..
```
We'd like to show that:
```
AS => (P && Q) || (R && S) || (T && U)
```
This is equivalent to:
```
AS && !(P&&Q) && !(R&&S) => T && U
AS && (!P || !Q) && (!R || !S) => T && U
```
Now you have to consider how each assumed disjunction could be true, in this example this requires 4 proofs:
```
   AS && !P && !R => T && U
   AS && !P && !S => T && U
   AS && !Q && !R => T && U
   AS && !Q && !S => T && U
```
In the first step we have some freedom to choose which of the disjuncts (i.e., guards) to move to the LHS.   It is best to move the ones with fewest conjunctions in the guard, as this would result in the fewest proof obligations.

yav (changes requested):

src/Cryptol/TypeCheck/Infer.hs
Riib11: Hmm, that is a more efficient way of checking than what I implemented, which is formulated as the following:
```
A => (P /\ Q) \/ (R /\ S) \/ (T /\ U)

<=>

(
  (A => P /\ Q)
) 
\/
( 
  (A /\ -P => R /\ S) \/ 
  (A /\ -Q => R /\ S) 
) 
\/
(
  (A /\ -P /\ -R => T /\ U) /\ 
  (A /\ -P /\ -S => T /\ U) /\ 
  (A /\ -Q /\ -R => T /\ U) /\
  (A /\ -Q /\ -S => T /\ U)
)
\/
(
  (A /\ -P /\ -R /\ -T => True) /\ 
  (A /\ -P /\ -R /\ -U => True) /\ 
  (A /\ -P /\ -S /\ -T => True) /\ 
  (A /\ -P /\ -S /\ -U => True) /\ 
  (A /\ -Q /\ -R /\ -T => True) /\
  (A /\ -Q /\ -R /\ -U => True) /\
  (A /\ -Q /\ -S /\ -T => True) /\
  (A /\ -Q /\ -S /\ -U => True)
)
```

Riib11: I'll try implementing it the more efficient way, as you formulated

robdockins:
I haven't had a chance to review this fully yet, but I notice the sampling code is producing a lot of warnings (mostly unused imports, looks like).  We try to keep the Cryptol codebase fairly warning-safe, so I'd like to see as many of those squashed as possible.

src/Cryptol/Transform/Specialize.hs
```diff
@@ -102,6 +104,13 @@ specializeExpr expr =
     EProofAbs p e -> EProofAbs p <$> specializeExpr e
     EProofApp {}  -> specializeConst expr
     EWhere e dgs  -> specializeEWhere e dgs
+    -- The type should be monomorphic, and the guarded expressions should
+    -- already be normalized, so we just need to choose the first expression
+    -- that's true.
+    EPropGuards guards _schema -> 
+      case List.find (all checkProp . fst) guards of
+        Just (_, e) -> specializeExpr e
+        Nothing -> fail "No guard constraint was satisfied"
```
yav, line 114: This probably shouldn't fail, but rather emit an expression that will raise an exception when evaluated (e.g., `error`)

src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -52,18 +57,21 @@ import           Cryptol.TypeCheck.FFI.FFIType
...
 import           Data.Maybe(isJust, fromMaybe, mapMaybe)
-import           Data.List(partition)
 import           Data.Ratio(numerator,denominator)
 import           Data.Traversable(forM)
 import           Data.Function(on)
-import           Control.Monad(zipWithM,unless,foldM,forM_,mplus,when)
+import           Control.Monad(zipWithM, unless, foldM, forM_, mplus, zipWithM,
+                               unless, foldM, forM_, mplus, when)
+import           Data.Bifunctor (Bifunctor(second))
+import qualified Data.List as List
```
yav, line 74: Apparently this import is redundant?  Just looking at the warnings below.

src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -973,25 +982,33 @@ checkMonoB b t =
                      , dDoc = P.bDoc b
                      }
 
+    P.DPropGuards _ ->
+      tcPanic "checkMonoB"
+        [ "Used constraint guards without a signature, dumbwit, at "
+        , show . pp $ P.bName b ]
+
```
yav, line 989: Drop the "dumbwit" :)

src/Cryptol/Parser/ExpandPropGuards.hs
```diff
@@ -0,0 +1,127 @@
...
+
+-- | Error
+data Error = NoSignature (Located PName)
+  deriving (Show,Generic, NFData)
+
+instance PP Error where
+  ppPrec _ err = case err of
+    NoSignature x -> 
+      text "At" <+> pp (srcRange x) <.> colon <+>
+      text "No signature provided for declaration that uses PROPGUARDS."
```
yav, line 53: Perhaps rephrase as:

Declarations using type level guards require a type signature.

src/Cryptol/Parser.y
```diff
@@ -302,11 +302,22 @@ mbDoc                   :: { Maybe (Located Text) }
...
+propguards_quals :: { [Prop PName] }
+  : type {% fmap thing (mkProp $1) }
 
 decl                    :: { Decl PName }
   : vars_comma ':' schema  { at (head $1,$3) $ DSignature (reverse $1) $3   }
   | ipat '=' expr          { at ($1,$3) $ DPatBind $1 $3                    }
   | '(' op ')' '=' expr    { at ($1,$5) $ DPatBind (PVar $2) $5             }
+  | var apats_indices propguards_cases
+                           { mkIndexedPropGuardsDecl $1 $2 $3 }
   | var apats_indices '=' expr
```
yav, line 321: It might be a bit of a corner case, but it probably makes sense to allow guarded constants.  For example:
```

isZero : {n}  Bit
isZero
  | n == 0 => True
  | ()     => False
```

src/Cryptol/Parser/ExpandPropGuards.hs
```diff
@@ -0,0 +1,127 @@
...
+        where mu decl = Decl $ topLevelDecl { tlValue = decl }
+      f topDecl = pure [topDecl]
+
+instance ExpandPropGuards [Decl PName] where 
+  expandPropGuards decls = concat <$> traverse f decls 
+    where 
+      f (DBind bind) = fmap DBind <$> expandPropGuards [bind]
+      f decl = pure [decl]
+
+instance ExpandPropGuards [Bind PName] where
```
yav, line 79: `f` and `g` are not great names for the local functions

src/Cryptol/Parser/ExpandPropGuards.hs
```diff
@@ -0,0 +1,127 @@
...
+          Forall params props t rng <-
+            case bSignature bind of
+              Just schema -> pure schema 
+              Nothing -> Left . NoSignature $ bName bind
+          let
+            g :: ([Prop PName], Expr PName) -> ExpandPropGuardsM (([Prop PName], Expr PName), Bind PName)
+            g (props', e) = do
+              bName' <- newName (bName bind) props'
+              -- call to generated function
+              let e' = foldr EApp (EVar $ thing bName') (patternToExpr <$> bParams bind)
```
yav, line 94: I think here you need to also apply the type arguments.  To see why, consider the following example:
```
isZero : {n} () -> Bit
isZero x
  | n == 0 => True
  | ()     => False
```
If you get a warning about "unused `n`" which is is not right.  Also, if you load with `:set debug=on` (it shows you the IR), you get the following:
```Main::isZero : {n} () -> Bit
Main::isZero =
  \{n} ->
    (propguard | n`903 ==
    0 => Main::isZero(n == 0) 0 <> |  => Main::isZero() 0)
```
so there seems to be some confusion with the quantifiers

src/Cryptol/TypeCheck/AST.hs
```diff
@@ -267,6 +269,11 @@ instance PP (WithNames Expr) where
                          [ ppW e
                          , hang "where" 2 (vcat (map ppW ds))
                          ]
+      
+      EPropGuards guards _ ->
```
yav, line 273: Probably more readable to pretty print as:
```
propguards
  | P => e1
  | Q => e2
```

(i.e., `vcat` instead of `hsep` and indent under the `propguards`)

yav (changes requested):

src/Cryptol/Parser.y
Riib11: that's quite an interesting idea

Riib11: However, it seems that there is an error that will arise from ambiguous type variables. For example:
```
isFin : {n} Bit
isFin | fin n => True
      | () => False
```
yields the error
```
[error] at tests/constraint-guards/constant.cry:2:1--2:6:
  Ambiguous numeric type: type argument 'n' of 'Main::isFin(fin n)'
```
Is that something you'd want to have check for an appearance of the type variable in the constraint guards before raising?

Riib11: Actually, you can see from the name of the function that raises the error `sFin(fin n)`, that its one of the declarations generated by expanding the prop guards. It doesn't know locally, currently, that itself came from a prop guard expansion

src/Cryptol/Transform/Specialize.hs
Riib11: I defined an `eError'` that is the same as `eError` but doesn't explicitly apply a type, and used it here

src/Cryptol/Parser/ExpandPropGuards.hs
Riib11: Ah yes, I see the problem. Usually it works when it can infer the right types from the arguments passed, but in this case it can't

src/Cryptol/Parser.y
Riib11: actually, this is fixed by doing [your other request](https://github.com/GaloisInc/cryptol/pull/1380#discussion_r958026018). Now it just gives warnings about unused type variables

yav (changes requested):
I think we are almost there, just fix up the `error` stuff

PR #1389, opened 2022-08-03
Numeric literals sampling

Riib11:
# The current problem with `:check`ing functions that are polymorphic in numeric types

When `check`ing a polymorphic function, Cryptol chooses default instantiations of the parametrized type variables that satisfy the function's type constraints. There is a system of rules for choosing default non-numeric type instantiations according to the set of type classes that type variable is constrained by. For numeric type variables, a default instance is found by asking the SMT to try and solve the numeric constraints and then give back a single solving substitution.

For testing functions that are polymorphic in numeric types, this method of default instantiation is 
unsatisfactory, because the SMT solver only provides a single solution, and that solution is usually trivial (i.e. the smallest literals that satisfy the constraints). For example:

```
module Test where

map_add1 : {n} (fin n) => [n][8] -> [n][8]
map_add1 xs = [ x + 1 | x <- xs ]

property clearly_unsatisfied xs = xs == map_add1 xs 
```
```
Cryptol> :l Test
Test> :check clearly_unsatisfied
  * Using '0' for signature variable 'n'
Using exhaustive testing.
Passed 1 tests.
Q.E.D.
```

Of course, `clearly_unsatisfied` is clearly unsatisfied, since adding `1` to all the elements of a sequence is not the identity. But, since the SMT solver found `n = 0` to be a solution that satisfies `fin n`, it just tests that and finds that `[] = []` indeed.

# New feature: sampling over solution substitutions of the numeric types

This PR implements a new feature that, when turned on, replaces defaulting when getting a concrete type for a function during `:check`ing.  It can be turned on via
```
Cryptol> :set literal-sampling = on
Cryptol> :set literal-sampling-bin-size = X
```
where `X` is the number of tests to run for each sampled substitution of literals (the default is 10).

The literal sampler algorithm takes into account the constraints over numeric literals in the type signature of the function being `:check`ed, and tries to sample with a certain distribution over the solutions. Only a certain domain of constraints are currently supported. Let `a, b, c, ...` be numeric type variables and `x, y, z, ...` be constant numeric literals. Then the following subset of the cryptol type constraint syntax is allowed to mention numeric type variables in a function in order to be accepted by the literal sampling algorithm:
- `fin _`
- `_==_`, `_ <= _`, `_ >= _`
- `_ + _`, `_ - _`
- `a * x`, `a % x`, `a / x`
- `x * y`, `x / y`, `x % y`, `x ^^ y`

If a variable is only bounded above by `inf`, then it is sampled along an exponential distribution up to a maximum value, where higher values are less likely to be sampled. If a variable is bounded above by a constant, then it is sampled on a uniform distribution up to that upper bounded. The order of variable sampling _should not_ affect the distribution (I'm still working out how true this is).

Once a solution substitution is sampled, the instantiation of the function type with that substitution is `:check`ed as if it was a normal top-level function, except for only `literal-sampling-bin-size` number of tests. Then another solution is sampled and `:check`ed, etc, until the the total number of tests performed is at least `tests`.

# Implementation Details

TODO


robdockins:
I haven't had a chance to review this fully yet, but I notice the sampling code is producing a lot of warnings (mostly unused imports, looks like). We try to keep the Cryptol codebase fairly warning-safe, so I'd like to see as many of those squashed as possible.

src/Cryptol/Eval/Concrete.hs
```diff
@@ -520,3 +520,9 @@ updateBack_word (Nat n) _eltTy bs (Left idx) val = do
 updateBack_word (Nat n) _eltTy bs (Right w) val = do
   idx <- bvVal <$> asWordVal Concrete w
   updateWordValue Concrete bs (n - idx - 1) (fromVBit <$> val)
+
```
robdockins, line 523: dead code?

src/Cryptol/REPL/Command.hs
```diff
@@ -391,32 +399,107 @@ dumpTestsCmd outFile str pos fnm =
...
+          let doc = fixNameDisp nd (pp texpr)
+          go doc Nothing texpr schema
+    _ -> do
+      expr <- replParseExpr str pos fnm
+      (_,texpr,schema) <- replCheckExpr expr
+      nd <- M.mctxNameDisp <$> getFocusedEnv
+      let doc = fixNameDisp nd (ppPrec 3 expr) -- function application has precedence 3
+      go doc (Just expr) texpr schema
+  where
+    go :: Doc -> Maybe (P.Expr P.PName) -> T.Expr -> T.Schema -> REPL ()
```
robdockins, line 439: This is too big to be in a `where`, I think.  Let's make it a new top-level definition, and do less rearrangement of the original function.

src/Cryptol/REPL/Command.hs
```diff
@@ -391,32 +399,107 @@ dumpTestsCmd outFile str pos fnm =
...
+        void (qcExpr qcMode doc texpr schema testNum val ty Nothing)
+      QCRandom -> do
+        testNum <- (toInteger :: Int -> Integer) <$> getKnownUser "tests"
+        useLitSampling <- getKnownUser "literalSampling" :: REPL Bool
+        if  useLitSampling &&
+            not (null (T.sVars schema)) &&
+            all ((T.KNum ==) . T.tpKind) (T.sVars schema)
+          then do
+          case mb_expr of
+            Nothing -> do
```
robdockins, line 453: Not sure I understand this case. Why do we only do the sampling in the `Just` case?

src/Cryptol/REPL/Command.hs
```diff
@@ -391,32 +399,107 @@ dumpTestsCmd outFile str pos fnm =
...
+              let bins = testNum `div` binSize
+              io (sampleLiterals schema (fromInteger bins)) >>= \case
+                Right samples_schemas -> do
+                  rPutStrLn "Using literal sampling."
+                  let -- this generates a new expression that has the appropriate
+                      -- type arguments given to it according to the sampling,
+                      -- which is then typechecked and evaluated again
+                      applySampleToTExpr :: Sample -> P.Expr P.PName -> P.Expr P.PName
+                      applySampleToTExpr sample e = P.EAppT e $ f <$> sample
+                        where
```
robdockins, line 468: This is a `where` inside a `let` inside a `case` inside an `if` inside a `where. I think we can find a way to break this up better.

src/Cryptol/REPL/Command.hs
```diff
@@ -425,27 +508,50 @@ data TestReport = TestReport
...
-  do (val,ty) <- replEvalCheckedExpr texpr schema >>= \mb_res -> case mb_res of
-       Just res -> pure res
-       -- If instance is not found, doesn't necessarily mean that there is no instance.
-       -- And due to nondeterminism in result from the solver (for finding solution to 
-       -- numeric type constraints), `:check` can get to this exception sometimes, but
-       -- successfully find an instance and test with it other times.
-       Nothing -> raise (InstantiationsNotFound schema)
-     testNum <- (toInteger :: Int -> Integer) <$> getKnownUser "tests"
-     tenv <- E.envTypes . M.deEnv <$> getDynEnv
+qcExpr qcMode exprDoc _texpr _schema testNum val ty mb_sampleInfo =
```
robdockins, line 532: If we don't need the texpr and schema arguments anymore, let's remove them.

robdockins: This is also a good opportunity to add haddocks to this function.

src/Cryptol/TypeCheck/Default.hs
```diff
@@ -207,19 +207,24 @@ the given type.  This is useful when we do evaluation at the REPL.
 The resulting types should satisfy the constraints of the schema.
 The parameters should be all of numeric kind, and the props should als
 be numeric -}
+-- TODO: I think this also needs to return a generator for solutions, since we
```
robdockins, line 210: Is this TODO still relevant?

src/Cryptol/TypeCheck/Default.hs
```diff
@@ -207,19 +207,24 @@ the given type.  This is useful when we do evaluation at the REPL.
...
        Nothing -> return Nothing
-       Just mdl0 ->
-         do mdl <- shrinkModel sol params props mdl0
-            let su = listSubst [ (x, tNat' n) | (x,n) <- mdl ]
-            return $
-              do guard (null (concatMap pSplitAnd (apSubst su props)))
-                 tys <- mapM (bindParam su) params
-                 return (zip as tys)
+       Just mdl0 -> do
+        -- TODO: here is where the literal sampler will be invoked
```
robdockins, line 220: Is there still something TODO here?

src/Cryptol/TypeCheck/PP.hs
```diff
@@ -6,7 +6,7 @@
 -- Stability   :  provisional
 -- Portability :  portable
 
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
```
robdockins, line 9: In this, and a lot of other places, there is a bunch of noise messing around with Safe annotations and such.  Is all that really necessary?

src/Cryptol/TypeCheck/Solve.hs
```diff
@@ -148,6 +150,29 @@ addIncompatible g i =
...
+-- For the polymorphic literals in a schema, uses a custom constraint-solver to
+-- sample over the solution space. Returns a list of Schemas, where each Schema
+-- corresponds to a sampled solution applied to that Schema (i.e. the solution's
+-- substituion is applied to the Schema and the params that corresponded to the
+-- substituted variables are omitted). If there are no polymorphic literals,
+-- then Nothing.
+-- TODO: explain sampling distributions
+sampleLiterals :: Schema -> Int -> IO (Either String [(Sampling.Sample, Schema)])
+sampleLiterals schema@Forall {sVars, sProps} nLiteralSamples = do
+  -- let (literalVars, otherVars) = partition ((KNum ==) . tpKind) sVars
```
robdockins, line 162: dead code?

src/Cryptol/TypeCheck/Solver/InfNat.hs
```diff
@@ -23,16 +24,33 @@ import Control.DeepSeq
...
+    Nat n -> show n
+    Inf -> "inf"
 
 fromNat :: Nat' -> Maybe Integer
 fromNat n' =
   case n' of
     Nat i -> Just i
     _     -> Nothing
 
+instance Num Nat' where
```
robdockins, line 40: Realllllly not sure I'm a fan of this instance.  How necessary is it? `negate` and `fromInteger` are just wrong.

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Preconstraints.hs
```diff
@@ -0,0 +1,261 @@
...
+        proc2 con ts =
+          toPExp `traverse` ts >>= \case
+            [e1, e2] -> do
+              let pprop = con e1 e2
+              pure $ addPProps [pprop] precons
+            _ -> undefined -- bad number of args
+        proc1 con ts =
+          toPExp `traverse` ts >>= \case
+            [e] -> pure $ addPProps [con e] precons
+            _ -> undefined -- bad number of args
```
robdockins, line 112: If this can't happen, `panic` instead

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Preconstraints.hs
```diff
@@ -0,0 +1,261 @@
...
+    precons
+      { preprops = preprops' <> preprops'',
+        nVars = nVars
+      }
+  where
+    normPProp :: PProp -> WriterT [PProp] (StateT Int SamplingM) PProp
+    normPProp = \case
+      PPEqual pe1 pe2 -> PPEqual <$> normPExp pe1 <*> normPExp pe2
+      PPNeq _a _b -> do
+        error "not sure how to handle Neq"
```
robdockins, line 176: Can user input trigger this case? If not, panic, if so, let's find a better way to display this error.

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Q'.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Cryptol.TypeCheck.Solver.Numeric.Sampling.Q' where
```
robdockins, line 4: What is this numeric type? Why do we need it?

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Q'.hs
```diff
@@ -0,0 +1,47 @@
...
+
+normQ' :: Q' -> Q'
+normQ' Q' {..} = Q' {q, i = signum i}
+
+instance Num Q' where
+  Q' q1 i1 + Q' q2 i2
+    | signum i1 == signum i2 = normQ' $ Q' (q1 + q2) (i1 + i2)
+    | otherwise = toQ' 0
+
+  Q' q1 i1 * Q' q2 i2 = normQ' $ Q' (q1 + q2) (i1 + i2)
```
robdockins, line 35: This seems very suspect?

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Q.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Cryptol.TypeCheck.Solver.Numeric.Sampling.Q where
```
robdockins, line 3: Why define this separately from `Rational`?

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Sampling.hs
```diff
@@ -0,0 +1,363 @@
...
+foldFind f b (a : as) = case f a b of
+  Just b' -> foldFind f b' as
+  Nothing -> Just a
+
+-- tw = totalWeight wr
+sampleWeightedRange :: WeightedRange -> Maybe Integer -> SamplingM Nat'
+sampleWeightedRange wr mb_tw = do
+  let tw = fromMaybe (totalWeight wr) mb_tw
+  i <- liftR $ randomR (0, tw)
+  let Just wri =
```
robdockins, line 127: Let's not have non-exhaustive `let` patterns.  `case` and `panic` if necessary

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Z'.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Cryptol.TypeCheck.Solver.Numeric.Sampling.Z' where
```
robdockins, line 4: I am again mystified by this type. What's going on with it?

robdockins (changes requested):
Mostly code style stuff here.

I haven't fully understood what the sampling algorithm is doing, but the steps looks well-documented. I'd definitely like to better understand what's going on with the `Q'` and `Z'` types.

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Q'.hs
Riib11: turns out not needed

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Q.hs
Riib11: it started off looking a bit different, but actually yeah this is just `Rational`. I'll refactor it away

Riib11:
I will write out a description of the algorithm, to make it more clear as well

Riib11:
> Mostly code style stuff here.
> 
> I haven't fully understood what the sampling algorithm is doing, but the steps looks well-documented. I'd definitely like to better understand what's going on with the `Q'` and `Z'` types.

Here's a high-level description of the entire sampling algorithm: https://github.com/GaloisInc/cryptol/blob/73dee71f58f1a5e6c55697d9efddcd384b363ed8/src/Cryptol/TypeCheck/Solver/Numeric/Sampling.hs#L26

src/Cryptol/TypeCheck/PP.hs
Riib11: I was getting a bunch of warnings from HLS, so that's why I did it

src/Cryptol/TypeCheck/Solver/InfNat.hs
Riib11: The reason this was so useful to me is because I defined `Exp` to be polymorphic in its domain (which I use as either `Nat'` or `Rational`), and so then defined a bunch of operations over `Exp` that required a `Num` instance

src/Cryptol/TypeCheck/Solver/Numeric/Sampling/Constraints.hs
```diff
@@ -0,0 +1,166 @@
...
+--
+-- Encodes the following:
+-- - `sys`: a system of equations
+-- - `tcs`: a list of non-arithmetic typeclass constraints
+-- - `tparams`: a vector of the numeric type parameters from a signature
+data Constraints a = Constraints
+  { sys :: System a,
+    tcs :: [Tc a],
+    tparams :: Vector TParam,
+    nVars :: Int
```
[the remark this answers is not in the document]
Riib11: I strongly expect no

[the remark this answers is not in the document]
Riib11: At most, the number of vars can be ~twice the number of numeric type var params

PR #1401, opened 2022-08-13
Print message in REPL when loading dynamic library for FFI

### LINKED ISSUE #1399: REPL output when loading shared library
Make the REPL print out which shared library it is loading when loading a module containing `foreign` declarations.

qsctr:
Also refactored some of the loading code to allow this to work nicely.

Fixes #1399.


yav (approved):
This looks good to me

PR #1403, opened 2022-08-15
CI: Use Ubuntu 22.04, drop 18.04

RyanGlScott:
GitHub Actions has deprecated its Ubuntu 18.04 runners, and they will be removed by December 1, 2022. Moreover, GitHub Actions now offers Ubuntu 22.04 runners.  It seems like a good time to upgrade our CI accordingly.

Somewhat annoyingly, the `haskell` Docker images that we use in our Dockerfiles use such an old version of Debian that their version of `glibc` is incompatible with any of the `what4-solvers` built for Ubuntu 20.04 or 22.04. As a result, I switched them from the `haskell` Docker image to the `ubuntu` one. This required some minor tweaks to how dependencies are installed, but nothing too serious.


RyanGlScott:
I also fired off a workflow-dispatched job [here](https://github.com/GaloisInc/cryptol/actions/runs/2860981825), which ensures that the nightly Docker-based jobs continue to build.

kquick (commented):
That glibc issue looks like it was annoying and had an unfortunately large impact on this changeset.  Thanks for getting this taken care of!

PR #1406, opened 2022-08-17
Support nested sequences in FFI

### LINKED ISSUE #1404: Nested sequences for FFI
Support nested sequences by flattening them and storing the elements contiguously, the same way C represents nested arrays.

Note: If the inner sequence is a sequence of bits (i.e. a word), the current representation of packing it into a machine word type still takes priority.
```
[4][8]Bit   -> array of 4 byte
[4][8]Float -> array of 32 float
```

qsctr:
Closes #1404.

Note: this PR is on top of #1401, please merge that one first.


src/Cryptol/Eval/FFI.hs
```diff
@@ -153,13 +149,23 @@ foreignPrim name FFIFunType {..} impl tenv = buildFun ffiArgTypes []
...
-      args <- traverse (>>= m) $ enumerateSeqMap n $ fromVSeq val
-      -- Since we need to do an Eval action in an IO callback, we need to
-      -- manually unwrap and wrap the Eval datatype
+  marshalArg (FFIArray (map evalFinType -> sizes) t) val f =
+    getMarshalBasicArg t \m ->
+      -- Since we need to do Eval actions in an IO callback, we need to manually
+      -- unwrap and wrap the Eval datatype
       Eval \stk ->
-        withArray args \ptr ->
+        allocaArray (fromInteger $ product sizes) \ptr -> do
```
yav, line 157: The loop here is a bit hard to follow.  I wonder if something like this (untested), might be better:
```
marshallArray [] _ _ = error "bad"

marshallArray [_] ptr v  =
  do runEval stk (m v) >>= pokeElemOff ptr
     pure (plusPtr ptr 1)       

marshallArray (_ : next) ptr v =
  do vs' <- traverse (runEval stk) $ enumerateSeqMap n $ fromVSeq v
     foldM (marshallArray next) ptr vs'
```

yav (changes requested):
I had some trouble following the loop for marshaling out arrays, so I added a comment with a suggestion that I think is more readable.  OTOH, I haven't even tried to compile it so it might be wrong :-)  So give it a go, if it works---great.  Otherwise, the changes look good to me

src/Cryptol/Eval/FFI.hs
qsctr: Yeah, I initially tried to write it in a tail-recursive way, but I realized it probably doesn't make much difference in terms of performance.

yav (approved):

PR #1410, opened 2022-08-22
Remove regression test for issue #835.

taimoorzaeem:
Fixes issue #1329.

### DIFF
--- a/tests/issues/issue835.icry
+++ b/tests/issues/issue835.icry
@@ -1,2 +1,0 @@
-:m Float
-:t (zero : Float 5 11) <$ zero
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Thanks for the pull request!

Rather than deleting the test case, I think we could just adapt it so that it actually exercises the part of the pretty printer that is responsible for parenthesizing compound expressions. Something like this would suffice:

```
Cryptol> :m Float
Loading module Cryptol
Loading module Float
Float> :t number`{1, Float _ _}
number`{1, Float _ _} :
  {n, m} (ValidFloat n m, Literal 1 (Float n m)) => Float n m
```

The fact that it is printed like `Literal 1 (Float n m)` is a sign that the fix for #835 is working.

PR #1414, opened 2022-08-25
CI: Drop GHC 8.8

RyanGlScott:
This is motivated by a couple of reasons:

* GHC 8.8 on Windows suffers from a bug that can cause building certain code to loop forever or even segfault. As far as I can tell, this bug doesn't affect GHC 8.10 or later, so dropping 8.8 is a fairly reliable way to avoid this problem.
* GHC 8.8 is outside of the three most recent stable GHC releases, so it doesn't make as much sense to keep maintaining support for it.


kquick (approved):
Considering 9.0 a "stable" release is perhaps different than considering it "adviseable", but in general, stepping away from 8.8 Windows issues in favor of stability is good.

PR #1416, opened 2022-08-25
AES-NI version of SuiteB module using FFI

### LINKED ISSUE #1398: FFI version of SuiteB AES functions
As a demo for the FFI, implement a drop-in replacement for the builtin SuiteB module that uses the FFI instead of Cryptol and Haskell primitives, and see if it results in a speedup in code like [this](https://github.com/weaversa/cryptol-course/blob/master/labs/KeyWrapping/spec/NIST_800_38F/AES128.cry).

qsctr:
This PR adds a module `SuiteB_FFI` which uses AES-NI hardware instructions on x86_64 when possible, and falls back to the Cryptol/Haskell primitive implementations in the builtin `SuiteB` module otherwise.

Closes #1398.


yav (approved):
Looks good to me.  It might also be nice to have a version that goes as fast as possible (i.e., changes the API to avoid the byte swapping, and also doesn't check if the instructions are supportd (i.e., only works if you have the hardware).   I'll make a separate thicket for that though.

Issue #1253, opened 2021-07-23
File name / module name mismatch error text


pnwamk:
TL;DR consider swapping out `Saw:` => `Actual:` in Cryptol error messages (or similar) and playing a little with the pretty-printed alignment of the `Actual` and `Expected` columns/rows.

This isn't a huge deal but might be a small improvement.

Backstory:

I got this error message running some scripts that invoke cryptol:

```
[error] :1:1--1:1:
    File name does not match module name:
    Saw: Main
    Expected: keygen_implementation
```

I now understand what this error is saying, but initially I was a little thrown and was looking for how [SAW](https://github.com/galoisinc/saw-script) related to the particular setup I had. Perhaps this is a one-off and I just need a cup of coffee, but if others find this potentially confusing, and given how often these two tools more and more are used either in conjunction or at least at different times by the same people, it might be nice to _slightly_ tweak the wording (and perhaps even the indentation/alignment?), e.g.:

```
[error] :1:1--1:1:
    File name does not match module name:
        Actual: Main
      Expected: keygen_implementation
```

It may also be worth noting that the [keygen_implementation](https://github.com/GaloisInc/BLST-Verification/blob/0c892afebc3bb26774a02c96795d3e9722915b12/spec/keygen_implementation.cry) file has no _specified_ module name, which is why it defaults to `Main`, but the error message doesn't allude to that at all.

RyanGlScott:
Fixed in #1419.

PR #1419, opened 2022-08-28
Improve error message for ModuleNameMismatch error and add regression test.

taimoorzaeem:
Fixes issue #1253.


RyanGlScott:
Thanks, @taimoorzaeem.

The CI is failing on Windows because the error message is printing out a relative path, which displays as `[error] ./issue1253b.cry:1:8--1:11:` on macOS and Linux but `[error] .\issue1253b.cry:1:8--1:11:` on Windows (note the backslash instead of the forward slash). You should be able to resolve this issue by checking in a separate `tests/issues/issue1253.icry.stdout.mingw32` file that uses a backslash instead of a forward slash, which the test suite will use only on Windows.

RyanGlScott (approved):

PR #1420, opened 2022-08-29
Add `:time` command

### LINKED ISSUE #1415: `:time` command for measuring evaluation time
We can add a `:time` command to measure how long it takes to evaluate an expression. We might be able to use something like [criterion-measurement](https://hackage.haskell.org/package/criterion-measurement) for accurate timings over many iterations. This would be useful for benchmarking Cryptol code.

Note that laziness might result in the first iteration of evaluating a top-level definition being slower, so maybe the user should be careful to pre-evaluate certain things if they don't want that to be part of the measurement.

qsctr:
Adds a `:time` command in the REPL for measuring the evaluation time of an expression. Uses [criterion-measurement](https://hackage.haskell.org/package/criterion-measurement) to do measurements.

The analysis that we do is very crude compared to what the full `criterion` package does (we basically just calculate the average times), but `criterion` pulls in much more dependencies so I don't think it's worth adding it.

Closes #1415.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@
 * Fix a bug in the What4 backend that could cause applications of `(@)` with
   symbolic `Integer` indices to become out of bounds (#1359).
 
+## New features
+
+* Add a `:time` command to benchmark the evaluation time of expressions.
+
 # 2.13.0
 
 ## Language changes
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -119,6 +119,7 @@ library
                        Cryptol.Utils.Patterns,
                        Cryptol.Utils.Logger,
                        Cryptol.Utils.Benchmark,
+                       Cryptol.Utils.Types,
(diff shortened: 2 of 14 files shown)


yav:
@qsctr could you add an entry to the CHANGES.md to record that we now have this.   I forgot to mention it when reviewing the FFI PR, but we should an entry for the FFI there as well.

yav (approved):
I think it looks good.  The only idea I had (but I am not sure it is really needed) is that it might be nice if the `:time` command binds the `it` variable with the result of the timing (e.g., using `bindItVariableVal`.  Not sure if it should bind a record, or just a single number, or if this is needed at all, but my thinking was that if we have that  (and also added some sort of "quiet" mode where the timings are just bound to `it` but not printed), then we could try to write benchmarking test cases checking that the times are within some range

yav (approved):
LGTM

PR #1429, opened 2022-09-08
Generate C header files from `foreign` declarations

### LINKED ISSUE #1396: Generate FFI header files
Automatically generate C header files from `foreign` declarations so the user doesn't have to manually translate the types and possibly get it wrong resulting in undefined behavior.

qsctr:
Add a new command `:generate-foreign-header` that parses and typechecks the given Cryptol file, finds the `foreign` declarations, translates them to C function declarations, and writes them to a header file.

Closes #1396.


RyanGlScott:
Rebasing on top of #1430 should fix the Windows CI failures.

PR #1438, opened 2022-09-20
Support FFI on Windows

### LINKED ISSUE #1394: Windows support for FFI
In #1376 we held off on Windows support for FFI due to issues with packaging `libffi` on Windows detailed in https://github.com/GaloisInc/cryptol/issues/1090#issuecomment-1178318186. But now that we have static linking of `libffi` on all platforms with [`libffi-0.2`](https://hackage.haskell.org/package/libffi-0.2), Windows support should just be a matter of using the [appropriate functions from the `Win32` package](https://hackage.haskell.org/package/Win32-2.13.2.1/docs/System-Win32-DLL.html#v:loadLibrary) to dynamically load shared libraries. We also need to figure out how to make the FFI tests work on Windows as well.

RyanGlScott:
This patch:

* Adds the appropriate conditional logic to use the `Win32` library to dynamically load shared libraries on Windows.
* Tweaks some FFI-related test cases to make them work portably on Windows. I have left comments describing each of the non-obvious tweaks that I had to make.
* Updates the reference manual accordingly.

Fixes #1394.


yav (approved):
Looks good to me.   I just merged in the FFI changes to add support for Integer and friends using GMP, so the only thing I can think of would be to update the tests to work with that---the Makefile in the `tests/ffi` directory now also uses LDLIBS which passes `-lgmp` and the output file for windows would have to be updated to add the few extra lines from the Linux/Mac tests

Issue #1441, opened 2022-09-23
Add support for restructured text (.rst) code blocks


podhrmic:
An example below:

```rst
Lorem Ipsum
=========================

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum

.. code-block:: cryptol

   R1 : {WIDTH} (fin WIDTH, WIDTH >= 2, WIDTH % 2 == 0) => [WIDTH/2-1] -> Integer
   R1 nhi = `(2^^WIDTH)

   R4 : {WIDTH} (fin WIDTH, WIDTH >= 2, WIDTH % 2 == 0) => [WIDTH/2-1] -> Integer
   R4 nhi = `(2^^(WIDTH+2))
```

podhrmic:
Use pandoc module.

podhrmic:
@yav closing as no longer relevant. .rst is a relatively complex format, so a simple parser similar to what we have for Markdown and Latex wouldn't cut it. A better way is probably to convert .rst code blocks to cryptol separately, and then directly load cryptol files.

yav:
I think this would be a nice feature to have and extracting Cryptol code blocks from an RST really shouldn't be that hard.  Can you say a bit more about what difficulties you envision?

podhrmic:
Of course. [Here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) are the specs for the code-block directive in rst.

An example:
```
.. code-block:: cryptol
   :linenos:
   :dedent: 4

       some cryptol code
       
       some other cryptol code

Regular text here. Code block ended on the previous line.
```       

The trick is you have to track the opening token `.. code-block:: cryptol` and correctly identify the ending newline. So probably some state tracking? Anyway, I agree that getting something that works shouldn't be that hard, but given my Haskell skills it is beyond the time I have available for this feature.

yav:
@podhrmic I implemented a simple RST pre-processor in PR #1443.  Give it a go!

PR #1443, opened 2022-09-27
Add support for literat Cryptol with RST

yav:
This implements supports for literate Cryptol with RST, see #1441

### DIFF
--- a/src/Cryptol/Parser/Unlit.hs
+++ b/src/Cryptol/Parser/Unlit.hs
@@ -8,7 +8,7 @@
 --
 -- Convert a literate source file into an ordinary source file.
 
-{-# LANGUAGE OverloadedStrings, Safe, PatternGuards #-}
+{-# LANGUAGE OverloadedStrings, PatternGuards #-}
 module Cryptol.Parser.Unlit
   ( unLit, PreProc(..), guessPreProc, knownExts
   ) where
@@ -20,7 +20,7 @@ import           System.FilePath(takeExtension)
 
 import           Cryptol.Utils.Panic
 
-data PreProc = None | Markdown | LaTeX
+data PreProc = None | Markdown | LaTeX | RST
 
 knownExts :: [String]
 knownExts  =
@@ -28,6 +28,7 @@ knownExts  =
   , "tex"
   , "markdown"
   , "md"
+  , "rst"
   ]
 
 guessPreProc :: FilePath -> PreProc
@@ -35,6 +36,7 @@ guessPreProc file = case takeExtension file of
                       ".tex"      -> LaTeX
                       ".markdown" -> Markdown
                       ".md"       -> Markdown
+                      ".rst"      -> RST
                       _           -> None
 
 unLit :: PreProc -> Text -> Text
@@ -47,6 +49,7 @@ preProc p =
     None     -> return . Code
     Markdown -> markdown
     LaTeX    -> latex
+    RST      -> rst
 
 
 data Block = Code [Text] | Comment [Text]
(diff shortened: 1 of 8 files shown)


src/Cryptol/Parser/Unlit.hs
```diff
@@ -8,7 +8,7 @@
 --
 -- Convert a literate source file into an ordinary source file.
 
-{-# LANGUAGE OverloadedStrings, Safe, PatternGuards #-}
+{-# LANGUAGE OverloadedStrings, PatternGuards #-}
```
RyanGlScott, line 11: Is the code in this module no longer `Safe`?

src/Cryptol/Parser/Unlit.hs
```diff
@@ -134,6 +137,34 @@ latex = comment []
   isBeginCode l = "\\begin{code}" `Text.isPrefixOf` l
   isEndCode l   = "\\end{code}"   `Text.isPrefixOf` l
 
-
+rst :: [Text] -> [Block]
+rst = comment []
+  where
+  isBeginCode l = case filter (not . Text.null) (Text.splitOn " " l) of
```
RyanGlScott, line 143: I wonder if this should use `Text.split isSpace` instead, both for consistency with `isEmpty`/`isCode` and to catch other whitespace characters.

src/Cryptol/Parser/Unlit.hs
```diff
@@ -134,6 +137,34 @@ latex = comment []
...
+                    Just (c, _) -> isSpace c
+                    Nothing     -> True
+
+  comment acc ls =
+    case ls of
+      [] -> mk Comment acc
+      l : ls1 | isBeginCode l -> codeOptoins (l : acc) ls1
+              | otherwise     -> comment (l : acc) ls1
+
+  codeOptoins acc ls =
```
RyanGlScott, line 158:
```suggestion
  codeOptions acc ls =
```

src/Cryptol/Parser/Unlit.hs
```diff
@@ -134,6 +137,34 @@ latex = comment []
...
+
+  isEmpty       = Text.all isSpace
+  isCode l      = case Text.uncons l of
+                    Just (c, _) -> isSpace c
+                    Nothing     -> True
+
+  comment acc ls =
+    case ls of
+      [] -> mk Comment acc
+      l : ls1 | isBeginCode l -> codeOptoins (l : acc) ls1
```
RyanGlScott, line 155:
```suggestion
      l : ls1 | isBeginCode l -> codeOptions (l : acc) ls1
```

src/Cryptol/Parser/Unlit.hs
```diff
@@ -134,6 +137,34 @@ latex = comment []
...
+              | otherwise     -> comment (l : acc) ls1
+
+  codeOptoins acc ls =
+    case ls of
+      [] -> mk Comment acc
+      l : ls1 | isEmpty l -> mk Comment (l : acc) ++ code [] ls1
+              | otherwise -> codeOptoins (l : acc) ls1
```
RyanGlScott, line 162:
```suggestion
              | otherwise -> codeOptions (l : acc) ls1
```

src/Cryptol/Parser/Unlit.hs
```diff
@@ -134,6 +137,34 @@ latex = comment []
   isBeginCode l = "\\begin{code}" `Text.isPrefixOf` l
   isEndCode l   = "\\end{code}"   `Text.isPrefixOf` l
 
-
+rst :: [Text] -> [Block]
+rst = comment []
+  where
+  isBeginCode l = case filter (not . Text.null) (Text.splitOn " " l) of
+                    ["..", "code::", "cryptol"] -> True
```
RyanGlScott, line 144: I wonder if we should recognize `code-block::` and `sourcecode::` in addition to `code::`? The first two are [officially recognized directives](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) in Sphinx.

In fact, `code` is technically a [different thing altogether](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-code), although I don't know if strict adherence to Sphinx's rules is a goal for Cryptol.

src/Cryptol/Parser/Unlit.hs
yav: It is still Safe.   I think we've been using `Safe` incorrectly, probably due to a historical misunderstanding on my part about what the pragma means, so I thought I might as well start fixing things slowly.  In particular, GHC can already infer which modules are safe and which are not, without the pragma.  Instead,   the pragma explicitly asks GHC to *check* if a module is safe, and complain if it is not.  So I don't think we need to annotate every single module with `Safe`, but rather we should add it to modules where we want this checked. We currently don't really have a use case for this feature, so I am not sure which modules should have the annotation...

src/Cryptol/Parser/Unlit.hs
yav: Good idea, will change.

src/Cryptol/Parser/Unlit.hs
yav: You are quite right, I was reading the docs for `code-block` and I guess auto-corrected it `code` :)

src/Cryptol/Parser/Unlit.hs
RyanGlScott: Ah, I see. I don't have too strong of an opinion about `Safe` myself (if anything, I am slightly biased against it), so I'm happy to defer to your judgment about whether they should be kept or removed.

podhrmic:
@yav I tested it and it correctly parses all rst examples I have available. Thanks for putting it together!

PR #1464, opened 2022-10-26
Copy the syntax-highlight directory under share/cyrptol for installation

yav:
Add syntax highlighting to installation script

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -102,6 +102,7 @@ bundle_files() {
   cp docs/*pdf $doc
   mkdir -p $lib
   cp -r lib/* $lib
+  cp -r syntax-highlight $lib
 
   # Copy the two interesting examples over
   cp docs/ProgrammingCryptol/{aes/AES,enigma/Enigma}.cry $doc/examples/


.github/ci.sh
```diff
@@ -102,6 +102,7 @@ bundle_files() {
   cp docs/*pdf $doc
   mkdir -p $lib
   cp -r lib/* $lib
+  cp -r syntax-highlight $lib
```
RyanGlScott, line 105: This will put the syntax-highlighting–related files among the `.cry` files, which is perhaps not the most intuitive place to put them. I wonder if it would make more sense to have a separate `syntax-highlight/` subdirectory for this? Something like this, perhaps:

```
syntax_highlight=dist/share/syntax-highlight
...
mkdir -p $syntax_highlight
cp -r syntax-highlight $syntax_highlight
```

RyanGlScott (commented):
I've looked at a bindist for this PR and can confirm that the syntax highlighting files are included. I have one comment about the location of these files, but this otherwise LGTM.

.github/ci.sh
yav: Hm, I was hoping that since I don't have a slash after `syntax-highlight` it will copy the directory rather than the files in it... Did it not do that?

RyanGlScott: It does copy the directory. What I am referring to is that you have a situation like this:

```
$ ls cryptol-2.13.0.99-ubuntu-22.04-x86_64/share/cryptol/
Array.cry         Cryptol.cry       Float.cry         SuiteB.cry        
Cryptol/          CryptolTC.z3      PrimeEC.cry       syntax-highlight/
```

Perhaps this is what you intended, but I'm not sure that I would have thought to look here for the `syntax-highlight` directory. I'll let you make the call.

yav:
@RyanGlScott what do we want to do with this PR?  Should we instead modify the scripts to get syntax-highlighting from the other repo?

RyanGlScott:
> Should we instead modify the scripts to get syntax-highlighting from the other repo?

Yes, that sounds reasonable to me.

Issue #1475, opened 2022-11-30
`:set smtFile` doesn't work with online `what4`-based provers


RyanGlScott:
If you run this script:

```
:set smtFile=a.smt2
let f (x : [2]) = take y : [10] where y = x # y
:set prover=w4-z3
:sat \x -> f x == zero
```

With:

```
$ ~/Software/cryptol-2.13.0/bin/cryptol -b a.icry
```

Then despite having set `smtFile=a.smt2`, no `a.smt2` file will be created. Note that `a.smt2` _will_ be created if you change `prover` to `sbv-z3` or `w4-z3`, however, suggesting an inconsistency in the online `what4` backend.

RyanGlScott:
Fixed in #1476.

PR #1476, opened 2022-11-30
Teach online `what4` backend about smtFile option

RyanGlScott:
This passes the value of `smtFile` to each call to `startSolverProcess` in the online `what4` backend, thereby fixing #1475. After this patch, the test case in #1475 now produces an `.smt2` file, as expected.

### DIFF
--- a/src/Cryptol/Symbolic/What4.hs
+++ b/src/Cryptol/Symbolic/What4.hs
@@ -37,7 +37,7 @@ import Control.Concurrent.MVar
 import Control.Monad.IO.Class
 import Control.Monad (when, foldM, forM_, void)
 import qualified Control.Exception as X
-import System.IO (Handle)
+import System.IO (Handle, IOMode(..), withFile)
 import Data.Time
 import Data.IORef
 import Data.List (intercalate, tails, inits)
@@ -373,7 +373,7 @@ satProve ::
   ProverCommand ->
   M.ModuleCmd (Maybe String, ProverResult)
 
-satProve solverCfg hashConsing warnUninterp ProverCommand {..} =
+satProve solverCfg hashConsing warnUninterp pc@ProverCommand {..} =
   protectStack proverError \modIn ->
   M.runModuleM modIn
   do w4sym   <- liftIO makeSym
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Hm. I opted to copy the existing code that writes to `smtFile` when using `w4-offline`. However, one thing that I didn't realize is that this code _always_ prints a message saying "`Writing to SMT-Lib file` ...", even if `smtFile` isn't explicitly set. This is what is causing the various test suite failures, as many more tests are now printing out "`Writing to SMT-Lib file` ..." than before.

Any objections if I change the code such that this "`Writing to SMT-Lib file` ..." message is only printed if `smtFile` is explicitly set?

src/Cryptol/REPL/Command.hs
```diff
@@ -917,50 +920,66 @@ offlineProveSat proverName qtype expr schema mfile = do
...
          case result of
            Just msg -> rPutStrLn msg
            Nothing -> return ()
 
+-- | Write SMT-Lib output using a What4-based backend.
+w4WriteSMTLibOutput ::
+  QueryType ->
+  Maybe FilePath ->
+  (String -> IO ()) {- ^ Continuation for displaying user messages -} ->
+  (Handle -> IO ()) {- ^ Continuation for writing file output -} ->
```
yav, line 951: I think a more traditional name for these are "callbacks" rather than "continuations"?

yav (approved):
LGTM

RyanGlScott:
@yav, do you have an opinion on https://github.com/GaloisInc/cryptol/pull/1476#issuecomment-1333654070?

yav:
Yes, I thin we should only print the message when we were explicitly asked to save the file

RyanGlScott:
Looking at this more closely, the way I am going about this is entirely wrong. The behavior of `{sbv,w4}-offline` is to _always_ print out the SMT-LIB2 term, regardless of whether you are using `stdout` or an `smtFile`. As a result, I really shouldn't be copying the code that implements `w4-offline`, as that's meant for a different purpose.

Luckily, there is a much simpler way to solve this issue: we are currently calling `startSolverProcess` (which initiates an online `what4` solver connection) like so:

https://github.com/GaloisInc/cryptol/blob/8cca24568ad499f06032c2e4eaa7dfd4c542efb6/src/Cryptol/Symbolic/What4.hs#L504

That `Nothing` (of type `Maybe Handle`) indicates that the output will not be duplicated to any file handle. All we have to do to fix the issue at hand is obtain `pcSmtFile`'s `Handle` and pass it there. I've pushed a change which does this, and now all of the test cases pass once more.

This is now ready for another round of review.

Issue #1501, opened 2023-02-21
Literate cryptol parser not permissive enough


benjaminselfridge:
I have a literate Cryptol markdown document that uses the following syntax:

```
```{.cryptol}
<code here>
```

However, Cryptol only accepts 
```
```cryptol
```

I suggest we add a case to cover the former syntax, since it's standard and is produced by `pandoc` in most cases.

RyanGlScott:
I wasn't aware of this before looking it up: this is an extension to Pandoc's particular flavor of markdown called [`fenced_code_attributes`](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes). This is a pretty powerful feature in general, as it lets you write things like this:

~~~
``` {#mycode .haskell .numberLines startFrom="100"}
qsort []     = []
qsort (x:xs) = qsort (filter (< x) xs) ++ [x] ++
               qsort (filter (>= x) xs)
```
~~~

I suppose that if we wanted to support this feature in its full generality, we would need to scan for _any_ occurrence of `.cryptol` within a set of open-close curly braces that appear after the opening ` ``` `. Then again, simply recognizing ` ```{.cryptol} ` might also suffice.

benjaminselfridge:
I also think there may be a way to just tell pandoc to generate 

```
```cryptol
```

instead. If that is possible, it might be a better solution for what I need.

benjaminselfridge:
Fun -- with `-t commonmark-fenced_code_attributes`, I can get it to output

```
``` cryptol
```

with a space before `cryptol`. So I am submitting a small patch to Cryptol to strip the whitespace there before checking if the string is equal to `cryptol`.

RyanGlScott:
Resolved in #1502.

benjaminselfridge:
I have hit enough new issues with pandoc that I now believe the correct approach is to simply see if "cryptol" appears anywhere on the line. I'm getting all sorts of crazy stuff from different versions of pandoc with different command line args.

yav:
Can you say a bit more about what your use case it?   I am asking because it sounds like you are using `pandoc` to generate `markdown`, which is backward from the usual use cases.   Treating every line that contains `cryptol` somewhere in it as code that Cryptol should interpret seems too permissive.

benjaminselfridge:
Sure -- depending on the version of pandoc I'm using, I sometimes get

```
``` {.sourceCode .cryptol}
```

or sometimes

```
``` sourceCode cryptol
```

[...]

PR #1502, opened 2023-02-21
Strips whitespace to be more compatible with pandoc output

benjaminselfridge:
Addresses https://github.com/GaloisInc/cryptol/issues/1501


src/Cryptol/Parser/Unlit.hs
```diff
@@ -105,7 +105,7 @@ markdown = blanks []
 
   isOpenFence l
     | "```" `Text.isPrefixOf` l' =
-      Just $ case Text.drop 3 l' of
+      Just $ case Text.strip (Text.drop 3 l') of
```
RyanGlScott, line 108: Would it suffice to just use `Text.dropWhile isSpace` here instead of `Text.strip`? The latter strips off whitespace from the end of the line, but I don't think that is necessary here, given that the code is already checking is `cryptol` is a prefix. As a result, it shouldn't matter if there is trailing whitespace at the end. (Please correct me if I've overlooked something, of course.)

RyanGlScott (commented):
Thanks for the patch! One minor implementation-related question inline.

src/Cryptol/Parser/Unlit.hs
benjaminselfridge: Nope, that's right :) I'll make that tweak and push...

RyanGlScott:
One more thing: would you be able to add a test case that exercises this bugfix? (If not, I can push a test case myself when I have a chance.)

benjaminselfridge:
> 
Sure, let me take a look

PR #1504, opened 2023-02-27
Add CVC5 support

### LINKED ISSUE #1503: Support CVC5
Cryptol uses `what-1.4` as of #1499, which means that Cryptol now has everything that it needs to support CVC5 as a solver backend, both on the What4 and SBV sides. We should add them.

RyanGlScott:
This patch:

* Makes the necessary changes on the `cryptol` and `cryptol-remote-api` side needed add `cvc5`, `w4-cvc5`, and `sbv-cvc5` solvers. This requires SBV 9.1 or later to include the changes from LeventErkok/sbv#630, which re-exports CVC5-related functionality from all of the places that Cryptol imports from.
* Adds a test case to ensure that basic CVC5 support works.
* Updates the CI and Dockerfile to ensure that CVC5 is included from the `what4-solvers` bindists.

Fixes #1503.


kquick (approved):
Someday it would be nice to unify all those solver snapshot references into a single location.  Somehow.

RyanGlScott:
I think I've set up everything correctly, but the `test-lib, issues, windows-2019` CI job is mysteriously timing out with the message:

```
 Terminate batch job (Y/N)? 
```

Ugh. I will see if I can reproduce this locally.

kquick:
My vote would be to leave it in and note the issue: I doubt we'll have immediate engagement on cvc5 such that it's worth the effort to remove it now and re-add it later.  Assuming you can easily add a CI bypass, that is.

RyanGlScott:
I have good news and bad news. The good news is that the underlying issue in CVC5 (https://github.com/cvc5/cvc5/issues/8900) has already been fixed upstream. I have no idea when we can expect a new CVC5 release with this fix, but I'm inclined to backport the bugfix to the version of CVC5 that we include in `what4-solvers`.

The bad news is that we still haven't quite figured out what causes SBV to go into an infinite loop when using CVC5 on Windows. I discovered in https://github.com/LeventErkok/sbv/issues/644#issuecomment-1454302788 two things:

1. Changing from the default, POSIX IO manager on Windows to the newer (and generally more correct) native IO manager avoids the issue entirely. We may want to consider switching Cryptol to use the native IO manager, even independently of this issue.
2. When using the POSIX IO manager on Windows, there is likely a GHC bug that is causing SBV to hang. We haven't quite gotten to the bottom of what it is yet.

RyanGlScott:
> I have no idea when we can expect a new CVC5 release with this fix, but I'm inclined to backport the bugfix to the version of CVC5 that we include in `what4-solvers`.

Argh, except it isn't that simple. The problem is that `what4-solvers` isn't actually building its own Windows versions of CVC5, as it is currently impossible to build them natively on Windows—you have to use a somewhat elaborate cross-compilation setup. See GaloisInc/what4-solvers#4. As such, we have to download Windows releases from the CVC5 website. Unless we want to figure out how to replicate the cross-compilation setup in `what4-solvers`' CI (and truth be told, I really don't want to), I think it would be best to just wait for a new CVC5 release containing a fix for cvc5/cvc5#8900. I've opened GaloisInc/what4-solvers#31 as a reminder to bump the CVC5 version after this happens.

In the meantime, I will simply comment out the part of the test case that tests `sbv-cvc5`. We could always add it back later if we wanted.

CHANGES.md
```diff
@@ -30,10 +30,14 @@
...
   available in `syntax-highlight/cryptol.vim`
 
-* Add `:new-seed` and `:set-seed` commands to the REPL. 
+* Add `:new-seed` and `:set-seed` commands to the REPL.
   These affect random test generation,
   and help write reproducable Cryptol scripts.
 
+* Add support for the CVC5 solver, which can be selected with
+  `:set prover=cvc5`. If you want to specify a What4 or SBV backend, you can
+  use `:set prover=w4-cvc5` or `:set prover=sbv-cvc5`, respectively.
```
kquick, line 39: We should probably add a note here and/or in the README that sbv-cvc5 on Windows is non-functional at this time due to external issues.

RyanGlScott: I've added a disclaimer here.

PR #1506, opened 2023-03-10
`cryptol-remote-api`: Restore the ability to build with `vector-0.12.*`

RyanGlScott:
The downstream `saw-remote-api` library doesn't quite support `vector-0.13.*` yet, but GaloisInc/cryptol#1499 accidentally restricted `cryptol-remote-api`'s build plan to only allow `vector-0.13.*`. This restores the ability to build `cryptol-remote-api` with `vector-0.12.*` for `saw-remote-api`'s sake.


RyanGlScott:
> I assume there's nothing specific to vector 0.13 that was needed.

Nope, it was only done for the sake of getting a version of `vector` whose version bounds permit building with GHC 9.4.

Issue #1492, opened 2023-01-27
[Remote API] `property` introspection?


WeeknightMVP:
Given a loaded module `M` and a name `n`, how can one apply the Cryptol Remote API for Python to determine whether `n` is a `property`, and which module in scope defines it?

```cryptol
// test_names/P.cry
module P where
  n: [32] -> Bool
  property n x = and x \/ (x + 1 > x)

// test_names/M.cry
module M where
  import P (n)
```

In the Cryptol REPL, `:browse` reports this information:

```
$ export CRYPTOLPATH=$(pwd)/test_names
$ cat test_names/P.cry
// test_names/P.cry
module P where
  n: [32] -> Bool
  property n x = and x \/ (x + 1 > x)
$ cat test_names/M.cry
// test_names/M.cry
module M where
  import P (n)
$ cd cryptol
cryptol$ git pull
...
cryptol$ ./cry run
...
Cryptol> :m M
Loading module Cryptol
Loading module P
Loading module M
M> :b
...

Properties
==========

  From P
  ------

    n : [32] -> Bool

...
```

However, `CryptolConnection.names()` in the Cryptol Remote API for Python does not:

```python
cryptol$ export CRYPTOL_SERVER=$(cabal v2-exec which cryptol-remote-api)
cryptol$ cd ../test_names
test_names$ cat pyproject.toml
...
[tool.poetry.dependencies]
python = "^3.10"
cryptol = {path = "../cryptol/cryptol-remote-api/python"}
...
test_names$ poetry shell
...
test_names$ python3
...
>>> from pprint import pprint
>>> from cryptol import connect
>>> c = connect(reset_server=True)
>>> m = c.load_module()
>>> names = c.names()
>>> n = next(x for x in names if x['name'] == 'n')
>>> pprint(n)
{'name': 'n',
 'type': {'forall': [],
          'propositions': [],
          'type': {'domain': {'type': 'bitvector',
                              'width': {'type': 'number', 'value': 32}},
                   'range': {'type': 'Bit'},
                   'type': 'function'}},
 'type string': '[32] -> Bit'}
```

[...]

PR #1512, opened 2023-04-13
Add more info to `names()` in Python API

m-yac:
As touched upon in #1492, currently the remote API does not provide as much information about names as the `browse` command does in the REPL. This PR expands the result of the `visible names` method (i.e. `names()` in the Python API) to include:
- all the additional information provided by the `browse` command (i.e. what module the name is from, whether or not it is a parameter, and whether or it not it is a property), as well as
- fixity information (i.e. whether or not the name is an infix operator, and if so its associativity and precedence level).

This PR also adds the `property_names` and `parameter_names` functions to the Python API, which return only those names which are properties or parameters, respectively.

Finally, this PR also fixes a bug where `names()` would sometimes hang when a parameterized module is loaded.

See `test_names.py` for an example of this new output of `names` as well as the new the `property_names` and `parameter_names` functions.

Looking ahead: At some point it might also be nice to have `type_names` and `module_names` commands which return similar results to `names`, just for these other two distinct namespaces in Cryptol. It might also be nice to have an easy way to prove a property retrieved from `property_names`.

### DIFF
--- a/cryptol-remote-api/CHANGELOG.md
+++ b/cryptol-remote-api/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Revision history for `cryptol-remote-api` and `cryptol-eval-server`
 
-## 2.13.0 -- YYYY-MM-DD
+## 2.13.2 -- YYYY-MM-DD
+
+* Add more fields (such as `pragmas`, `parameter`, `module`, and `infix`) to
+  the response to the RPC `visible names` method.
+* Do not error if `visible names` is called when a parameterized module is
+  loaded (this used to cause the appearance of the server hanging in such a
+  case).
+
+## 2.13.0 -- 2022-05-17
 
 * v2.13.0 release in tandem with the Cryptol 2.13.0 release. See the Cryptol
   2.13.0 release notes for relevant Cryptol changes. No notable changes to the
--- a/cryptol-remote-api/CHANGELOG.md
+++ b/cryptol-remote-api/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Revision history for `cryptol-remote-api` and `cryptol-eval-server`
 
-## 2.13.2 -- YYYY-MM-DD
+## NEXT -- YYYY-MM-DD
 
 * Add more fields (such as `pragmas`, `parameter`, `module`, and `infix`) to
   the response to the RPC `visible names` method.
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Revision history for `cryptol` Python package
 
-## 2.13.2 -- YYYY-MM-DD
+## NEXT -- YYYY-MM-DD
 
 * Add `property_names` and `parameter_names` commands, used to get only those
   loaded names which are properties or module parameters, respectively.
--- a/cryptol-remote-api/python/CHANGELOG.md
+++ b/cryptol-remote-api/python/CHANGELOG.md
@@ -1,7 +1,15 @@
 # Revision history for `cryptol` Python package
 
+## 2.13.2 -- YYYY-MM-DD
 
-## 2.13.0 -- YYYY-MM-DD
+* Add `property_names` and `parameter_names` commands, used to get only those
+  loaded names which are properties or module parameters, respectively.
+* Add more fields (such as `pragmas`, `parameter`, `module`, and `infix`) to
+  the result of `names` (see `cryptol-remote-api` `CHANGELOG`).
+* Do not hang if `names` is called when a parameterized module is loaded
+  (see `cryptol-remote-api` `CHANGELOG`).
+
+## 2.13.0 -- 2022-05-17
 
 * v2.13.0 release in tandem with the Cryptol 2.13.0 release. See the Cryptol
   2.13.0 release notes for relevant Cryptol changes. No notable changes to the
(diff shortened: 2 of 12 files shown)


cryptol-remote-api/src/CryptolServer/Names.hs
```diff
@@ -47,46 +52,80 @@ instance Doc.DescribedMethod VisibleNamesParams [NameInfo] where
       Doc.Paragraph [ Doc.Text " A"
                     , Doc.Link (Doc.TypeDesc (typeRep (Proxy @JSONSchema))) "JSON Cryptol type"
                     ])
+    , ("module",
+      Doc.Paragraph [Doc.Text "A human-readable representation of the module from which the name originates"])
+    , ("parameter",
+      Doc.Paragraph [ Doc.Text "An optional field which is present and ",
+                      Doc.Literal "True", Doc.Text " iff the name is a module parameter" ])
```
RyanGlScott, line 59: The convention of _only_ including the `parameter` field if it is equal to `True` is a bit odd, since this means that users would have to write code like this:

```py
if 'parameter' in x:
```

Rather than:

```py
if x['parameter']:
```

That is, the actual value of the boolean is never useful, just the presence or absence of the field itself. This might come down to personal taste, but I would favor always including the `parameter` field, even if it is `False`.

Similar reasoning applies to the `infix` and `pragmas` fields, which are also not included under certain circumstances.

RyanGlScott (commented):
This looks reasonable to me, aside from one design question.

cryptol-remote-api/src/CryptolServer/Names.hs
m-yac: This is an excellent point and something I wasn't sure about. First of all, I got the idea from the existing codebase, since this is how the existing `documentation` parameter works: when there is no documentation, the `documentation` parameter simply isn't there.

But my main reason came from remembering that in the end everything gets sent as a JSON string through the RPC interface. There are 125 Prelude definitions that get sent over `names()`, so if every one of them has "`"parameter: False", `" (20 characters), for example, then that's 2,500 characters of redundant information sent every time `names()` is called. Now, I don't know if we even care about that, 2/3K is really not that much data. But it was enough to sway me in the moment.

I don't have strong feelings, so if you don't think we care about how big of a string we send, then I'd be very happy to switch to the (admittedly, easier to use) interface you suggest.

I suppose a compromise would be to keep the RPC API as-is, but on the python side add `parameter: False` whenever `parameter` is not present, and so on. Again, no strong feelings, so I'm curious what you think.

RyanGlScott: I'm not particularly bothered by the fact that we omit fields in certain situations, as that is a pretty natural way to encode a `Maybe` type in an RPC context. That is, omitting the field represents `Nothing`, and including the field represents `Just`. The `documentation` field, for instance, is a `Maybe String`, and the way it's encoded matches my expectations about how `Maybe`s work in RPC.

The thing that bothers me here is that `parameter` _isn't_ a `Maybe`, but rather a raw `Bool`. Except that it's not encoded like a typical RPC `Bool`: in this case, `False` is encoded by omitting the field, and `True` is encoded by including `parameter: True`. In effect, we've made `parameter: False` an unrepresentable state.

Then again, your point about RPC message length is a very compelling argument, and I don't have a good counterargument. In that case, I relent :)

cryptol-remote-api/CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 # Revision history for `cryptol-remote-api` and `cryptol-eval-server`
 
-## 2.13.0 -- YYYY-MM-DD
+## 2.13.2 -- YYYY-MM-DD
```
RyanGlScott, line 3: Nit pick: we should use something like `next` here as the version number until we actually bump the version number in `cryptol-remote-api.cabal` itself.

cryptol-remote-api/python/CHANGELOG.md
```diff
@@ -1,7 +1,15 @@
 # Revision history for `cryptol` Python package
 
+## 2.13.2 -- YYYY-MM-DD
```
RyanGlScott, line 3: Similarly, use something like `next` as the version number here.

cryptol-remote-api/src/CryptolServer/Names.hs
m-yac: Oh, true – that's an excellent point about `Maybe`! So I guess if I want this behavior, I should really encode `parameter` as a `Maybe ()`, i.e. in JSON it's either present and `[]` or not present. I think I'm going to do that and make it clear in the documentation the right way to check if a name is a parameter is to do `"parameter" in n`. This gets rid of that unused `False` state.

m-yac: What I settled on is `parameter` is always either unit or not present. In the JSON this means `[]` or not present, and in the Python this means `()` and not present.

m-yac:
This PR should be ready now. This last commit gives types to the output of `names`, `property_names`, `parameter_names`, and `focused_module` in the Python API as additional documentation for what sort of values these functions return (see the newly added `CryptolNameInfo` and `CryptolModuleInfo` in `cryptoltypes.py`).

Note that in making that change, I converted the `type` fields of the output of `names()` to `CryptolSchema`s for the first time, and found a bug where for unknown constructors, the `arguments` field wasn't getting converted to JSON.

RyanGlScott (approved):

weaversa:
I realize I'm months late, but would it be easy to add the line number where the definition of the name appears to the info that's returned?

PR #1517, opened 2023-05-03
Make Cryptol.Utils.Types safe

pennyannn:
This PR fixes the typecheck failure "Cryptol.Utils.Types: Can't be safely imported!".

### DIFF
--- a/src/Cryptol/Utils/Types.hs
+++ b/src/Cryptol/Utils/Types.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Safe #-}
 -- | Useful information about various types.
 module Cryptol.Utils.Types where
 


pennyannn:
> Thanks for the patch! For some reason, I'm not able to reproduce the error that you are getting. Can you include instructions on how to reproduce this error? (It's quite possible that the fix that you've given is necessary, but I'd like to understand under what circumstances.)
That's a good question. I didn't get this error through the compilation instructions in README. Surprisingly, I got it from the hls I set up for the emacs editor. This error blocked my editor from referencing into files that imported this module because of this error. Let me see if I can reproduce it outside of the hls and get back to you.

PR #1518, opened 2023-05-05
CI: Upgrade Poetry and Python version

RyanGlScott:
The version we were previously using (1.1.6) is affected by this bug, causing invocations of Poetry to fail:
https://github.com/ionrock/cachecontrol/issues/292


kquick (approved):
Can you add a comment or update the description with a note about why a more recent version of poetry is not being used, and a sample of the error if a newer version is used?

PR #1519, opened 2023-05-09
Python: Use poetry.core.masonry.api as build-backend

RyanGlScott:
This is motivated by a desire to switch away from `setuptools`, which has some unfortunate interactions with static analysis tools.

This also bumps the `argo` submodule to bring in the changes from https://github.com/GaloisInc/argo/pull/197, which applies similar changes.


kquick (approved):
c.f. https://github.com/python-poetry/poetry/issues/7892#issuecomment-1540443407

PR #1521, opened 2023-05-23
Document the classes that inhabit the various overloading classes.

yav:


RyanGlScott (approved):
This looks sensible to me. It might be worthwhile to leave comments in the code that this was extracted from with a reminder to update the relevant parts of the reference manual whenever new instances are added (or existing instances are changed).

PR #1523, opened 2023-05-26
Support building with `sbv-10.*`

### LINKED ISSUE #1513: Support `sbv-10.*`
`sbv-10.0` and `sbv-10.1` were recently released, and with them come some internal changes that Cryptol will need to adapt to. In particular, Cryptol uses the `generateSMTBenchmark` function here:

https://github.com/GaloisInc/cryptol/blob/c941eddc235f0f800413fbd3c84a96ff11f4e27a/src/Cryptol/Symbolic/SBV.hs#L473

But `sbv-10.0` [removes `generateSMTBenchmark`](https://github.com/LeventErkok/sbv/commit/96da0a81e15f9f77ee5cd9ffd180033a11293764) in favor of two separate `generateSMTBenchmarkSat` and `generateSMTBenchmarkProof` functions.

RyanGlScott:
`sbv-10.0` brings in two changes to the SBV internals that affect Cryptol.

1. `sbv-10.0` removes the `generateSMTBenchmark` function in favor of two separate `generateSMTBenchmarkSat` and `generateSMTBenchmarkProof` functions.
2. `sbv-10.0` removes the `allSatHasPrefixExistentials` field of `AllSatResult`.

We now use the appropriate CPP to make Cryptol compile before and after these changes.

Fixes #1513.

### DIFF
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -73,7 +73,7 @@ library
                        pretty,
                        prettyprinter     >= 1.7.0,
                        process           >= 1.2,
-                       sbv               >= 9.1 && < 9.3,
+                       sbv               >= 9.1 && < 10.2,
                        simple-smt        >= 0.9.7,
                        stm               >= 2.4,
                        strict,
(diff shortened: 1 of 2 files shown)


yav (approved):
I don't really know this code, but the change look reasonable to me.

PR #1527, opened 2023-06-09
CI: Use cabal 3.10

qsctr:
Should hopefully fix failing Windows tests...


.github/workflows/ci.yml
```diff
@@ -269,6 +269,7 @@ jobs:
       - uses: haskell/actions/setup@v1
```
RyanGlScott, line 269: Does upgrading to `haskell/actions/setup@v2` (and similarly on line 97, which GitHub won't let me highlight) make a difference?

qsctr: Seems like that worked, thanks! I've made that change in a separate PR #1528.

PR #1529, opened 2023-06-21
Enable Github CI build of documentation for all tags and PRs.

kquick:
Updates Github CI process to publish multiple versions of the documentation to the online documentation (github-pages) site.  Originally, only master documentation was deployed.  This change means that there are alternate versions of the documentation available (via a selector on the documentation page) for every PR and every tagged version (for which documentation is available, starting with v2.12.0).

### DIFF
--- a/.github/gen_html_context.py
+++ b/.github/gen_html_context.py
@@ -1,0 +1,43 @@
+import os, sys
+
+# Run this script from one of the doc checkout dirs RefMan doc dir
+# (e.g. docs/ci_build_out/DIR/docs/RefMan})
+#
+# Arguments: owner/repository current_version
+
+d = os.listdir('../../..')
+bldtgt = lambda n: os.path.join('/', sys.argv[1].split('/')[1], n, 'RefMan.html')
+
+html_context={'current_version': sys.argv[2],
+              'versions':[(v, bldtgt(v)) for v in d if not v.startswith('PR_')],
+              'pull_reqs':[(v, bldtgt(v)) for v in d if v.startswith('PR_')]}
+
+if not os.path.isdir('_templates'):
+    # Older versions of the repo do not have this file present as it was only
+    # added in the global gh_pages support, so manually generate it here.
+    os.mkdir('_templates')
+    open('_templates/versions.html', 'w').write('''
+<div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="{{ _('Versions') }}">
+  <span class="rst-current-version" data-toggle="rst-current-version">
+    <span class="fa fa-book">Doc version</span>
+    v: {{ current_version }}
+    <span class="fa fa-caret-down"></span>
+    <div class="rst-other-versions">
+      <dl>
+        <dt>{{ _('Versions') }}</dt>
+        {% for slug, url in versions %}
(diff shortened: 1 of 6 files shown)


kquick:
This is now already live at https://galoisinc.github.io/cryptol

.github/gen_html_context.py
```diff
@@ -0,0 +1,43 @@
...
+      </dl>
+      <dl>
+        <dt>{{ _('Pull Requests') }}</dt>
+        {% for slug, url in pull_reqs %}
+        <dd><a href="{{ url }}">{{ slug }}</a></dd>
+        {% endfor %}
+      </dl>
+    </div>
+  </span>
+</div>
```
RyanGlScott, line 40: Is this the contents of `docs/RefMan/_templates/versions.html`, but copy-pasted? Is there a way to de-duplicate these?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,177 @@
...
+    # released version).
+    #
+    # The general process is to checkout a version of the repository at each
+    # pull-request branch and each tag.  Then the html documentation is generated
+    # for each (a more complicate scheme would access artifacts of the doc builds
+    # from those individual builds, but building the docs is relatively fast, so
+    # it's simpler to do it here).  A top-level index file is created that allows
+    # switching between the various versions of the documentation.  Then the
+    # final results are committed to the gh-pages branch so that the
+    # https://cryptol.github.io/ is useable to browse that documentation.
```
RyanGlScott, line 63: Perhaps say "the https://cryptol.github.io/ page" or just "https://cryptol.github.io/" rather than "the https://cryptol.github.io/".

RyanGlScott (approved):
This looks fantastic, @kquick! I only have a couple of very minor suggestions.

.github/gen_html_context.py
kquick: It is an unmodified copy of that file.  The reason it is here is because any PR created prior to this PR getting merged to main will not have this file, so the run on that PR branch will not have this needed file.  To de-duplicate and maintain needed functionality we would need to remove the direct html file and always use this generated form. I am not wild about that because (a) it's "magic" and hard to find, (b) it means that people running the sphinx generation in their local working directory won't get this (which is not a super-bad thing, since they won't have checked out all the other PR versions, but it means it would be harder for them to see/test changes) and (c) it's HTML embedded in a Python file so any html editing help/highlighting will not be usable, which makes it a little harder.

Longer term we could potentially de-duplicate by removing the auto-generated form from this file once we've closed/merged all the PR's generated from before this file was merged.

RyanGlScott: Ah, I see. Indeed, after reading your explanation, I don't think the cost of de-duplicating this file is worth the payoff. Perhaps it's worth leaving what you wrote here as a comment, but otherwise I'm fine with this approach.

kquick: Updated.

.github/workflows/docs.yml
kquick: Fixed.  It was also the wrong URL... :-(

RyanGlScott (approved):
LGTM (again)

PR #1531, opened 2023-06-22
Prepare for 3.0.0 release

RyanGlScott:


CHANGES.md
```diff
@@ -1,7 +1,25 @@
...
+    between different module parameters (i.e., ones that come from different
+    interfaces or multiple copies of the same interface).
+
+  See the
+  [manual section]()https://galoisinc.github.io/cryptol/master/Modules.html#instantiation-by-parametrizing-declarations
+  for more information.
+
 * Declarations may now use *numeric constraint guards*.   This is a feature
   that allows a function to behave differently depending on its numeric
   type parameters.  See the [manual section](https://galoisinc.github.io/cryptol/RefMan/_build/html/BasicSyntax.html#numeric-constraint-guards))
```
kquick, line 25: Due to the presence of PR1529, this link (and any other doc links) are already changed (even though that PR isn't merged). Properly I should update PR1529 with these changes, and I can do that unless you just want to handle that here rather than running the merges to get those updates.

kquick (approved):
Looks good aside from the online doc references I noted.  Also, don't forget to set the dates in the changelogs...

CHANGES.md
RyanGlScott: Good catch. If you could update #1529, I'm happy to wait until that lands so that I can rebase on top of it.

kquick: Updated #1529.  Turns out there were only two locations, both in this CHANGES.md file.

weaversa: Say, it may be worthwhile to mention the new FFI capability that @qsctr added last summer.

RyanGlScott: I think that's already mentioned here:

https://github.com/GaloisInc/cryptol/blob/4e722ec1ac8294ee4bd60e35a3b7e73472a2968d/CHANGES.md#L10-L12

Or was there something more specific about the FFI that you'd like to see mentioned?

RyanGlScott:
> Also, don't forget to set the dates in the changelogs...

To be clear, you want to add a date next to the "3.0.0" section in the changelog? Or are you envisioning more than that? (You did use plural "dates", after all.)

kquick:
> > Also, don't forget to set the dates in the changelogs...
> 
> To be clear, you want to add a date next to the "3.0.0" section in the changelog? Or are you envisioning more than that? (You did use plural "dates", after all.)

Nope, just that.  I used plural because there are two changelogs and thus two dates to update, but I agree that was imprecise.

CHANGES.md
weaversa: Nope! You got it! I missed that somehow.

PR #1534, opened 2023-06-26
`test.Dockerfile`: Sync Python, Poetry versions with CI

RyanGlScott:
This really ought to have been done at the same time as #1493/#1519, but this was overlooked due to CI-related oversights.

Without these changes, the `cryptol-remote-api` Docker image fails to build. (This was something I noticed when preparing the 3.0.0 release.) Once this lands, I will push these to the `release-3.0.0` branch as well.

### DIFF
--- a/cryptol-remote-api/test.Dockerfile
+++ b/cryptol-remote-api/test.Dockerfile
@@ -1,8 +1,8 @@
-FROM python:3.7
+FROM python:3.11
 # Intended to be built from the root of the cryptol git repository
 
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I manually fired off a CI workflow that builds the `cryptol-remote-api` Docker image, and this appears to fix the issues that I was observing, judging by the fact that this has a green mark: https://github.com/GaloisInc/cryptol/actions/runs/5380114787/jobs/9762336937

kquick (approved):

PR #1536, opened 2023-06-27
Fix doc ref in CHANGES and use better wording.

kquick:


kiniry:
Thanks for jumping on my suggestions/feedback, @kquick!

kiniry (approved):
LGTM

PR #1541, opened 2023-07-05
Sanitize strings in `to_cryptol_str` in quoting.py

m-yac:
This PR updates `to_cryptol_str` to correctly sanitize strings. For example, if `s = '" \ \n'`, then `cry_f('{s}')` now correctly gives `'"\\" \\\\ \\n"'` instead of the unescaped `'"" \\ \n"'`.

### DIFF
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -29,7 +29,8 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
-        return f'"{val}"'
+        val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
+        return f'"{val_san}"'
     elif hasattr(val, '__to_cryptol_str__'):
         return parenthesize(val.__to_cryptol_str__())
     else:
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -14,6 +14,8 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     if isinstance(val, bool):
         return 'True' if val else 'False'
     elif isinstance(val, tuple):
+        if len(val) == 1:
+            raise TypeError("Unable to convert 1-tuple to Cryptol syntax: " + str(val))
         return '(' + ', '.join(to_cryptol_str(x) for x in val) + ')'
     elif isinstance(val, dict):
         return '{' + ', '.join(f'{k} = {to_cryptol_str(v)}' for k,v in val.items()) + '}'
@@ -22,7 +24,7 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, list):
         return '[' + ', '.join(to_cryptol_str(x) for x in val) + ']'
     elif isinstance(val, BV):
-        if val.size() % 4 == 0:
+        if val.size() > 0 and val.size() % 4 == 0:
             return val.hex()
         else:
             return f'({val.to_signed_int()} : [{val.size()}])'
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -29,9 +29,8 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
-        # Sanitizing strings for Cryptol - see <string> of src/Cryptol/Parser/Lexer.x
-        val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
-        return f'"{val_san}"'
+        chars = list(val.encode('latin-1'))
+        return f'({to_cryptol_str(chars)} : [{len(chars)}][8])'
     elif hasattr(val, '__to_cryptol_str__'):
         return parenthesize(val.__to_cryptol_str__())
     else:
--- a/cryptol-remote-api/python/cryptol/quoting.py
+++ b/cryptol-remote-api/python/cryptol/quoting.py
@@ -29,6 +29,7 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
+        # Sanitizing strings for Cryptol - see <string> of src/Cryptol/Parser/Lexer.x
         val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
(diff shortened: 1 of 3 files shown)


cryptol-remote-api/python/cryptol/quoting.py
```diff
@@ -29,7 +29,8 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
-        return f'"{val}"'
+        val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
```
RyanGlScott, line 32: Should we include all of Python's escape characters here for completeness? https://python-reference.readthedocs.io/en/latest/docs/str/escapes.html

m-yac: In this case we're sanitizing the string for Cryptol, not for Python, so I believe backslash, double-quote, and newline should be enough? https://github.com/GaloisInc/cryptol/blob/master/src/Cryptol/Parser/Lexer.x#L77

RyanGlScott: Oops, I misunderstood the problem! Indeed, what you have should be enough, but do leave a reference to the Cryptol lexer as justification for this choice.

m-yac: Done - thanks!

RyanGlScott (approved):

cryptol-remote-api/python/cryptol/quoting.py
```diff
@@ -29,7 +29,9 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
-        return f'"{val}"'
+        # Sanitizing strings for Cryptol - see <string> of src/Cryptol/Parser/Lexer.x
+        val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
```
yav, line 33: As far as I understand, the purpose of this function is to turn a Python value to the corresponding Cryptol value, written as a String.  Unfortunately, the notation for Cryptol string literals and Python ones are not exactly the same, so I believe the correct thing here would be to go through each character and use the relevant representation.  This would require careful examination of the Cryptol lexer, and would probably be a good 
opportunity to document the exact notation we are using.

Some differences that come to mind:
   * Make sure that all special escape sequences match
   * Numeric escape sequence are different (in Python "\10" is 8, because it uses octal, but in Cryptol it is 10, because we use base 10)
    * I think Python strings support Unicode, but Cryptol ones are just bytes, so we'd need to decide on what to do in that case---maybe just report an error, that this is not a valid Cryptol value.

yav (changes requested):

cryptol-remote-api/python/cryptol/quoting.py
weaversa: I agree w/ @yav that to get this right requires more care, and . Here is an idea I don't see has been discussed -- don't send strings to Cryptol at all, let Python handle strings (using `bytes`) and just send sequences of 8-bit numbers to Cryptol.

```
list(bytes("hello, \10", 'ascii'))
[104, 101, 108, 108, 111, 44, 32, 8]
```

Of course, if we did this we would want to send the type along, so that Cryptol could infer the sequence had 8-bit values -- "[104, 101, 108, 108, 111, 44, 32, 8] : [8][8]".

weaversa: What about cases where Cryptol and Python disagree? I am leaning towards Python, since Python is the source language here.

```
Cryptol> :s base=10
Cryptol> "hello, \&"
[104, 101, 108, 108, 111, 44, 32]
```
```
>>> list(bytes("hello, \&", 'ascii'))
[104, 101, 108, 108, 111, 44, 32, 92, 38]
```

m-yac: These are excellent points @yav and @weaversa – thanks for the comments!

> This would require careful examination of the Cryptol lexer, and would probably be a good
opportunity to document the exact notation we are using.

I took a quick look, and it appears that the strategy is to lex strings using Haskell's `read` function! By that I mean the strategy is to collect all the characters in the string (including the quote delimiters and with no processing of escape sequences), then use the built-in Haskell [`reads` function](https://github.com/GaloisInc/cryptol/blob/57226efb27e801dc911b8caf865bd3d57425fc1c/src/Cryptol/Parser/LexerUtils.hs#L109).

So instead of trying to capture the complexity of Haskell strings in this tiny bit of Python code, I really like @weaversa's solution. It's simple and leaves no room for ambiguity.

As per your second comment, Sean, I agree that we should send over the bytes corresponding to the string from Python's perspective. The alternative – trying to figure out what the bytes would be if the literal were instead parsed as Cryptol – feels about as difficult as just figuring out how to properly send the string literal.

weaversa (approved):
This works for me!

cryptol-remote-api/python/cryptol/quoting.py
```diff
@@ -29,9 +29,8 @@ def to_cryptol_str(val : Union[CryptolValue, str, CryptolJSON]) -> str:
     elif isinstance(val, OpaqueValue):
         return str(val)
     elif isinstance(val, str):
-        # Sanitizing strings for Cryptol - see <string> of src/Cryptol/Parser/Lexer.x
-        val_san = val.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
-        return f'"{val_san}"'
+        chars = list(val.encode('latin-1'))
+        return f'({to_cryptol_str(chars)} : [{len(chars)}][8])'
```
yav, line 33: I think this look good now.

While we are fixing things here, could you also fix the instance for BV a few lines above?  I think it would do the wrong thing for 0-bit vectors, in that it would render them as `0x0` which would then be interpreted as a 4-bit vector by Cryptol.    You'd just need to change the conditional to `val.size() > 0 && val.size() % 4 == 0`

m-yac: Fixed both - thanks @yav!

Also looks like the Cryptol pretty-printer makes the same mistake that I did for `[0]`:
```
Cryptol> 0 : [0]
0x0
Cryptol> 0x0 : [0]

[error] at <interactive>:5:1--5:4:
  Type mismatch:
    Expected type: 0
    Inferred type: 4
    When checking user annotation
```
Finally I noticed the following – is `0x` the preferred way to represent the inhabitant of `[0]`?
```
Cryptol> :t 0x
0x0 : [0]
```

m-yac: Ah okay, good to know about the way printing on the REPL works. And I definitely agree with your suggestion of falling back on the else case for 0-bit values (I already implemented it in 93b3d79f3cbc0d11e2806d5400a2d903402f473d).

I think the PR is waiting on you to give the OK now – so if there's nothing else, let me know when you approve and I can merge it!

yav (approved):
Thanks!

weaversa:
@m-yac Would you be willing to update `https://pypi.org/project/cryptol/`?

PR #1547, opened 2023-07-06
Add `version` command to cryptol-remote-api

### LINKED ISSUE #1544: cryptol-remote-api version info
It would be helpful to have a `--version` command line argument and/or `Connection.version()` command in the Remote APIs (`saw-remote-api` as well as `cryptol-remote-api`).

m-yac:
This PR adds a `version` command to the `cryptol-remote-api` and its Python interface. For example:
```python
>>> connection.version()
CryptolVersionInfo(
  rpc_version='3.0.1.99',
  version='3.0.0.99',
  commit_hash='8c811b06a435f5009e8b75b36ecc8fc220e4d6bd',
  commit_branch='version-python-cmd',
  commit_dirty=True,
  ffi_enabled=True)
```
See the [docs for the command in `Version.hs`](https://github.com/GaloisInc/cryptol/blob/3d69f6dc7fe1a00235e3794eb9a26bb011e9bafe/cryptol-remote-api/src/CryptolServer/Version.hs#L41) or the [docstring of `CryptolVersionInfo` in the Python API](https://github.com/GaloisInc/cryptol/blob/3d69f6dc7fe1a00235e3794eb9a26bb011e9bafe/cryptol-remote-api/python/cryptol/synchronous.py#L86) for descriptions of these fields.

Resolves #1544

### DIFF
--- a/cryptol-remote-api/cryptol-remote-api.cabal
+++ b/cryptol-remote-api/cryptol-remote-api.cabal
@@ -1,6 +1,6 @@
 cabal-version:       2.4
 name:                cryptol-remote-api
-version:             3.0.0.99
+version:             3.0.1
 license:             BSD-3-Clause
 license-file:        LICENSE
 author:              Galois, Inc.
--- a/cryptol-remote-api/cryptol-remote-api.cabal
+++ b/cryptol-remote-api/cryptol-remote-api.cabal
@@ -1,6 +1,6 @@
 cabal-version:       2.4
 name:                cryptol-remote-api
-version:             3.0.1
+version:             3.0.1.99
 license:             BSD-3-Clause
 license-file:        LICENSE
 author:              Galois, Inc.
--- a/cryptol-remote-api/cryptol-remote-api.cabal
+++ b/cryptol-remote-api/cryptol-remote-api.cabal
@@ -85,6 +85,7 @@ library
 
   other-modules:
     CryptolServer.AesonCompat
+    Paths_cryptol_remote_api
 
 executable cryptol-remote-api
   import:              deps, warnings, errors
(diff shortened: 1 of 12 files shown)


cryptol-remote-api/cryptol-remote-api.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:       2.4
 name:                cryptol-remote-api
-version:             3.0.0.99
+version:             3.0.1.99
```
RyanGlScott, line 3: A comment about process: the convention I've been using for releases (in this case, a PyPI release) is to make a commit that bumps the version numbers to X, and then a subsequent commit which bumps the commit to X.99, thereby turning it into a development version for a future X+1 release. That way, we can tag the commit that bumps it to X.

Do you mind doing that here? That is, have one commit that bumps this to 3.0.1 (which we tag with something like `git tag cryptol-remote-api-3.0.1`), plus an subsequent commit on top of that which bumps it to 3.0.1.99 (the development version for a future 3.0.2/3.1.0 release, depending on how the 3.* release series evolves).

cryptol-remote-api/python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cryptol"
-version = "3.0.0.99"
+version = "3.0.1.99"
```
RyanGlScott, line 3: A similar comment about versioning processes here.

cryptol-remote-api/cryptol-remote-api.cabal
m-yac: Ah, that's a good system! I added commit ae4cb538c9dd1e94438dc8602076bb1c958f3c5d, which bumps the versions to `3.0.1.00`, commit a0903546187826e16aec2ec797da5db996f8e6af, which bumps the versions to `3.0.1.99`, and the tag [cryptol-remote-api-3.0.1](https://github.com/GaloisInc/cryptol/releases/tag/cryptol-remote-api-3.0.1) which is set to the first commit. Does that look right?

RyanGlScott (approved):

PR #1549, opened 2023-07-11
CI: Bump what4-solvers snapshot to 20230711

### LINKED ISSUE #1548: CVC5 fails with Cryptol 3.0.0
Using the latest Cryptol docker container, Cryptol doesn't detect CVC5 via the command `:s prover=cvc5`, even though it is present in the path.

Oddly, running `:s prover=w4-cvc5` works fine.
```
% docker run -it ghcr.io/galoisinc/cryptol
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻  
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃  
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.0.0

Loading module Cryptol
Cryptol> :s prover=cvc5
Warning: cvc5 installation not found
Cryptol> :s prover=w4-cvc5
Cryptol> 
```

RyanGlScott:
This includes the following changes:

* The bindist URLs now include the OS architecture, I have tweaked the parts of CI that downloads bindists accordingly.
* This snapshot includes `cvc5-1.0.5`, which has fixed a bug that was responsible for #1548. As such, this fixes #1548.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ USER cryptol
 WORKDIR /cryptol
 RUN mkdir -p rootfs/usr/local/bin
 WORKDIR /cryptol/rootfs/usr/local/bin
-RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20221212/ubuntu-22.04-bin.zip"
+RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20230711/ubuntu-22.04-X64-bin.zip"
 RUN unzip solvers.zip && rm solvers.zip && chmod +x *
 WORKDIR /cryptol
 ENV PATH=/cryptol/rootfs/usr/local/bin:/home/cryptol/.local/bin:/home/cryptol/.ghcup/bin:$PATH
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,8 @@ USER cryptol
 WORKDIR /cryptol
 RUN mkdir -p rootfs/usr/local/bin
 WORKDIR /cryptol/rootfs/usr/local/bin
+# The URL here is based on the same logic used to specify BIN_ZIP_FILE in
+# `.github/workflow/ci.yml`, but specialized to x86-64 Ubuntu.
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@ USER cryptol
 WORKDIR /cryptol
 RUN mkdir -p rootfs/usr/local/bin
 WORKDIR /cryptol/rootfs/usr/local/bin
-RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20221212/ubuntu-22.04-bin.zip"
+RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20230711/ubuntu-22.04-X64-bin.zip"
```
kquick, line 17: Maybe add a comment here (and the other Dockerfile) that the zipfile name is based on the CI matrix.os and runner.arch so that future devs know where this name came from.  Perhaps even a contra-comment in the YAML files as well indicating that the BIN_ZIP_FILE is consumed by these two dockerfiles.

kquick (approved):

Dockerfile
RyanGlScott: I can certainly leave a comment about the naming convention here, but the Dockerfiles have nothing to do with `BIN_ZIP_FILE`—that environment variable is only used by `.github/ci.sh`. (Perhaps you meant to suggest adding a comment there?)

kquick: Yes, I was unclear: it might be helpful to have a comment there that says that the name specified for BIN_ZIP_FILE will be consumed by the dockerfiles, not the BIN_ZIP_FILE variable itself.  My concern is that if someone changes the BIN_ZIP_FILE value in the CI yaml files that they should know they will need to make corresponding changes in the dockerfiles.

RyanGlScott: I've added a commit which adds some pointers next to `SOLVER_PKG_VERSION` and `BIN_ZIP_FILE` in the `ci.yml` file, as well as the lines in each Dockerfile that were inspired by them.

PR #1557, opened 2023-07-26
Changes that hopefully fix #1552.

### LINKED ISSUE #1552: Panic - Messed up size accounting
I ran across this panic today. I'm sorry I don't have a ready example.

```
Revision: UNKNOWN
Branch: UNKNOWN
Location: lookupSeqMap
Message: Messed up size accounting
        Nat 2
         2
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.0.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/Backend/SeqMap.hs:100:28 in cryptol-3.0.0.99-inplace:Cryptol.Backend.SeqMap
```

yav:
* Make the code more thread safe
* Do a bounds check in the GC closure

Not sure if the 2nd is needed, or if we have an assumption that `lookupSeqMap` will only be called with in bounds values---this is something we should investigate and document.

### DIFF
--- a/src/Cryptol/Backend/SeqMap.hs
+++ b/src/Cryptol/Backend/SeqMap.hs
@@ -72,8 +72,9 @@ data SeqMap sym a
                  !(SeqMap sym a)
   | MemoSeqMap
      !Nat'
-     !(IORef (Map Integer a))
-     !(IORef (Integer -> SEval sym a))
+     !(IORef (Map Integer a, Integer -> SEval sym a))
+     !(Integer -> SEval sym a)
+      -- Use this to overwrite the evaluation function when the cache is full
 
 indexSeqMap :: (Integer -> SEval sym a) -> SeqMap sym a
 indexSeqMap = IndexSeqMap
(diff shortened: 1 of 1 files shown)


src/Cryptol/Backend/SeqMap.hs
```diff
@@ -173,13 +176,21 @@ memoMap _sym _sz x@(MemoSeqMap{}) = pure x
...
   where
     eval stk i = sWithCallStack sym stk (lookupSeqMap x i)
 
+    -- Not 100% sure that we actually need to do the bounds check here,
+    -- or if we are assuming that the argument would be in bounds, but
+    -- it doesn't really hurt to do it, as if we *did* already do the check
+    -- this code should never run (unless we messed up something).
+    evalPanic i = case sz of
+                    Nat sz' | i < 0 || i >= sz' -> invalidIndex sym i
+                    _ -> panic "lookupSeqMap"
```
RyanGlScott, line 191: Nit pick: this function is no longer located in `lookupSeqMap`, so it would be worth changing the string used in the call to `panic` here.

src/Cryptol/Backend/SeqMap.hs
```diff
@@ -72,8 +72,9 @@ data SeqMap sym a
                  !(SeqMap sym a)
   | MemoSeqMap
      !Nat'
-     !(IORef (Map Integer a))
-     !(IORef (Integer -> SEval sym a))
+     !(IORef (Map Integer a, Integer -> SEval sym a))
+     !(Integer -> SEval sym a)
+      -- Use this to overwrite the evaluation function when the cache is full
```
RyanGlScott, line 77: I'm not entirely sure why we need to add this as an additional field to `MemoSeqMap`. Unless I'm missing something, the only place where we ever construct `MemoSeqMap` values is in the `memoMap` function, and `memoMap` only ever instantiates this field to `evalPanic`. Perhaps we should just turn `evalPanic` into a top-level function and refer to that instead?

yav: I had to add it there, because `evalPanic` (which perhaps should be renamed) throws an exception if the index is out of bounds, and that needs access to the backend `sym`.  Unfortunately, `sym` is not passed to `lookupSeqMap`, so instead of changing the API I just added the function to use when the cache is complete. 

I imagine passing in the `sym` would not be a huge change though, so perhaps we should do that instead.

One other thing I am not sure, and we really should document, is what are the invariants on `lookupSeqMap`.  In particular, I am not sure if it might be assuming that someone has already checked that the index is in bounds, so maybe we don't even need that exception...   I added it because in the original ticket it looked like that map size is 2, and the index is 2, so that would make it out of bounds, which would suggest that we don't have such an invariant.

Issue #1495, opened 2023-02-06
Document that foreign functions should be pure


yav:
We should add note to the FFI section to make it explicit that our intention is that foreign function are pure (i.e., do not have observable side effects).

ramsdell:
Let me expand on this request.

The RefMan in the cryptol repository gives its readers no hint as to the fact that the order in which FFI calls are evaluated can influence the meaning of an expression.  For example, suppose there is a C function called count that has some internal state, and it increments an integer with each call, ignores its input, and returns the integer.  After loading this into cryptol
 
```
foreign count: [32] -> [32]
 
seq : [10][32]
seq = [ count i | i <- [0 .. 9 : [32]] ]
```
 
and typing `seq ! 2; seq` you get a different answer than had you started by typing `seq`.
 
Please expand the FFI section in the RefMan to help readers understand why this happens so they understand why one should not load functions with observable state.

RyanGlScott:
Done in #1558.

PR #1558, opened 2023-07-26
RefMan: document that the FFI expects pure functions

### LINKED ISSUE #1554: FFI call semantics
The FFI is supposed to be used with stateless C functions so as to ensure referential transparency.  However, there is a use case in which one wants to use a stateful C routine to generate a sequence of random values by repeatedly calling the C function with a call of the same form.  Here is an example.

Suppose you load the K&R rand function that is available on the net with:

```
foreign rand: () -> [64]
```

Currently, in the interpreter, if one types `rand ()` repeatedly, one gets different values.

```
cryptol -b rand.bat
Loading module Cryptol
Loading module Cryptol
Loading module Main
Loading dynamic library rand.dylib
0x00000000000041c6
0x000000000000167e
0x0000000000002781
0x000000000000446b
0x000000000000794b
ramsdell@MM270114-PC ffi_rand % 
```

Where `rand.bat` is

```
:load rand.cry
rand ()
rand ()
rand ()
rand ()
rand ()
```

Are future versions of cryptol guaranteed to preserve this behavior?

RyanGlScott:
This is a disclaimer that while you _can_ use impure functions in the Cryptol FFI, we make no guarantees about the precise behavior of the side effects that they functions may perform.

Fixes #1554.

### DIFF
--- a/docs/RefMan/FFI.rst
+++ b/docs/RefMan/FFI.rst
@@ -345,6 +345,14 @@ Evaluation
 
 All Cryptol arguments are fully evaluated when a foreign function is called.
 
+The FFI is intended to be used with pure functions that do not perform side
+effects such as mutating global state, printing to the screen, interacting with
+the file system, etc. Cryptol does not enforce this convention, however, so it
+is possible to use impure functions from the FFI if you wish. Cryptol does not
+make any guarantees about the order in which side effects will be executed, nor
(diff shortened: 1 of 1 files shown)


qsctr (approved):
Looks good to me. This also fixes #1495.

Issue #1382, opened 2022-07-14
Allow navigation into nested modules in the REPL


m-yac:
Currently `:m`/`:module` only works for top-level modules, not nested modules.

glguy:
PR #1682 introduces a new `:focus` command that can be used to navigate into loaded submodules

PR #1559, opened 2023-07-29
Make names in scope in functors accessible when their instantiations are loaded at the REPL

### LINKED ISSUE #1237: Module Parameters public or private?
In the recent past (just a few months ago), parameterized module parameters have been public, by which I mean that if I instantiated a parameterized module, module parameters were in scope in the REPL. However, it seems they are now private (no longer in scope in the REPL).

Two things, I guess...1) I don't see this behavior documented. 2) What is the right thing to do here?

### LINKED ISSUE #1455: [New module system] Cryptol Prelude not available when instantiating top-level modules
Here are some examples:

```
// in file F.cry
module F where
  parameter
    y : [32]
  f x = x + y

// In file A.cry
module A where
  y = 5

// In file X.cry
module X = F where
   y = 5

// In file Y.cry
module Y = F { A }
```

When loading `X.cry` or `Y.cry` the modules load correctly and the things defined in them are accessible, but things from the Cryptol Prelude are not.  For example:
```
f 2 == 5
```
Complains with `==` is not defined.   Using the `:browse` confirms that the Cryptol Prelude is not in scope.

### LINKED ISSUE #1556: Same internal name used for functor parameters and their instantiations
While implementing a solution to #1455, we found that functor instantiations use the same internal names as the corresponding parameters in the functor. Specifically, if we have
```
interface module I where
  x : [8]
```
```
module F where
  import interface I
  y = x + 1
```
```
module M where
  x = 2
```
```
module N = F { M }
```
then `N::x` reuses the exact same internal name as `F::x` (the full name being in this example `F::I::x_5244`) rather than a fresh name.

Currently, this does not present an immediate issue with evaluating expressions at the REPL, since the instantiated parameters are not accessible from outside of the instantiated module, and we can only ever look inside a single module in the REPL at once, so the REPL does not get confused by the duplicate names. However, the `OrigName` of the instantiated parameter would be wrong and could cause issues elsewhere.

### LINKED ISSUE #1561: Type synonyms with same name in distinct interface imports getting confused when instantiating functors
If we have
```
interface module I where
type n : #
type m = 2 * n
```
```
module F where
import interface I as J
import interface I as K
```
```
module M where
type n = 2
```
```
module N where
type n = 3
```
```
module Inst = F { J = M, K = N }
```
then `F` correctly defines `F::J::m = 4` and `F::K::m = 6`, but in Inst we have `Inst::J::m = 6` and `Inst::K::m = 6`.

Furthermore, if we instead have
```
module F where
import interface I
import interface I as K
```
```
module Inst = F { I = M, K = N }
```
then in `Inst` we only have `Inst::m = 6`.

### LINKED ISSUE #1562: Interface type synomym shadowing
```cryptol
module InterfaceSubmoduleShadowing where

interface submodule I where
  type Zp = [8]

submodule S1 where
  import interface submodule I

submodule S2 where
  import interface submodule I
  import submodule S1 { interface I } ()  // commenting out removes error

submodule S where
  import interface submodule I
  import submodule S1 { interface I } ()  // commenting out removes error
  import submodule S2 { interface I } ()  // [error] (commenting out removes it)
```

```
Cryptol> :l InterfaceSubmoduleShadowing.cry
Loading module Cryptol
Loading module InterfaceSubmoduleShadowing

[error] at InterfaceSubmoduleShadowing.cry:1:1--16:9:
  Type synonym 'InterfaceSubmoduleShadowing::S1::I::Zp' shadowing an existing type synonym with the same name.
```

Commenting out any of the indicated lines removes the error, but aliasing interface imports to different names does not.

This might be related to #1561.

qsctr:
Fixes #1455, making anything in scope of the functor in scope at the REPL as well when an instantiation of the functor is loaded and focused. In particular, the prelude will be in scope. As part of handling functor parameters correctly, this also fixes #1556, fixes #1237, fixes #1561, and fixes #1562.

In addition, this does some work towards implementing #1382, by storing the in-scope information for nested modules into `Cryptol.ModuleSystem.Interface.IFaceNames`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+# UNRELEASED
+
+* Fixed #1556, #1237, and #1561.
+* Fixed #1455, making anything in scope of the functor in scope at the REPL as
+  well when an instantiation of the functor is loaded and focused,
+  design choice (3) on the ticket.  In particular, the prelude will be in scope.
+
+
 # 3.0.0 -- 2023-06-26
 
 ## Language changes
--- a/cryptol.cabal
+++ b/cryptol.cabal
@@ -141,6 +141,7 @@ library
                        Cryptol.ModuleSystem.Name,
                        Cryptol.ModuleSystem.Names,
                        Cryptol.ModuleSystem.NamingEnv,
+                       Cryptol.ModuleSystem.NamingEnv.Types,
                        Cryptol.ModuleSystem.Binds
                        Cryptol.ModuleSystem.Exports,
                        Cryptol.ModuleSystem.Renamer,
--- a/src/Cryptol/ModuleSystem/Base.hs
+++ b/src/Cryptol/ModuleSystem/Base.hs
@@ -631,7 +631,11 @@ checkModule isrc m = do
   rewMod <- case tcm of
               T.TCTopModule mo -> T.TCTopModule <$> liftSupply (`rewModule` mo)
               T.TCTopSignature {} -> pure tcm
-  pure (R.rmInScope renMod,rewMod)
+  let nameEnv = case tcm of
+                  T.TCTopModule mo -> T.mInScope mo
+                  -- Name env for signatures does not change after typechecking
+                  T.TCTopSignature {} -> mInScope (R.rmModule renMod)
+  pure (nameEnv,rewMod)
 
 data TCLinter o = TCLinter
   { lintCheck ::
--- a/src/Cryptol/ModuleSystem/Binds.hs
+++ b/src/Cryptol/ModuleSystem/Binds.hs
@@ -13,6 +13,7 @@ module Cryptol.ModuleSystem.Binds
   , topModuleDefs
   , topDeclsDefs
   , newModParam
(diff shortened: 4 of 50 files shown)


weaversa:
Thank you!

Say, does this capability port over to SAW? That is, when a Cryptol module is loaded in SAW, will the scope be the same as when a module is loaded into Cryptol, or does this work need to be ported to SAW as well?

qsctr:
For the most part, this change only fixes the behavior of what is in scope in the Cryptol REPL, so it does not affect SAW. However I don't think we need a fix for SAW since SAW does not follow Cryptol's normal module system scoping rules and everything would already be accessible in a SAW `import`. In particular, Prelude functions, things defined in the functor, instantiated value parameters, and type synonyms in the interface are already all available when a module instantiation is imported in SAW.

The one exception is instantiated type parameters: currently these are not available when an instantiation is imported in SAW since they are treated slightly differently internally in Cryptol. This PR should fix that.

qsctr:
Reviewed in person by @yav

PR #1565, opened 2023-08-21
Allow cryptol implementation of foreign functions

### LINKED ISSUE #1397: Cryptol implementation of `foreign` functions
Allow `foreign` declarations to have an implementation in Cryptol as well. We could fall back to using the Cryptol implementation in situations where we can't do FFI, such as on unsupported platforms or in symbolic backends. It would also be nice to have some kind of integration with SAW so we can verify that the C function matches the Cryptol one.

qsctr:
Closes #1397.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # UNRELEASED
 
+## Language changes
+
+* `foreign` functions can now have an optional Cryptol implementation, which by
+  default is used when the foreign implementation cannot be found, or if the FFI
+  is unavailable. The `:set evalForeign` REPL option controls this behavior.
+
+## Bug fixes
+
 * Fixed #1556, #1237, and #1561.
 * Fixed #1455, making anything in scope of the functor in scope at the REPL as
   well when an instantiation of the functor is loaded and focused,
--- a/cryptol-remote-api/src/CryptolServer/Data/Expression.hs
+++ b/cryptol-remote-api/src/CryptolServer/Data/Expression.hs
@@ -518,7 +518,7 @@ getCryptolExpr (Let binds body) =
               , CP.bExport = CP.Public
               }) .
       fakeLoc .
-      CP.DExpr <$>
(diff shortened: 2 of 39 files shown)


CHANGES.md
```diff
@@ -1,5 +1,13 @@
 # UNRELEASED
 
+## Language changes
+
+* `foreign` functions can now have an optional Cryptol implementation, which by
+  default is used when the foreign implementation cannot be found, or if the FFI
+  is unavailable. The `:set evalForeign` REPL option controls this behavior.
```
RyanGlScott, line 7: I haven't looked much into the actual code, but what exactly does "when the foreign implementation cannot be found" mean? I ask since I remember getting a shared library for use with Cryptol's FFI to be a somewhat tricky process, and I worry that if a user screws this up, then Cryptol might silently switch to the non-FFI implementation of a function without the user realizing it. Is it obvious when a function's foreign implementation cannot be found?

qsctr: Yes, we emit a warning at load time. It looks something like
https://github.com/GaloisInc/cryptol/blob/5a7746799a9151cc4090e7ec17954c86d2afcc2b/tests/ffi/ffi-fallback.icry.stdout#L5-L8
or
https://github.com/GaloisInc/cryptol/blob/5a7746799a9151cc4090e7ec17954c86d2afcc2b/tests/ffi/ffi-fallback.icry.stdout#L100-L103

RyanGlScott: Ah, I completely overlooked the `ffi-fallback.c` test case. That makes me feel much better, thanks.

qsctr: And to answer your first question, it is essentially when either `dlopen` fails (in which case we use the cryptol implementation for all foreign functions in the module) or one or more `dlsym` fails (in which case we use the cryptol implementation for the failed ones).

yav (approved):
Looks good.

PR #1570, opened 2023-08-30
Apply `apSubst` to `Type` field of `EPropGuards` in `TVars` instance for `Expr`

### LINKED ISSUE #1569: Incorrect `EPropGuards` case in `TVars` instance for `Expr`
The `EPropGuards` case in the `TVars` instance for `Expr` does not perform substitution inside the `Type` field of `EPropGuards`:

https://github.com/GaloisInc/cryptol/blob/63f0ac6968967fb7898333778b5206083fdcb375/src/Cryptol/TypeCheck/Subst.hs#L438

This is causing declarations in prop guards whose type mentions a module type parameter to not get instantiated properly by `moduleInstance`, which causes the bug in GaloisInc/saw-script#1923.

qsctr:
Fixes #1569.

Not really sure how to test this since the `Type` field is not really used within Cryptol itself. GaloisInc/saw-script#1923 does seem to be fixed by this change though.

### DIFF
--- a/src/Cryptol/TypeCheck/Subst.hs
+++ b/src/Cryptol/TypeCheck/Subst.hs
@@ -436,7 +436,7 @@ instance TVars Expr where
         EWhere e ds   -> EWhere !$ (go e) !$ (apSubst su ds)
 
         EPropGuards guards ty -> EPropGuards
-          !$ (\(props, e) -> (apSubst su `fmap'` props, apSubst su e)) `fmap'` guards
+          !$ (\(props, e) -> (apSubst su `fmap'` props, go e)) `fmap'` guards
           !$ apSubst su ty
 
 instance TVars Match where
--- a/src/Cryptol/TypeCheck/Subst.hs
+++ b/src/Cryptol/TypeCheck/Subst.hs
@@ -435,7 +435,9 @@ instance TVars Expr where
 
         EWhere e ds   -> EWhere !$ (go e) !$ (apSubst su ds)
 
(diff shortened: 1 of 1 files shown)


src/Cryptol/TypeCheck/Subst.hs
```diff
@@ -435,7 +435,9 @@ instance TVars Expr where
 
         EWhere e ds   -> EWhere !$ (go e) !$ (apSubst su ds)
 
-        EPropGuards guards ty -> EPropGuards !$ (\(props, e) -> (apSubst su `fmap'` props, apSubst su e)) `fmap'` guards .$ ty
+        EPropGuards guards ty -> EPropGuards
+          !$ (\(props, e) -> (apSubst su `fmap'` props, apSubst su e)) `fmap'` guards
+          !$ apSubst su ty
```
RyanGlScott, line 440: Perhaps slightly cleaner would be writing `go` instead of `apSubst su`, since the substitution isn't changing in any of the recursive calls:

```suggestion
          !$ (\(props, e) -> (go `fmap'` props, go e)) `fmap'` guards
          !$ go ty
```

This isn't really your fault, however, since the other cases for this function are already inconsistent about using `apSubst su` versus `go`. We should probably stick to using `go` everywhere.

qsctr: `go` has type `Expr -> Expr` so we can only use it for recursive calls on `Expr`s. The rest of the function seems to be consistent in doing that except for the call to `apSubst su e` here, so I'll just change this one.

RyanGlScott (approved):

PR #1573, opened 2023-08-31
Collect free variables from `Type` field of `EPropGuards` in `FreeVars Expr` instance

qsctr:
Similar problem to #1570


qsctr:
Approved by @yav

PR #1579, opened 2023-10-09
Respect `VWord`/`VSeq` invariants in `parmap` implementation

### LINKED ISSUE #1437: Make `GenValue` abstract
The `GenValue` data type maintains the invariant that `VSeq` is never used for finite sequences of bits, as `VWord` is specialized for this purpose. It takes some care to ensure that this invariant is upheld in Cryptol, however. See #1435 for a recent example where we accidentally violated this invariant with a careless use of `VSeq`.

To help maintain this invariant, we should consider making `GenValue` abstract, only allowing the construction of `GenValue` values with smart constructors. For instance, the [`mkSeq`](https://github.com/GaloisInc/cryptol/blob/8cca24568ad499f06032c2e4eaa7dfd4c542efb6/src/Cryptol/Eval/Value.hs#L332-L339) smart constructor is intelligent enough to return a `VWord` or a `VSeq` depending on the element type, so using `mkSeq` everywhere would avoid potential issues like the one in #1435. We could also offer pattern synonyms for `GenValue` to allow pattern-matching directly on `VWord` and `VSeq`, but not the ability to construct them directly.

### LINKED ISSUE #1578: Use of parmap results in "not a word" error
I was trying out the experimental `parmap` function in a nightly Cryptol Docker image in the context of some recreational mathematics. Gist here:
https://gist.github.com/jpziegler/a3f0e8867edf92727dbd1451e4513d0e

Note that on line 31 I have replaced `map` (which works fine) with `parmap`.
https://gist.github.com/jpziegler/a3f0e8867edf92727dbd1451e4513d0e#file-ramsey_bug-cry-L31

The following invocation demonstrates the crash. However, most any invocation of the `R` function exhibits this behavior.

```
$ cryptol ramsey_bug.cry -c crash 
Loading module Cryptol
Loading module Main

  Revision:  UNKNOWN
  Branch:    UNKNOWN
  Location:  [Eval] fromVWord
  Message:   not a word
             cmpValue
             seq:93
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.0.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/Backend/Monad.hs:410:17 in cryptol-3.0.0.99-inplace:Cryptol.Backend.Monad
  evalPanic, called at src/Cryptol/Eval/Value.hs:388:23 in cryptol-3.0.0.99-inplace:Cryptol.Eval.Value
```

RyanGlScott:
Previously, `parmap` could incorrectly return a `VWord` when the element type was not `Bit`, and it could also return a `VSeq` when the element type was `Bit`. This changes the implementation of `parmap` to use the `mkSeq` smart constructor, which properly chooses what sort of `GenValue` to return depending on the element type.

Fixes #1578.


qsctr (approved):
Looks good to me, did you mean fixes #1578 instead of #1437?

RyanGlScott:
> did you mean fixes https://github.com/GaloisInc/cryptol/issues/1578 instead of https://github.com/GaloisInc/cryptol/issues/1437?

Oops, I had caught that mistake in the commit message, but not the PR description. I've now fixed the latter.

PR #1585, opened 2023-10-27
Fixes #1584

### LINKED ISSUE #1584: Messed up size accounting panic
The following expression causes a panic:
```
[ x+y | x <- [(0:Integer)...], y <- [1] ]
```
```
 Location:  lookupSeqMap
  Message:   Messed up size accounting
             Nat 1
             1
```

yav:
Fix incorrect evaluation of list comprehensions with infinite generators.  The problem happens when we have
an infinite generator, followed by a finite one.   For an example, see #1584


src/Cryptol/Eval.hs
```diff
@@ -742,27 +742,21 @@ evalMatch sym (lsz, lenv) m = seq lsz $ case m of
...
-        return (lsz * nLen, bindVarList n vs lenv')
+        return (nMul lsz len, bindVarList n vs lenv')
 
-      -- Select from a sequence of infinite length.  Note that this means we
-      -- will never need to backtrack into previous branches.  Thus, we can convert
-      -- `leVars` elements of the comprehension environment into `leStatic` elements
-      -- by selecting out the 0th element.
+      {- Select from a sequence of infinite length.  Note that only the
+         first generator in a sequence of generators may have infinite length,
+         so we can just evaluate it once an forall (i.e., it does not change
```
qsctr, line 749:
```suggestion
         so we can just evaluate it once and for all (i.e., it does not change
```

qsctr (approved):
Looks good, it might be good to add a test for the example in #1584 too.

RyanGlScott (approved):

Issue #1588, opened 2023-11-13
Sum types


RyanGlScott:
I propose that we add native support for sum types to the Cryptol language. This would be of particular use in two places:

1. When specifying algorithms that can possibly fail, one must currently indicate this with a return type that resembles `([32], Bit)`, where the `Bit` is `True` if the algorithm succeeded and `False` otherwise. This is clunky, however, since one must still provide a return value of type `[32]` in the `False` case, even though that return value is useless. What's worse, depending on how the algorithm is defined, returning a partial result of type `[32]` might potentially leak sensitive information that was provided to the algorithm.

   Both of these issues could be fixed by changing the return type to something like `Option u32`, where `Option` is a sum type that has one variant with a single field (in this case, of type `u32`) to represent success, and another variant with no fields at all to represent failure. This way, the failure case does not reveal anything about the inputs at all.
2. In SAW's MIR backend, it would be convenient to embed MIR `enum` values (which are effectively sum types themselves) into Cryptol and back. There isn't a particularly clean way to do this in general, aside from encoding sum types using some sort of Cryptol primitives. Having native sum types as part of Cryptol would make this process much smoother.

Having discussed this on @yav, we are both of the opinion that sum types would be useful to add, but we would first need to agree upon a particular design. Here are some specific design questions that would need to be answered:

1. Should sum types be nominal or structural? That is, should we implement tagged unions (nominal), perhaps resembling this Haskell-inspired syntax?

   ```
   data Foo
     = Bar [32]
     | Baz [64]
   ```
 
   Or should we implement anonymous variants (structural), similar to how Cryptol's records are anonymous?

   ```
   type Foo = ( [32] | [64] )
   ```

   @yav believes nominal sum types would be somewhat easier to implement, but also that structural sum types may be a better fit with the design of Cryptol.
2. If we implement nominal sum types, should we permit recursive types, such as in this example?

   ```
   data List
     = Nil
     | Cons [32] List
   ```

   @yav is inclined not to allow this.
3. If we implement structural sum types, should we permit subtyping? That is, should a value of type `( [32] | [64] )` be permitted to be used in a function that expects an argument of type `( [32] | [64] | [128] )`? Furthermore, should we permit _polymorphic_ variants such as `( [32] | [64] | ... )`, which indicate that other variants are permitted provided that `( [32] | [64] )` are included?

[...]

Issue #1587, opened 2023-11-11
Deriving instances for newtypes and enums


qsctr:
Newtypes currently do not work with any of the typeclasses which means you can't even check if two things are the same.
```
newtype T = { val : Integer }
```
```
Main> T { val = 1 } == T { val = 1 }

[error] at <interactive>:2:1--2:31:
  • Type `T` does not support equality.
      arising from
      use of expression (==)
      at <interactive>:2:1--2:31
```
It would be nice if the user could derive instances for classes based on the underlying type.

RyanGlScott:
A similar issue exists for enums (#1602). It might be nice to make certain classes derivable for enums (e.g., `Eq`).

RyanGlScott:
Here is a more detailed proposal. Throughout the example, I'll reference `enum Maybe a = Nothing | Just a`.

1. Newtype and enum declarations can have an optional `deriving (C_1, ..., C_n)` clause attached to the end, where each of `C_1` through `C_n` is a type of kind `# -> Prop`. Unlike Haskell, it is not valid to say `deriving C`; you must instead wrap it in parentheses, i.e., `deriving (C)`.
2. For enums, the only derivable constraints are `Eq`, `Cmp`, and `SignedCmp`.

   A derived `Eq` instance will consider two values of the same constructor to be equal iff their fields are also equal, and it considers two values of different constructors to be unequal. For example, `Nothing == Nothing ~> True`, `Nothing == Just 0b0 ~> False`, `Just 0b0 == Just 0b0 ~> True`, and `Just 0b0 == Just 0b1 ~> False`.

   Derived `Cmp` and `SignedCmp` instances obey a lexicographic ordering where earlier constructors are less than later constructors, and comparing two values of the same constructor will compare the fields pointwise. For instance, `Nothing < Just 0b0 ~> True`, `Just 0b0 < Nothing ~> False`, `Just 0b0 < Just 0b0 ~> False`, and `Just 0b0 < Just 0b1 ~> True`.
3. For newtypes, one can derive any class for which the underlying record type has a valid instance. This means that in addition to deriving `Eq`, `Cmp`, and `SignedCmp` instances, one could also derive `Zero`, `Logic`, and `Ring` instances.
4. The generated instances may have other conditions that need to be met in order to be valid. For instance, a derived `Eq (Maybe a)` instance would also require `Eq a`.

[...]

PR #1602, opened 2024-01-12
Sum types

yav:
This branch implements non-recursive nominal sum types for Cryptol. See also https://github.com/GaloisInc/cryptol/issues/1588.


docs/RefMan/TypeDeclarations.rst
```diff
@@ -51,3 +51,136 @@ of newtypes to extract the values in the body of the type.
...
+
+**The Matched Expression Must Have a Knwon Enum Type.** Cryptol will reject
+the definition of ``f``, where ``f`` lacks a type signature, or is not not
+an ``enum`` type.
+
+.. code-block:: cryptol
+
+  f x =
+    case x of
+      _ -> ()
```
RyanGlScott, line 174: This part reads a bit awkwardly, since it gives two scenarios where a function might be rejected, but the code example only applies to the first scenario. Perhaps we should provide two examples, one for each scenario. I'm thinking something like:

```
**The Matched Expression Must Have a Knwon Enum Type.** Cryptol will reject
the definition of ``f``, where ``f`` lacks a type signature: 

.. code-block:: cryptol
  f x =
    case x of
      _ -> ()

This is because it is not clear... [as before]

Cryptol will also reject this definition, where the type of the value 
being matched is not an ``enum`` type:

.. code-block:: cryptol
  g : Integer -> ()
  g x =
    case x of
      _ -> ()
```

src/Cryptol/Eval/Reference.lhs
```diff
@@ -541,24 +564,40 @@ the new bindings.
...
 
 At runtime, newtypes values are represented in exactly
 the same way as records.  The constructor function for
 newtypes is thus basically just an identity function
 that consumes and ignores its type arguments.
 
-> evalNewtypeDecl :: Env -> Newtype -> Env
-> evalNewtypeDecl env nt = bindVar (ntConName nt, pure val) env
+Enums are represented with a tag, which indicates what constructor
+was used to create a value, and the values of stored
```
RyanGlScott, line 578: "[...] and the values of stored" seems like an incomplete sentence.

src/Cryptol/Eval.hs
```diff
@@ -296,40 +314,65 @@ cacheCallStack sym v = case v of
...
+       pure (bindVarDirect (ecName c)
+                           (mkCon (foldr fu done (ecFields c) [])) env)
 
   tabs tp body =
     case tpKind tp of
       KType -> PTyPoly  (\ _ -> body)
       KNum  -> PNumPoly (\ _ -> body)
-      k -> evalPanic "evalNewtypeDecl" ["illegal newtype parameter kind", show (pp k)]
+      k ->
+        evalPanic "evalNewtypeDecl" [ "illegal newtype parameter kind"
```
RyanGlScott, line 361:
```suggestion
        evalPanic "evalNominalDecl" [ "illegal newtype parameter kind"
```

src/Cryptol/Eval/Reference.lhs
```diff
@@ -541,24 +564,40 @@ the new bindings.
...
+>       where
+>       tag                 = VEnum (nameIdent (ecName c)) . reverse
+>       addField _t mk args = VFun (\v -> pure (mk (v:args)))
+>
 >     tabs tp body =
 >       case tpKind tp of
 >         KType -> VPoly (\_ -> pure body)
 >         KNum  -> VNumPoly (\_ -> pure body)
->         k -> evalPanic "evalNewtypeDecl" ["illegal newtype parameter kind", show k]
+>         k -> evalPanic "evalNewtypeDecl"
```
RyanGlScott, line 599:
```suggestion
>         k -> evalPanic "evalNominalDecl"
```

src/Cryptol/Eval/Type.hs
```diff
@@ -39,12 +44,35 @@ data TValue
...
+  } deriving (Generic,NFData,Eq,Functor,Foldable,Traversable)
+
+isNullaryCon :: ConInfo a -> Bool
+isNullaryCon = Vector.null . conFields
+
+zipConInfo :: (a -> b -> c) -> ConInfo a -> ConInfo b -> ConInfo c
+zipConInfo f x y
+  | conIdent x == conIdent y =
+    x { conFields = Vector.zipWith f (conFields x) (conFields y) }
+  | otherwise = panic "zipConInfo" [ "Miamatched constructors"
```
RyanGlScott, line 70:
```suggestion
  | otherwise = panic "zipConInfo" [ "Mismatched constructors"
```

src/Cryptol/Eval/Value.hs
```diff
@@ -105,6 +113,10 @@ data EvalOpts = EvalOpts
 data GenValue sym
   = VRecord !(RecordMap Ident (SEval sym (GenValue sym))) -- ^ @ { .. } @
   | VTuple ![SEval sym (GenValue sym)]              -- ^ @ ( .. ) @
+  | VEnum !(SInteger sym) !(IntMap (ConValue sym))
+    -- ^ @Just 2@, the integer is the tag (e.g., `Just`), and the map contains
+    -- the fields (e.g., `{ 0 -> ("Just",2) }`.
+    -- The map is only really needed to represent symblic values.
```
RyanGlScott, line 119:
```suggestion
    -- The map is only really needed to represent symbolic values.
```

src/Cryptol/Eval/Value.hs
```diff
@@ -105,6 +113,10 @@ data EvalOpts = EvalOpts
 data GenValue sym
   = VRecord !(RecordMap Ident (SEval sym (GenValue sym))) -- ^ @ { .. } @
   | VTuple ![SEval sym (GenValue sym)]              -- ^ @ ( .. ) @
+  | VEnum !(SInteger sym) !(IntMap (ConValue sym))
+    -- ^ @Just 2@, the integer is the tag (e.g., `Just`), and the map contains
+    -- the fields (e.g., `{ 0 -> ("Just",2) }`.
```
RyanGlScott, line 118: The "integer is the tag" part is a bit confusing, since there are _two_ integers here: the tag (`0`) and the field of the constructor (`2`). Perhaps this would be clearer if you changed `2` to, say, `()`.

src/Cryptol/Eval/Value.hs
```diff
@@ -187,6 +213,19 @@ ppValue x opts = loop
...
+    case integerAsLit x i of
+      Just c ->
+        case IMap.lookup (fromInteger c) mp of
+          Just con
+            | isNullaryCon con -> pure (pp (conIdent con))
+            | otherwise ->
+              do vds <- traverse (>>= loop 1) (conFields con)
+                 let d = pp (conIdent con) <+> hsep (Vector.toList vds)
+                 pure (if prec > 0 then parens d else d)
+          Nothing -> panic "ppEnumVal" ["Malformed enumv value", show c]
```
RyanGlScott, line 226:
```suggestion
          Nothing -> panic "ppEnumVal" ["Malformed enum value", show c]
```

RyanGlScott:
With my most recent push, I believe all of the remaining work has been completed. The one thing that is missing is `deriving` support (https://github.com/GaloisInc/cryptol/issues/1587), but we can defer this for later.

@yav, do you have any review comments? The things that are new since the last time you looked are:

1. A mention of this work in `CHANGES.md`
2. Adding `Option` and `Result` to the Cryptol standard library
3. Documentation thereof in the reference manual

RyanGlScott (approved):

RyanGlScott:
The failing CI on https://github.com/GaloisInc/saw-script/pull/2020 in [this job](https://github.com/GaloisInc/saw-script/actions/runs/7756436609/job/21155024092?pr=2020#step:11:76) reveals a regression in the way Cryptol handles abstract types. Cryptol 3.0.0 accepts this:

```
module Either where

primitive type Either : * -> * -> *
primitive Left : {a, b} a -> Either a b
```

But it now rejects it with the `sum-types` branch:

```
$ cabal run exe:cryptol -- Either.cry
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻  
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃  
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.0.0.99 (a6b48b4, modified)

Loading module Cryptol
Loading module Either
[error] at Either.cry:4:30--4:40:
  Malformed type.
    Type synonym `Either::Either` was applied to 2 extra parameterss.
```

(There's also a typo in "parameterss".)

I'll mark this PR as a draft until we sort this out.

RyanGlScott:
Ah, I think the problem lies in this code:

https://github.com/GaloisInc/cryptol/blob/a9e35fe0049b042942b089078e0aa32040c413b7/src/Cryptol/TypeCheck/Kind.hs#L339-L361

The case that handles `Abstract` types only fires if `builtInType` returns `Just`. In the example above, however, `Example` is not a built-in type, so `builtInType` returns `Nothing`. Therefore, it falls through to the case that is supposed to handle newtypes and enums, which leads to disaster.

Relatedly, we ought to change the error messages for `TooFewTyParams` and `TooManyTySynParams` to say something besides "Type synonym" when the error involves a nominal type.

RyanGlScott:
I've pushed a fix for the issue described in https://github.com/GaloisInc/cryptol/pull/1602#issuecomment-1923996768 in commit https://github.com/GaloisInc/cryptol/pull/1602/commits/ca0eb7e064fb0cdc82431501a17f065c31f4c189.

PR #1614, opened 2024-01-23
use file-embed to make TH file embedding more robust

Ptival:
NOTE: We don't necessarily need to merge this. It does bring up a new library and removes the `Safe` flag from the modified file. It's useful for VSCode users, and may no longer be necessary in the near-to-medium future. At least, having this branch, I can help new users who run into the same issue.

---

Haskell Language Server still struggles a bit with opening projects that are subprojects of larger projects (in the git submodule sense).

When opening such projects from within their parent project, the overall project path remains the outermost one, which breaks relative paths of subprojects.

The file-embed has a utility to help with such path adjustments. Theoretically, HLS will eventually support "multi-home units", where different files will be compiled under different home folders if they belong to subprojects, but in the meantime, this un-breaks HLS for users that open Cryptol as a subproject of, say, SAW.


Ptival:
To explain this a little better, when you open the `saw-script` project, you may want to access files in `deps/cryptol/...`. But when you access these files, `Prelude.cry` will raise a compile-time exception, complaining that `lib/Cryptol.cry` does not exist. That's because, until multi-home units are well supported, the current working directory is still `saw-script/`, so `lib/Cryptol.cry` does not exist, but `deps/cryptol/lib/Cryptol.cry` does.

The `makeRelativeToProject` function takes care of making such adjustments where needed.

yav:
Seems like a reasonable change, removing the `Safe` pragma, and the new dependencies is not a big deal, `file-embed` only depends on things that we already depend on anyway, and it is very small.

RyanGlScott (approved):
I concur with @yav.

RyanGlScott:
[`file-embed-0.0.16.0`](https://hackage.haskell.org/package/file-embed-0.0.16.0) now includes a [`embedFileRelative x`](https://hackage.haskell.org/package/file-embed-0.0.16.0/docs/Data-FileEmbed.html#v:embedFileRelative) combinator that behaves the same way as `makeRelativeToProject x >>= embedFile`, which could be used to patch this patch slightly cleaner.

Ptival:
Nice find, updated!

RyanGlScott:
I've pushed a follow-up commit to regenerate the freeze files so that `cabal` can find build plans involving `file-embed-0.0.16.0`.

Issue #1061, opened 2021-02-04
macOS Universal Binaries


auricratio:
Guess who got a new laptop with Apple silicon. :)

No big hurry on this, Rosetta seems to be handling Intel Cryptol just fine.

atomb:
This blog post details when we can expect to be able to support the M1 natively: https://www.haskell.org/ghc/blog/20210309-apple-m1-story.html

robdockins:
GHC 9.2.1 dropped today, so we might be able to make progress on this before too long:
https://www.haskell.org/ghc/blog/20211029-ghc-9.2.1-released.html

chenrui333:
Any update on this thread?

RyanGlScott:
Not yet, but we should look into this for the next release. I can see multiple ways of approaching the problem:

* Have GitHub Actions build an AArch64 Mac build of Cryptol. This would require actions on GitHub's end first: https://github.com/actions/virtual-environments/issues/2187
* Develop tooling to build AArch64 Mac builds by hand. While slightly more cumbersome, this might be the way to go until GitHub Actions catches up.

chenrui333:
Sounds good, just checking. :)

RyanGlScott:
We now offer a ARM64 macOS binary on the [Cryptol downloads page](https://cryptol.net/downloads.html).

PR #1627, opened 2024-02-12
CI: Build and test both x86-64 and AArch64 macOS

RyanGlScott:
Fixes https://github.com/GaloisInc/cryptol/issues/1061.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,17 +64,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -280,6 +288,24 @@ jobs:
         with:
           ghc-version: '9.2.8'
 
+      # Homebrew installs packages in different directories depending on which
+      # macOS architecuture you are using:
```
kquick, line 292: architecture

kquick (approved):

PR #1633, opened 2024-02-23
Disable CI-based github page deployment for forks.

kquick:


.github/workflows/docs.yml
```diff
@@ -47,6 +47,7 @@ jobs:
             -p 'python3.withPackages (pp: [pp.sphinx pp.sphinx_rtd_theme])' \
             --run 'make html'
   build-pages-docs:
+    if: github.repository == "GaloisInc/cryptol"
```
RyanGlScott, line 50: Wouldn't it be better to use `github.event.pull_request.head.repo.fork == false` instead? That's how we configure it elsewhere in the CI:

https://github.com/GaloisInc/cryptol/blob/facc682fe81b319000b32eab46cae161a1fc47ae/.github/workflows/ci.yml#L207-L213

PR #1636, opened 2024-02-26
rem/enriching-properties

mccleeary-galois:
Adding additional notation to properties and new command to be able to check those properties with the options specified.

### DIFF
--- a/src/Cryptol/Parser.y
+++ b/src/Cryptol/Parser.y
@@ -71,6 +71,8 @@ import Paths_cryptol
   'parameter' { Located $$ (Token (KW KW_parameter) _)}
   'property'  { Located $$ (Token (KW KW_property)  _)}
   'check'     { Located $$ (Token (KW KW_check)     _)}
+  'prove'     { Located $$ (Token (KW KW_prove)     _)}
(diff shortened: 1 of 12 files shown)


weaversa:
May I suggest using doc comments to separate out REPL commands (like how Rust does doc tests -- https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html)?

```
/**
 * This property checks that ...
 *
 * ```
 * :check t0
 * ``` 
 */
property t0 = True

/**
 * This property demonstrates ... about the following polymorphic property.
 *
 * ```
 * :prove t1 `{a=1}
 * :prove t1 `{a=2}
 * ```
 */
t1 : {a} (1 <= a, a <= 2) => [a] -> Bit
property t1 x = True
```

eddywestbrook:
@weaversa One of the issues with the code fences approach that we were trying to talk about with you and @weeknightmvp the other day is that listing REPL commands doesn't really associate any structure to the assurance case. Instead, we are trying to structure the assurance case for properties so you can e.g. write queries over all properties that are checked but not proved, or all properties that are proved with Z3 and not yices (if your system doesn't have yices installed), etc.

Issue #1621, opened 2024-02-01
:check, :prove, etc. ignore the 'private' directive


weaversa:
Consider the example below where we have a private property `x` in a module `a`.  `a` is imported to `b`. `x` should not be available to `b`, but somehow `:check` has access when it shouldn't.

```a.cry
module a where

private

property x = True
```

```b.cry
module b where

import a
```

```cryptol
Cryptol> :m a
Loading module Cryptol
Loading module a
a> x
True
a> :m b
Loading module Cryptol
Loading module a
Loading module b
b> x
[error] at <interactive>:21:1--21:2
    Value not in scope: x
b> :check x
[error] at <interactive>:22:8--22:9
    Value not in scope: x
b> :check a::x
[error] at <interactive>:23:8--23:12
    Value not in scope: a::x
b> :check
property a::x Using exhaustive testing.
Passed 1 tests.
Q.E.D.
```

mccleeary-galois:
Reverting to handle #1646, reopening this issue.

PR #1642, opened 2024-03-14
Make it so only public IfaceDecls are loaded from other modules

mccleeary-galois:
#1621
Make it so only public IfaceDecls are loaded from other modules

### DIFF
--- a/src/Cryptol/ModuleSystem/Env.hs
+++ b/src/Cryptol/ModuleSystem/Env.hs
@@ -278,8 +278,7 @@ modContextOf mname me =
      let localIface  = lmInterface lm
          localNames  = lmNamingEnv lm
 
-         -- XXX: do we want only public ones here?
-         loadedDecls = map (ifDefines . lmInterface)
+         loadedPublicDecls = map getPublicDecls
                      $ getLoadedModules (meLoadedModules me)
 
          params = ifParams localIface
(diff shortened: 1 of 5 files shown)


weaversa:
Thank you for doing this! This issue was biting me earlier today.

mccleeary-galois:
CI failing due to https://github.com/GaloisInc/cryptol/issues/1643. Issue is unrelated to this PR.

src/Cryptol/ModuleSystem/Env.hs
```diff
@@ -304,6 +303,12 @@ modContextOf mname me =
        , mctxNames    = localNames
        , mctxNameDisp = R.toNameDisp localNames
        }
+  where getPublicDecls :: LoadedModule -> IfaceDecls
```
yav, line 306: Just some stylistic things that might be nice to change:
  * Avoid the reverse indentation in `do` blocks 
  * Don't mix composition and `$`. So instead of `f . g . h $ x` I generally tend to write `f (g (h x))`, or sometimes if things get too nested or I am trying to line up things for readability `f $ g $ h x`.

yav (approved):
Seems reasonable.  You could also probably remove the `XXX` in the signature version, as I don't think those have public/private things in them.  Or just do the same change there also, so they are consistent.

Issue #1643, opened 2024-03-18
`mypy --install-types` not freezing dependencies


mccleeary-galois:
CI is currently broken due to a dependency update from `mypy --install-types` 
This is needed due to https://github.com/GaloisInc/argo/pull/195#discussion_r1088437569

https://github.com/GaloisInc/cryptol/actions/runs/8324626968/job/22777200617

Current breaking change was on update from 
types-requests-2.31.0.20240310 [this successful log](https://github.com/GaloisInc/cryptol/actions/runs/8221150569/job/22481489129#step:13:41) to types-requests-2.31.0.20240311 [failing log](https://github.com/GaloisInc/cryptol/actions/runs/8230899271/job/22505655570#step:13:41)

We need to freeze these dependencies in some way, there is some information here for requirementes.txt which we do not use directly but the information here should be helpful regardless. https://github.com/python/mypy/issues/10600#issuecomment-857690947

PR #1644, opened 2024-03-19
Add types-requests to toml file.

mccleeary-galois:
#1643


cryptol-remote-api/python/pyproject.toml
```diff
@@ -14,10 +14,11 @@ include = [
 
 [tool.poetry.dependencies]
 python = ">=3.8.0,<4"
-requests = "^2.31.0"
+requests = "<2.31.0.20240310"
```
RyanGlScott, line 17: I think this constraint needs to apply to [`types-requests`](https://pypi.org/project/types-requests/) as well, since this constraint on `requests` alone doesn't appear to do the trick. (It's unclear to me what the relationship between the two packages' versions are, but better to constraint both, I'd say.)

cryptol-remote-api/python/pyproject.toml
```diff
@@ -14,10 +14,12 @@ include = [
...
 BitVector = "^3.4.9"
 argo-client = "0.0.11"
 typing-extensions = "^4.1.1"
+urllib3 = "<2.2.0"
+types-requests = "<2.31.0.20240310"
```
RyanGlScott, line 22: Leave a comment above these two dependencies stating that while the code itself doesn't directly depend on them, `mypy` _does_ depend on them, so we include them anyway to ensure that their versions are pinned. Worth citing https://github.com/python/mypy/issues/10600#issuecomment-857690947 as inspiration for this trick.

RyanGlScott (approved):
Hooray! One small suggestion, but otherwise this LGTM.

cryptol-remote-api/python/pyproject.toml
mccleeary-galois: Added in most recent version.

PR #1648, opened 2024-03-25
Fix typechecking of constraint guards with type signature constraints

### LINKED ISSUE #1647: Make `coreLint` sanity-check numeric constraint guards
Currently, the `coreLint` code for sanity-checking numeric constraint guards doesn't actually check any interesting properties:

https://github.com/GaloisInc/cryptol/blob/1e06a4b03d596c3c5f2f6908e8f4a08feacf31d2/src/Cryptol/TypeCheck/Sanity.hs#L367-L368

In particular, it would be nice if `coreLint` could check that the `guards` are well formed. Not doing so would have made it easier to detect issues such as https://github.com/GaloisInc/saw-script/issues/2043.

RyanGlScott:
Previously, `checkPropGuardCase` would not consider type signature constraints when determining how many proof applications to generate, which ultimately led to the issues observed in https://github.com/GaloisInc/saw-script/issues/2043. This is relatively easy to fix by ensuring that we pass these constraints as an additional argument to `checkPropGuardCase`.

This bug would have been caught if `coreLint` supported numeric constraint guards (#1647), so I have implemented `coreLint` support in a separate commit. As such, this PR fixes #1647.

It is difficult to fully validate that this fix works without SAW, but I have checked in a smoke test that checks if the guards in a numeric constraint guard definition have the expected number of proof applications.

### DIFF
--- a/src/Cryptol/TypeCheck/Infer.hs
+++ b/src/Cryptol/TypeCheck/Infer.hs
@@ -1266,7 +1266,7 @@ checkSigB b (Forall as asmps0 t0, validSchema) =
         P.DPropGuards cases0 -> do
           asmps1 <- applySubstPreds asmps0
           t1     <- applySubst t0
-          cases1 <- mapM checkPropGuardCase cases0
+          cases1 <- mapM (checkPropGuardCase asmps1) cases0
 
           exh <- checkExhaustive (P.bName b) as asmps1 (map fst cases1)
           unless exh $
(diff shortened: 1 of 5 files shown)


src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -1395,19 +1395,62 @@ checkExhaustive name as asmps guards =
...
-         rhsPs = foldl (\e _p -> EProofApp e) rhsTs     ps
-         rhs   = foldl EApp                   rhsPs     (map getV es)
+     let -- ...then apply the guard function to the kind-checked types...
+         rhsTs = foldl ETApp (getV eV) tys
+         -- ...then apply that to the constraints from the type signature
+         -- (`asmps`) and the constraints in the guard (`ps`). Note that `asmps`
+         -- is guaranteed to contain all of the non-guard constraints in scope,
+         -- as numeric constraint guards can only be used in top-level (i.e.,
+         -- non-local) functions...
+         rhsPs = foldl (\e _p -> EProofApp e) rhsTs (asmps ++ ps)
```
RyanGlScott, line 1450: Reviewer's note: changing `ps` to `asmps ++ ps` here is the key change.

yav (approved):
Looks good.  Thanks for fixing this!

PR #1656, opened 2024-04-17
Update libnuma in cryptol-remote-api Dockerfile

weaversa:
I was experimenting w/ shrinking the image and noticed that `apt autoremove` removes the `libnuma` .so file due to the dev version being installed. Here I change the Dockerfile to use the non-dev version of `libnuma`.

### DIFF
--- a/cryptol-remote-api/Dockerfile
+++ b/cryptol-remote-api/Dockerfile
@@ -89,7 +89,7 @@ RUN chown -R root:root /cryptol/rootfs
 
 FROM ubuntu:22.04
 RUN apt-get update \
-    && apt-get install -y libgmp10 libgomp1 libffi8 libncurses6 libtinfo6 libreadline8 libnuma-dev openssl \
+    && apt-get install -y libgmp10 libgomp1 libffi8 libncurses6 libtinfo6 libreadline8 libnuma1 openssl \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 RUN useradd -m cryptol && chown -R cryptol:cryptol /home/cryptol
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks for the patch!

Could you explain a bit further how you're pruning unused dependencies? I tried running `apt autoremove` as `root` in a `cryptol-remote-api` container, but that doesn't do anything when I try it:

```
$ docker run -u root --rm -it --entrypoint /bin/bash ghcr.io/galoisinc/cryptol-remote-api:3.1.0
root@f496157db665:/home/cryptol# apt autoremove
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

(I can certainly believe that we could remove some dependencies, however, so I'm open to your ideas on how we can do better here.)

RyanGlScott:
Revisiting this some time later, I now realize that the `libnuma1` dependency probably shouldn't be pruned. This is because due to a [GHC bug](https://gitlab.haskell.org/ghc/ghc/-/issues/20876#note_399802), the AArch64 version of GHC needs `libnuma1` to be installed in order for GHC installation to work. While we _could_ try to install `libnuma1` selectively based on the host architecture, that would likely complicate the Dockerfile logic to an undesirable degree.

I'm going to close this as wontfix, and I'll open a separate PR to clarify why `libnuma1` is here, taking inspiration from the comments in the SAW Dockerfile [here](https://github.com/GaloisInc/saw-script/blob/1b4a54ba54ed6e1858230a0cd7cc895ad4ba2c14/saw/Dockerfile#L13-L16).

PR #1659, opened 2024-05-14
fix #1653: distutils.find_executable -> shutil.which

### LINKED ISSUE #1653: Remove distutils import from cryptol-remote-api
distutils has been removed from the most recent versions of Python and we should remove uses of it

distutils.spawn.find_executable — use the shutil.which function

https://peps.python.org/pep-0632/#id11

Ptival:
@mccleeary-galois I think this would be the essence of the fix, but I don't know how cryptol-remote-api is tested.

Could consider changing [this](https://github.com/GaloisInc/cryptol/blob/be32aaae7c0bc6ef6b0e2a93027882ce926ce53c/.github/workflows/ci.yml#L467) to 3.12 to test it, though I don't know whether the rest of the dependencies are 3.12-compatible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.11'
+          python-version: '3.12'
 
       - uses: abatilo/actions-poetry@v2.0.0
         with:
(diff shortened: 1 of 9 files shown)


mccleeary-galois:
Going to add a commit to bump CI to 3.12 and unlock the toml file.

mccleeary-galois:
Attempt at trivial 3.12 bump failed see run here. Will need to cut a new ticket of 3.12 is target. https://github.com/GaloisInc/cryptol/actions/runs/9088621274/job/24978742509

mccleeary-galois:
Closing this PR as and pulling the commit into #1660

PR #1660, opened 2024-05-15
Update cryptol-remote-api to Python 3.12

### LINKED ISSUE #1645: Support latest version of `types-requests`
Currently `types-requests` is locked due to breaking CI, see #1643. We will want to eventually support the latest version.

### LINKED ISSUE #1653: Remove distutils import from cryptol-remote-api

mccleeary-galois:
Fix #1653 Fix #1645

This PR also bumps cryptol-remote-api to python 3.12. Added a note in the cryptol-remote-api readme for supported python versions.

### DIFF
--- a/cryptol-remote-api/python/README.md
+++ b/cryptol-remote-api/python/README.md
@@ -194,3 +194,8 @@ cryptol.connect(url="http://localhost:8080/", reset_server=True)
 
 will connect to a Cryptol server running at `http://localhost:8080/` and will
 guarantee any previous state on the server is cleared.
+
+
+## Supported Python Versions
+
+Currently `cryptol-remote-api supports` python `3.12` and is untested on other versions.
--- a/cryptol-remote-api/python/README.md
+++ b/cryptol-remote-api/python/README.md
@@ -198,4 +198,4 @@ guarantee any previous state on the server is cleared.
 
 ## Supported Python Versions
 
-Currently `cryptol-remote-api` supports python `3.12`. The `cryptol-remote-api` currently uses the default ubuntu 22.04 version and will be updated in issue #1661.
+Currently, `cryptol-remote-api` officially supports python `3.12`. The `cryptol-remote-api` Docker image currently uses Python 3.10 and will be updated in issue #1661.
--- a/cryptol-remote-api/python/README.md
+++ b/cryptol-remote-api/python/README.md
@@ -198,4 +198,4 @@ guarantee any previous state on the server is cleared.
 
 ## Supported Python Versions
 
-Currently `cryptol-remote-api supports` python `3.12` and is untested on other versions.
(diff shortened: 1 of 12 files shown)


RyanGlScott:
Thanks for working on this!

Regarding the `mypy` error, some things to try:

* The Python version in the `mypy.ini` is hardcoded to `3.7` [here](https://github.com/GaloisInc/cryptol/blob/be32aaae7c0bc6ef6b0e2a93027882ce926ce53c/cryptol-remote-api/python/mypy.ini#L3). We should try bumping that to 3.12. (Also, let's leave comments in all the places we set the Python versions that point to each other. That way, we remember to update them all in one go in the future.)
* If that fails, it may be worth upgrading the `mypy` version to see if a newer version is improved. (We are currently locking `mypy` to `0.991`, but the latest version is `1.10`.)

cryptol-remote-api/python/README.md
```diff
@@ -194,3 +194,8 @@ cryptol.connect(url="http://localhost:8080/", reset_server=True)
 
 will connect to a Cryptol server running at `http://localhost:8080/` and will
 guarantee any previous state on the server is cleared.
+
+
+## Supported Python Versions
+
+Currently `cryptol-remote-api supports` python `3.12` and is untested on other versions.
```
RyanGlScott, line 201: There is one caveat to this statement: in the [`cryptol-remote-api` Dockerfile](https://github.com/GaloisInc/cryptol/blob/be32aaae7c0bc6ef6b0e2a93027882ce926ce53c/cryptol-remote-api/Dockerfile#L23) (not to be confused with `test.Dockerfile`, which you modified elsewhere in this PR), we use the version of Python provided by Ubuntu 22.04 (currently 3.10.12). Perhaps we should migrate this Dockerfile over to Python 3.12 for consistency's sake.

Given that we want to land this relatively soon (and that the fix will work without needing to change that Dockerfile), I propose that we open a separate issue to track this.

mccleeary-galois: #1661

cryptol-remote-api/python/README.md
```diff
@@ -194,3 +194,8 @@ cryptol.connect(url="http://localhost:8080/", reset_server=True)
 
 will connect to a Cryptol server running at `http://localhost:8080/` and will
 guarantee any previous state on the server is cleared.
+
+
+## Supported Python Versions
+
+Currently `cryptol-remote-api` supports python `3.12`. The `cryptol-remote-api` currently uses the default ubuntu 22.04 version and will be updated in issue #1661.
```
RyanGlScott, line 201: Minor word-smithing:

```suggestion
Currently, `cryptol-remote-api` officially supports python `3.12`. The `cryptol-remote-api` Docker image currently uses Python 3.10 and will be updated in issue #1661.
```

RyanGlScott (approved):

PR #1664, opened 2024-05-17
Avoid discarding any docstring comments

glguy:
* include, import, and type constraint docstrings are now preserved
* parameter and private block docstrings are not supported

### DIFF
--- a/src/Cryptol/ModuleSystem/Base.hs
+++ b/src/Cryptol/ModuleSystem/Base.hs
@@ -430,6 +430,7 @@ addPrelude m
       , iAs      = Nothing
       , iSpec    = Nothing
       , iInst    = Nothing
+      , iDoc     = Nothing
       }
     }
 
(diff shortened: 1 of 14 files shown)


RyanGlScott:
Thanks for the PR, @glguy!

I realize that this is still a draft, but when you get a chance, could you open an issue describing what problem you are trying to solve? (If you can include one or two test cases showing what goes wrong with Cryptol's current behavior, all the better.) That way, this PR can reference the issue and provide better traceability into why we need to make the changes here.

yav:
@glguy do you think there's any reason to allow doc comments on imports/includes?   We can't really refer to them easily, and they don't really seem to be part of the API of a module...   I haven't looked carefully, but my guess is that they were there in the first place to avoid ambiguities in the grammar---I am thinking it might be easier to just report an error instead of saving them.

glguy:
I tracked them because they were there. I think it might make more sense not to allow them, however. 

Int place I could imagine them making sense is if you wanted to assert something about the way you instantiated a parameterized module because you're relying on properties if it specific to your instantiation. Is that plausible?

yav:
I think we should make them an error, and keep the AST simpler.  I think if you wanted to say something about instantiation you should probably do it where you define the instantiation.  Which reminds me that we may want to allow doc comments at the top of a module, to say what the module is for (we don't have to do it in this PR though)

Issue #1624, opened 2024-02-06
Support ghc 9.8


chenrui333:
While upgrading cryptol to the latest release 3.1.0, but run into some build issue. The error log is as below:

<details>
<summary>error build log</summary>

```
  ==> cabal v2-install --jobs=6 --max-backjumps=100000 --install-method=copy --installdir=/usr/local/Cellar/cryptol/3.1.0/bin
  Error: cabal: Could not resolve dependencies:
  [__0] trying: cryptol-3.1.0 (user goal)
  [__1] next goal: sbv (dependency of cryptol)
  [__1] rejecting: sbv-10.3 (conflict: cryptol => sbv>=9.1 && <10.3)
  [__1] trying: sbv-10.2
  [__2] next goal: base (dependency of cryptol)
  [__2] rejecting: base-4.19.0.0/installed-inplace (conflict: sbv => base>=4.16
  && <4.19)
  [__2] skipping: base-4.19.0.0 (has the same characteristics that caused the
  previous version to fail: excluded by constraint '>=4.16 && <4.19' from 'sbv')
  [__2] rejecting: base-4.18.2.0, base-4.18.1.0, base-4.18.0.0, base-4.17.2.1,
  base-4.17.2.0, base-4.17.1.0, base-4.17.0.0, base-4.16.4.0, base-4.16.3.0,
  base-4.16.2.0, base-4.16.1.0, base-4.16.0.0, base-4.15.1.0, base-4.15.0.0,
  base-4.14.3.0, base-4.14.2.0, base-4.14.1.0, base-4.14.0.0, base-4.13.0.0,
  base-4.12.0.0, base-4.11.1.0, base-4.11.0.0, base-4.10.1.0, base-4.10.0.0,
  base-4.9.1.0, base-4.9.0.0, base-4.8.2.0, base-4.8.1.0, base-4.8.0.0,
  base-4.7.0.2, base-4.7.0.1, base-4.7.0.0, base-4.6.0.1, base-4.6.0.0,
  base-4.5.1.0, base-4.5.0.0, base-4.4.1.0, base-4.4.0.0, base-4.3.1.0,
  base-4.3.0.0, base-4.2.0.2, base-4.2.0.1, base-4.2.0.0, base-4.1.0.0,
  base-4.0.0.0, base-3.0.3.2, base-3.0.3.1 (constraint from non-upgradeable
  package requires installed instance)
  [__2] fail (backjumping, conflict set: base, cryptol, sbv)
[...]

PR #1665, opened 2024-05-20
Support building with GHC 9.8

RyanGlScott:
This patch:

* Tightens up imports in `Cryptol.Eval.FFI.GenHeader` to avoid importing two different versions of `unzip` (note that GHC 9.8 defines `unzip` in `Data.Functor`, which is different from the version defined in `Data.List`).
* Bumps the upper version bounds on various dependencies (`base`, `sbv`, etc.) to support versions that are compatible with GHC 9.8.

Fixes https://github.com/GaloisInc/cryptol/issues/1624.


RyanGlScott:
Another error, uncovered when building the `cryptol-remote-api` Docker image ([here](https://github.com/GaloisInc/cryptol/actions/runs/9161055640/job/25185076745#step:8:2086)):

```
125.8 make --no-print-directory -f ghc.mk phase=0 phase_0_builds
125.8 ghc.mk:141: *** The make build system requires a boot compiler older than ghc-9.2. Your boot compiler is too new and cannot be used to build ghc-9.4 with make. Either boot with ghc 9.0.2 or build with hadrian. See https://www.haskell.org/ghc/blog/20220805-make-to-hadrian.html for advice on transitioning to hadrian..  Stop.
125.8 make: *** [Makefile:124: all] Error 2
```

cryptol-remote-api/cryptol-remote-api.cabal
```diff
@@ -40,19 +40,19 @@ common errors
 
 common deps
   build-depends:
-    base                 >=4.11.1.0 && <4.19,
+    base                 >=4.11.1.0 && <4.20,
     argo,
     aeson                >= 1.4.2 && < 2.3,
     base64-bytestring    >= 1.0,
-    bytestring           >= 0.10.8 && < 0.12,
+    bytestring           >= 0.10.8 && < 0.13,
```
kquick, line 47: I wonder if we could just stop specifying upper (or any) bounds on bytestring and text (and possibly any others listed at https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history) because since they are distributed as boot libraries along with GHC, they are effectively constrained by the `base` bounds?  This would mean fewer places needing an update when we do these version bumps.

src/Cryptol/IR/FreeVars.hs
```diff
@@ -52,9 +53,9 @@ rmVals p x = x { valDeps = Set.difference (valDeps x) p }
 -- | Compute the transitive closure of the given dependencies.
 transDeps :: Map Name Deps -> Map Name Deps
 transDeps mp0 = fst
-              $ head
+              $ headInfList
               $ dropWhile (uncurry (/=))
-              $ zip steps (tail steps)
+              $ zipTail steps
```
kquick, line 58: Could this just be `zip steps $ drop 1 steps` to avoid having to define `zipTail`?

src/Cryptol/ModuleSystem/Names.hs
```diff
@@ -16,14 +16,22 @@ import Cryptol.ModuleSystem.Name
...
     One x -> [x]
     Ambig ns -> Set.toList ns
 
 anyOne :: Names -> Name
-anyOne = head . namesToList
+anyOne xs =
+  case xs of
+    One x -> x
+    Ambig ns
+      |  x:_ <- Set.toList ns
```
kquick, line 31: Rather than converting the Set to a List and then only taking the first element (which is probably lazily cheap) would this be more direct as something like `Ambig ns -> if Set.null ns then panic .. else Set.elemAt 0 ns` ?

kquick (approved):
A couple of minor suggestions, but  looks good.

cryptol-remote-api/cryptol-remote-api.cabal
RyanGlScott: These boot library versions aren't wholly determined by `base`, as it's still possible to manually specify different versions of boot libraries than the ones that come included with GHC. (Indeed, some of our users do exactly this, as seen [here](https://github.com/GaloisInc/what4/pull/254).) Personally, I'd rather err on the side of caution here, since it is quite possible for major version bumps in boot libraries to break things (see, for instance, the `mtl-2.3.*`-related changes in #1572).

src/Cryptol/IR/FreeVars.hs
RyanGlScott: Yes, I like your version better.

src/Cryptol/ModuleSystem/Names.hs
RyanGlScott: Sure, I think your version works just as well. I'll change it.

PR #1669, opened 2024-05-22
CI: Don't download GHC in test jobs

### LINKED ISSUE #1668: CI: Downloading GHC on test jobs is wasteful
Currently, each CI job that runs part of the Cryptol test suite begins by downloading GHC, which takes several minutes. There is no need to do this, however, since the CI jobs that run the test suite never actually need to build any Haskell code. Instead, they download executables (`cryptol`, the test suite runner, etc.) that were built in a previous CI job and simply run those. As such, we should be able to avoid needing to download GHC at all on these jobs, which would likely save many CI minutes on each full run of the CI.

RyanGlScott:
Doing so is completely unnecessary, as we never actually need to build any Haskell code in these jobs (rather, we invoke pre-built executables instead).

Fixes #1668.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,10 +285,6 @@ jobs:
         with:
           submodules: true
 
-      - uses: haskell/actions/setup@v2
-        with:
-          ghc-version: '9.4.8'
-
       # Homebrew installs packages in different directories depending on which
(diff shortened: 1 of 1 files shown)


kquick (approved):
Nice! Looks like this is a considerable savings in the time taken for the test jobs.

Issue #1673, opened 2024-05-29
Make a dev_setup script


marsella:
Right now, users that want to build Cryptol from source have to follow the README to make sure they installed all their dependencies correctly. While this documentation is useful, it would be convenient to have a setup script that installs any missing dependencies needed to build and test the code base.

New releases are provided for 4 platforms; ideally, the setup script(s) will also support these platforms.
- MacOS 12 x86_64
- Ubuntu 20.04 ([sample `dev_setup.sh` script](https://github.com/GaloisInc/mctrace/blob/master/dev_setup.sh))
- Ubuntu 22.04
- Windows 2019

Eventually, these scripts can be integrated into the CI to ensure that local installations, containers, and the CI are all using the same dependency setup, but that's a large enough task to be a separate issue.

- [ ] Make a setup script for MacOS
- [ ] Make a setup script for Ubuntu 20.04 and 22.04
- [ ] Make a setup script for Windows 2019

RyanGlScott:
One minor correction to the list of platforms above: going forward, we want to support macOS 12 (x86_64) and macOS 14 (AArch64). I suspect that the scripts for both versions of macOS will largely be similar, but I can imagine there being meaningful differences due to the different architectures.

sauclovian-g:
In particular since homebrew uses different install paths on aarch64 it'll be hard to avoid getting entangled in that...

PR #1674, opened 2024-05-30
add dev_setup script for macOS 14 

marsella:
Partially addresses #1673 

This adds a `dev_setup.sh` script to install dependencies for macos 14. 
- I noted the places where I think things are different for macos 12
- Some of this transfers to ubuntu; I'd expect changes in the gmp (different package manager) and what4 (different what4 solvers URL) installation 

Some other questions for reviewers:
- Normally I would put this in `scripts/` but there isn't any such directory here. Is there a more appropriate place to keep it, or is in the root directory fine?

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -15,8 +15,13 @@ set -e
 HERE=$(cd `dirname $0`; pwd)
 LOG=$HERE/dev_setup.log
 
+WHAT4_SOLVERS_SNAPSHOT="snapshot-20240212"
+WHAT4_SOLVERS_URL="https://github.com/GaloisInc/what4-solvers/releases/download/$WHAT4_SOLVERS_SNAPSHOT/"
+WHAT4_SOLVERS_MACOS_12="macos-12-X64-bin.zip"
+WHAT4_SOLVERS_MACOS_14="macos-14-ARM64-bin.zip"
+
 function notice {
-    echo "[NOTICE] $*"
+    echo -e "[NOTICE] $*"
 }
 
 # Requires: LOG set to log file path.
@@ -37,6 +42,7 @@ function logged {
 }
 
 function update_submodules {
+    # todo: change names here
     cd $HERE
     notice "Updating submodules"
     git submodule update --init
@@ -55,45 +61,61 @@ function install_ghcup {
 
 }
 
+function is_macos {
+    [ "$OSTYPE" = "darwin"* ]
+}
+
 # Indicate whether this is running macOS on the Apple M series hardware.
 function is_macos_aarch {
     # Is it running macOS?
-    if [[ "$OSTYPE" != "darwin"* ]]; then
-        notice "dev setup does not currently support your OS / hardware"
-        exit 1 
+    if [ ! is_macos ]; then
+        return 1
     fi
 
     # Does it use an M-series chip?
+    # Actually this command apparently doesn't exist in macOS 12, so this will fail the wrong way
     chip=$(system_profiler SPHardwareDataType | grep Chip)
-    [[ $chip == *M1* || $chip == *M2* || $chip == *M3* ]]
+    [ $chip == *M1* || $chip == *M2* || $chip == *M3* ]
 }
 
 function install_gmp {
-    if is_macos_aarch; then
-        notice "Installing gmp via Homebrew, if it's not already installed"
-        logged brew list gmp || brew install gmp
+    if [ is_macos ]; then
+        notice "Installing GMP via Homebrew, if it's not already installed"
+        logged brew install gmp
 
-        # on macOS 12 (x86_64), I think homebrew uses different locations
+        # `brew --prefix` is different on macOS 12 and macOS 14
         notice "You may need to add the following environment variables to your '.profile': \n"\
-            "export CPATH=/opt/homebrew/include\n"\
-            "export LIBRARY_PATH=/opt/homebrew/lib\n"
+            "export CPATH=$(brew --prefix)/include\n"\
+            "export LIBRARY_PATH=$(brew --prefix)/lib\n"
+    else
+        notice "Did not install GMP. This script only supports macOS 12 and 14"
     fi
 }
 
+# Installs the two solvers required to run the test suite for the repo.
+# Users may want to install other solvers, and indeed the what4 solvers repo
+# includes a half dozen other solvers that are compatible with cryptol. 
 function install_what4_solvers {
     if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
         notice "Installing cvc4 and/or cvc5 solvers"
 
-        # There are different URLs for other supported platforms
-        if is_macos_aarch; then
-            what4_solvers_url="https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20240212/macos-14-ARM64-bin.zip"
+        if [ ! is_macos ]; then
+            notice "Did not install what4 solvers. This script only supports macOS 12 and 14"
+            return
+        fi
+
+        if [ is_macos_aarch ]; then
+            solvers_version=$WHAT4_SOLVERS_MACOS_14
+        else
+            # This assumes that developers have read the docs and only run this
+            # script if they're on 12 or 14. This might bork on older versions.
+            solvers_version=$WHAT4_SOLVERS_MACOS_12
         fi
 
         solvers_dir=$(mktemp -d)
-        curl --proto '=https' --tlsv1.2 -sSfL $what4_solvers_url > "$solvers_dir/solvers.bin.zip"
+        curl --proto '=https' --tlsv1.2 -sSfL "$WHAT4_SOLVERS_URL$solvers_version" > "$solvers_dir/solvers.bin.zip"
         cd $solvers_dir
         logged unzip solvers.bin.zip
-        rm solvers.bin.zip
         
         # If we want to install more solvers by default, we can do so here,
         # although we might need a different check than `--version`
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -25,6 +25,9 @@ VAR_FILE=$ROOT/env.sh
 truncate -s 0 $VAR_FILE
 echo "# Generated environment variables. Manual changes will get clobbered by dev_setup.sh" >> $VAR_FILE
 
+GHC_VERSION="9.4.8"
+CABAL_VERSION="3.10.3.0"
+
 WHAT4_SOLVERS_SNAPSHOT="snapshot-20240212"
 WHAT4_SOLVERS_URL="https://github.com/GaloisInc/what4-solvers/releases/download/$WHAT4_SOLVERS_SNAPSHOT/"
 WHAT4_SOLVERS_MACOS_12="macos-12-X64-bin.zip"
@@ -93,7 +96,7 @@ function update_submodules {
 
 function install_ghcup {
     if ! is_installed ghcup; then
-        notice "Installing GHC toolchain via GHCup"
+        notice "Installing GHCup via Homebrew"
         # Manually install Haskel toolchain. If this doesn't work, try
         # using the install script at https://www.haskell.org/ghcup/
         if [ $CRYPTOL_PLATFORM = $MACOS12 ] || [ $CRYPTOL_PLATFORM = $MACOS14 ]; then
@@ -102,15 +105,17 @@ function install_ghcup {
         else
             notice "Did not install GHCup. This script only supports macOS 14"
         fi
-
-        logged ghcup install ghc --set recommended
-        logged ghcup install cabal --set recommended
-        logged ghcup install stack --set recommended
-        logged ghcup install hls --set recommended
-        logged cabal update
     else
-        notice "Using existing GHCup installation, and assuming that if you"\
-            "have GHCup, you also have the rest of the Haskell toolchain"
+        notice "Using existing GHCup installation"
+    fi
+
+    notice "Installing ghc and cabal via GHCup"
+    logged ghcup install ghc $GHC_VERSION
+    logged ghcup install cabal $CABAL_VERSION
+    logged ~/.ghcup/bin/cabal-$CABAL_VERSION update
+
+    if ! is_installed ghc || ! is_installed cabal; then
+        echo "export PATH=$PATH:~/.ghcup/bin" >> $VAR_FILE
     fi
 
 }
@@ -190,8 +195,6 @@ function put_brew_in_path {
         # `brew --prefix` is different on macOS 12 and macOS 14
         echo "export CPATH=$(brew --prefix)/include" >> $VAR_FILE
         echo "export LIBRARY_PATH=$(brew --prefix)/lib" >> $VAR_FILE
-        notice "You may need to source new environment variables added here:\n" \
-            "\$ source $VAR_FILE"
     fi
 }
 
@@ -208,3 +211,6 @@ install_ghcup
 install_gmp
 install_solvers
 put_brew_in_path
+
+notice "You may need to source new environment variables added here:\n" \
+    "\$ source $VAR_FILE"
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+
+# cryptol setup script
+#
+# Supported distribution(s):
+#  * macOS 14 (AArch64)
+#
+# This script installs everything needed to get a working development
+# environment for cryptol. Any new environment requirements should be
+# added to this script.
+#
+
+set -e
+
+HERE=$(cd `dirname $0`; pwd)
+LOG=$HERE/dev_setup.log
+
+function notice {
+    echo "[NOTICE] $*"
+}
+
+# Requires: LOG set to log file path.
+function logged {
+    if ! [ -z "$LOG" ]
+    then
+        mkdir -p `dirname $LOG`
+        echo $* >>$LOG
+        if ! $* >>$LOG 2>&1
+        then
+            echo
+            echo "An error occurred; please see $LOG"
+            echo "Here are the last 50 lines:"
+            tail -n 50 $LOG
+            exit 1
+        fi
+    fi
+}
+
+function update_submodules {
+    cd $HERE
+    notice "Updating submodules"
+    git submodule update --init
+}
+
+function install_ghcup {
+    if ! ghcup --version &> /dev/null
+    then
+        notice "Installing ghcup, GHC, and cabal"
+        # technically the installation only requires cabal, but it's 
+        # recommended to get the whole GCH shebang in one package
+        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+    else
+        notice "Using existing ghcup installation"
+    fi
+
+}
+
+# Indicate whether this is running macOS on the Apple M series hardware.
+function is_macos_aarch {
+    # Is it running macOS?
+    if [[ "$OSTYPE" != "darwin"* ]]; then
+        notice "dev setup does not currently support your OS / hardware"
+        exit 1 
+    fi
+
+    # Does it use an M-series chip?
+    chip=$(system_profiler SPHardwareDataType | grep Chip)
+    [[ $chip == *M1* || $chip == *M2* || $chip == *M3* ]]
+}
+
+function install_gmp {
+    if is_macos_aarch; then
+        notice "Installing gmp via Homebrew, if it's not already installed"
+        logged brew list gmp || brew install gmp
+
+        # on macOS 12 (x86_64), I think homebrew uses different locations
+        notice "You may need to add the following environment variables to your '.profile': \n"\
+            "export CPATH=/opt/homebrew/include\n"\
+            "export LIBRARY_PATH=/opt/homebrew/lib\n"
+    fi
+}
+
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
+
+        # There are different URLs for other supported platforms
+        if is_macos_aarch; then
+            what4_solvers_url="https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20240212/macos-14-ARM64-bin.zip"
+        fi
+
+        solvers_dir=$(mktemp -d)
+        curl --proto '=https' --tlsv1.2 -sSfL $what4_solvers_url > "$solvers_dir/solvers.bin.zip"
+        cd $solvers_dir
+        logged unzip solvers.bin.zip
+        rm solvers.bin.zip
+        
+        # If we want to install more solvers by default, we can do so here,
+        # although we might need a different check than `--version`
+        for solver in cvc4 cvc5; do
+            if ! $solver --version &> /dev/null; then 
+                notice "Installing $solver"
+                chmod u+x $solver
+                sudo mv $solver /usr/local/bin
+            fi
+        done
+    fi
+}
+
+update_submodules
+install_ghcup
+install_gmp
+install_what4_solvers
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -28,29 +28,40 @@ WHAT4_SOLVERS_SNAPSHOT="snapshot-20240212"
 WHAT4_SOLVERS_URL="https://github.com/GaloisInc/what4-solvers/releases/download/$WHAT4_SOLVERS_SNAPSHOT/"
 WHAT4_SOLVERS_MACOS_12="macos-12-X64-bin.zip"
 WHAT4_SOLVERS_MACOS_14="macos-14-ARM64-bin.zip"
+WHAT4_CVC4_VERSION="version 1.8"
+WHAT4_CVC5_VERSION="version 1.1.1"
 
+# Set of supported platforms:
 MACOS14="macos14"
-MACOS12="macos12"
-# Make sure we're running on a supported platform
-case $(uname -s) in
-    Darwin)
-        if [ $(uname -m) = "arm64" ]; then
-            CRYPTOL_PLATFORM=$MACOS14
-        # This is how we'd support macOS 12. Since this hasn't been tested yet,
-        # we withhold official support.
-        # This might bork on something running macOS <12, since we're basing
-        # the it on the hardware, not the specific version.
-        elif [ $(uname -m) = "x86_64" ]; then
-            CRYPTOL_PLATFORM=$MACOS12
-            echo "Unsupported platform" 2>&1; exit 1;
-        else
-            echo "Unsupported platform" 2>&1; exit 1;
-        fi;;
-    *) echo "Unsupported platform" 2>&1; exit 1;;
-esac
+MACOS12="macos12" # actually, this isn't supported yet
 
+# Returns a string indicating the platform (from the set above), or empty if
+# a supported platform is not detected.
+function supported_platform {
+    # Make sure we're running on a supported platform
+    case $(uname -s) in
+        Darwin)
+            if [ $(uname -m) = "arm64" ]; then
+                echo $MACOS14
+            # This is how we'd support macOS 12. Since this hasn't been tested yet,
+            # we withhold official support.
+            # This might bork on something running macOS <12, since we're basing
+            # the it on the hardware, not the specific version.
+            elif [ $(uname -m) = "x86_64" ]; then
+                # CRYPTOL_PLATFORM=$MACOS12
+                echo ""
+            fi;;
+        *) echo ""
+    esac
+}
+
+RED="\033[0;31m"
 function notice {
     echo "[NOTICE] $*"
+}
+
+function is_installed {
+    command -v $* >/dev/null 2>&1
 }
 
 # Requires: LOG set to log file path.
@@ -63,9 +74,10 @@ function logged {
         if ! "$@" >>$LOG 2>&1
         then
             echo
-            echo "An error occurred; please see $LOG"
-            echo "Here are the last 50 lines:"
             tail -n 50 $LOG
+            echo
+            echo "[ERROR] An error occurred; please see $LOG"
+            echo "[ERROR] The last 50 lines are printed above for convenience"
             exit 1
         fi
     fi
@@ -78,13 +90,24 @@ function update_submodules {
 }
 
 function install_ghcup {
-    if ! ghcup --version &> /dev/null
+    if ! is_installed ghcup
     then
-        notice "Installing ghcup, GHC, and cabal"
+        ghcup_url="https://get-ghcup.haskell.org"
+        notice "ghcup not found; do you want to install from $ghcup_url"
+        read -p "Press Enter to continue or 'n' to skip: " consent
+        case $consent in
+            [Nn]* ) notice "Skipping ghcup installation"; return;;
+            * ) notice "Installing ghcup";;
+        esac
+
         # Technically the installation only requires cabal, but it's
         # recommended to get the whole GCH shebang in one package.
         # The output is not routed to log because the installer is interactive.
-        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+        ghc_dir=$(mktemp -d)
+        logged curl --proto '=https' --tlsv1.2 -sSf -o "$ghc_dir/ghcup.bin" $ghcup_url
+        cd $ghc_dir
+        sh ghcup.bin
+        rm -r $ghc_dir
     else
         notice "Using existing ghcup installation"
     fi
@@ -110,8 +133,8 @@ function install_gmp {
 # Users may want to install other solvers, and indeed the what4 solvers repo
 # includes a half dozen other solvers that are compatible with cryptol. 
 function install_what4_solvers {
-    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
-        notice "Installing cvc4 and/or cvc5 solvers"
+    if ! is_installed cvc4 || ! is_installed cvc5; then
+        notice "Installing cvc4 and/or cvc5 solvers from $WHAT4_SOLVERS_URL"
 
         if [ $CRYPTOL_PLATFORM = $MACOS14 ]; then
             solvers_version=$WHAT4_SOLVERS_MACOS_14
@@ -124,10 +147,9 @@ function install_what4_solvers {
         cd $solvers_dir
         logged unzip solvers.bin.zip
         
-        # If we want to install more solvers by default, we can do so here,
-        # although we might need a different check than `--version`
+        # If we want to install more solvers by default, we can do so here
         for solver in cvc4 cvc5; do
-            if ! $solver --version &> /dev/null; then 
+            if ! is_installed $solver ; then
                 notice "Installing $solver"
                 chmod u+x $solver
                 sudo mv $solver /usr/local/bin
@@ -138,16 +160,28 @@ function install_what4_solvers {
     else
         notice "Not installing cvc4 or cvc5 solvers because they already exist"
 
-        if ! (grep -q "version 1.8" <<< "$(cvc4 --version)"); then
-            notice "Your version of cvc4 is incorrect; expected 1.8"
+        if ! (grep -q $WHAT4_CVC4_VERSION <<< "$(cvc4 --version)"); then
+            notice "Your version of cvc4 is unexpected; expected $WHAT4_CVC4_VERSION"
             notice "Got: $(grep "cvc4 version" <<< "$(cvc4 --version)")"
+            notice "To ensure compatibility, you might want to uninstall the "\
+                "existing version and re-run this script."
         fi
-        if ! (grep -q "version 1.1.1" <<< "$(cvc5 --version)" ); then
-            notice "Your version of cvc5 is incorrect; expected 1.1.1"
+        if ! (grep -q $WHAT4_CVC5_VERSION <<< "$(cvc5 --version)" ); then
+            notice "Your version of cvc5 is unexpected; expected $WHAT5_CVC5_VERSION"
             notice "Got: $(grep "cvc5 version" <<< "$(cvc5 --version)")"
+            notice "To ensure compatibility, you might want to uninstall the "\
+                "existing version and re-run this script."
         fi
     fi
 }
+
+
+# Make sure script is running on a supported platform
+CRYPTOL_PLATFORM=$(supported_platform)
+if [ -z "$CRYPTOL_PLATFORM" ]; then
+    echo "Unsupported platform"
+    exit 1
+fi
 
 update_submodules
 install_ghcup
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -7,7 +7,8 @@
 #
 # This script installs everything needed to get a working development
 # environment for cryptol. Any new environment requirements should be
-# added to this script.
+# added to this script. This script is not interactive but it may define
+# some environment variables that need to be loaded after it runs.
 #
 # There is some half-baked support for macOS 12 (x86_64), but it hasn't been
 # tested.
@@ -35,6 +36,8 @@ WHAT4_CVC5_VERSION="version 1.1.1"
 MACOS14="macos14"
 MACOS12="macos12" # actually, this isn't supported yet
 
+USED_BREW=false
+
 # Returns a string indicating the platform (from the set above), or empty if
 # a supported platform is not detected.
 function supported_platform {
@@ -48,7 +51,6 @@ function supported_platform {
             # This might bork on something running macOS <12, since we're basing
             # the it on the hardware, not the specific version.
             elif [ $(uname -m) = "x86_64" ]; then
-                # CRYPTOL_PLATFORM=$MACOS12
                 echo ""
             fi;;
         *) echo ""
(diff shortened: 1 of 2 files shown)


dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+        notice "Using existing ghcup installation"
+    fi
+
+}
+
+# Indicate whether this is running macOS on the Apple M series hardware.
+function is_macos_aarch {
+    # Is it running macOS?
+    if [[ "$OSTYPE" != "darwin"* ]]; then
+        notice "dev setup does not currently support your OS / hardware"
```
jtdaugherty, line 62: For this error, I recommend including in the error the platform(s) that the script supports. If I saw this error, I'd immediately ask "what platform do I need to be on to get past this error?"

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+        exit 1
```
jtdaugherty, line 63: A thought on flow control here: in my personal approach, I'd typically write functions like `is_macos_aarch` as purely for flow control at the call site, but never to kill the script or emit output (which could be redirected or produced at the caller's surprise). Since this `exit` is for non-Darwin hosts, I would instead add `is_darwin` and then have the top-level script invoke `exit` when `is_darwin` is false. That keeps the functions constrained to just getting the diagnostic info needed, and pushes policy decisions about what to do with that information to the top level of the script where it is easier to discover and change.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+    # Does it use an M-series chip?
+    chip=$(system_profiler SPHardwareDataType | grep Chip)
+    [[ $chip == *M1* || $chip == *M2* || $chip == *M3* ]]
+}
+
+function install_gmp {
+    if is_macos_aarch; then
+        notice "Installing gmp via Homebrew, if it's not already installed"
+        logged brew list gmp || brew install gmp
```
jtdaugherty, line 74: `logged` is missing on the second call to `brew` here, if you were intending to redirect all `brew` output to the log here.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+        # on macOS 12 (x86_64), I think homebrew uses different locations
+        notice "You may need to add the following environment variables to your '.profile': \n"\
```
jtdaugherty, line 77: This is a slightly bigger-picture comment, but I would recommend putting these environment variables in, say, `$repo_root/env.sh` and then having those sourced as a matter of course when working with the project. My thinking behind this suggestion is to centralize all environment requirements within the physical confines of the repo rather than asking the user to update (say) `~/.profile`, which can be easy to forget as an integral part of getting a working system set up.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    fi
+}
+
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
+
+        # There are different URLs for other supported platforms
+        if is_macos_aarch; then
+            what4_solvers_url="https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20240212/macos-14-ARM64-bin.zip"
```
jtdaugherty, line 89: I recommend moving this string literal to a top-level variable and using that here, e.g. `WHAT4_SOLVERS_URL`. The objective behind this is to float all settings to the top of the script where they are easier to locate and change, and to avoid deep embedding in the source as is done here.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+        # on macOS 12 (x86_64), I think homebrew uses different locations
+        notice "You may need to add the following environment variables to your '.profile': \n"\
+            "export CPATH=/opt/homebrew/include\n"\
+            "export LIBRARY_PATH=/opt/homebrew/lib\n"
+    fi
+}
+
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
```
jtdaugherty, line 84: Is there any more specific version of cvc4 or cvc5 that needs to be checked for? (Maybe not, but I'm asking just in case.) Like, maybe you want `>= cvc4 4.1.2` or something like that. If so, it might be important to enrich these checks from just `$PATH` checks to version checks.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+        solvers_dir=$(mktemp -d)
+        curl --proto '=https' --tlsv1.2 -sSfL $what4_solvers_url > "$solvers_dir/solvers.bin.zip"
+        cd $solvers_dir
+        logged unzip solvers.bin.zip
+        rm solvers.bin.zip
+        
+        # If we want to install more solvers by default, we can do so here,
+        # although we might need a different check than `--version`
+        for solver in cvc4 cvc5; do
```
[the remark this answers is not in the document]
jtdaugherty: Reading further, I guess that suggestion doesn't make sense for this second part where you are installing *all* supported CVCs.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    git submodule update --init
+}
+
+function install_ghcup {
+    if ! ghcup --version &> /dev/null
+    then
+        notice "Installing ghcup, GHC, and cabal"
+        # technically the installation only requires cabal, but it's 
+        # recommended to get the whole GCH shebang in one package
+        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
```
jtdaugherty, line 51: It might be helpful to comment here that this doesn't use `logged` because the GHCUP installation process is interactive.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+            echo "An error occurred; please see $LOG"
+            echo "Here are the last 50 lines:"
+            tail -n 50 $LOG
+            exit 1
+        fi
+    fi
+}
+
+function update_submodules {
+    cd $HERE
```
jtdaugherty, line 40: When I write these scripts I use `$HERE` to mean "the place where this script is located." That is sometimes not the same as the repository root, but it is used as a repository root here. I recommend adding an additional variable at the top, adjacent to the declaration of `HERE`, as `ROOT=$HERE` and then using `$ROOT` here. That way, if this script gets `git mv`'d somewhere else, only one line needs to be changed (the value of `ROOT`, relative to `HERE`).

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
```
sauclovian-g, line 1: In general I'd recommend sticking to #!/bin/sh and avoiding any bashisms (unless there's some specific reason to need bash, which seems unlikely)

... otherwise "install bash" can end up being a prereq for running the setup script and we don't want that.

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    else
+        notice "Using existing ghcup installation"
+    fi
+
+}
+
+# Indicate whether this is running macOS on the Apple M series hardware.
+function is_macos_aarch {
+    # Is it running macOS?
+    if [[ "$OSTYPE" != "darwin"* ]]; then
```
sauclovian-g, line 61: [[ ]] with patterns is a bashism (but you probably want to make this a case statement anyway once there's more than one OS)

sauclovian-g (changes requested):
Some suggestions, mostly fairly minor, but the quoting stuff does matter...

sauclovian-g:
also note that my review and Jonathan's were simultaneous and probably have duplicate things in them...

sauclovian-g:
on the scripts/ front, I think we probably do want a subdir to hold developer-facing utility material... but probably most of the stuff we'll be putting in it probably isn't scripts. So I dunno what to call it and probably it should wait until we have more stuff...

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+function notice {
+    echo "[NOTICE] $*"
+}
+
+# Requires: LOG set to log file path.
+function logged {
+    if ! [ -z "$LOG" ]
+    then
+        mkdir -p `dirname $LOG`
```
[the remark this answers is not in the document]
jtdaugherty: It might be, but not if `LOG` points to something other than `HERE` (which has been the case in some of my scripts).

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+        echo $* >>$LOG
+        if ! $* >>$LOG 2>&1
```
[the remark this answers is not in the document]
jtdaugherty: Thanks, this code was brought over from my original script. :)

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
+
+        # There are different URLs for other supported platforms
+        if is_macos_aarch; then
+            what4_solvers_url="https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20240212/macos-14-ARM64-bin.zip"
+        fi
+
+        solvers_dir=$(mktemp -d)
+        curl --proto '=https' --tlsv1.2 -sSfL $what4_solvers_url > "$solvers_dir/solvers.bin.zip"
```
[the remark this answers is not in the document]
jtdaugherty: I don't think so, since the output gets redirected to a zip file here. Although it might produce stderr, but `logged` captures that and stdout as well.

dev_setup.sh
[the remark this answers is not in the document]
sauclovian-g: good point

(probably worth sticking a comment in)

dev_setup.sh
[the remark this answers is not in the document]
sauclovian-g: hmm

So in general it would be better to log it... in principle it should be possible to pass ```logged``` a command that includes I/O redirection, and then wrapping that in more should just not collect anything from stdout, but making that work correctly will be a headache, especially in combination with quoting.

oh, curl has an -o option to set the output file. so probably the solution is to use that :-)

dev_setup.sh
jtdaugherty: I suggest this partly to encourage standardization of essential settings (and people could always have local files to customize those). Mainly, to avoid, hat in hand, having to tell the user to set variables just to get basic project tasks done. But also, I'm assuming that such an `env.sh` can be written to capture all the assumptions about repo structure, which is then hopefully stable across all development environments and can be maintained by the whole team. If you don't have that, then my `.profile` can become stale because of a change someone makes to a script in the repo.

dev_setup.sh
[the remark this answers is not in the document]
jtdaugherty: Ah nice, in that case I'd vote +1 on `-o` and then use `logged`.

dev_setup.sh
jtdaugherty: (To summarize it a little more concisely, I like to find ways to make sure that the repo owns the definition of what a complete setup looks like and how it works as one source of truth, since spreading out that truth over both the repo and local files can result in builds that should work but don't, or do but shouldn't.)

dev_setup.sh
```diff
@@ -0,0 +1,113 @@
...
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
+
+        # There are different URLs for other supported platforms
+        if is_macos_aarch; then
+            what4_solvers_url="https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20240212/macos-14-ARM64-bin.zip"
+        fi
+
+        solvers_dir=$(mktemp -d)
```
jtdaugherty, line 92: I just realized that this doesn't get cleaned up.

dev_setup.sh
sauclovian-g: I can't say I disagree really; it's just that unless you only support exactly identically cloned environments, there's some amount of tree config that is going to be (in general) different for everyone. I guess the actual issue is that it shouldn't be done with environment variables.

In my world, there's a place in the tree that is not managed by source control where you put your local tree config, and it's structured so that if you don't have anything there you get reasonable defaults, and otherwise there's a reference for what you can set and what the pertinent considerations are. Basically this evolved from the premise that anything should build with ```./configure && make``` with the observation that well-written (and thus non-autoconf) configure scripts produce an output file that a clueful user can equally well manage by hand. But it relies on having build systems that are adequately expressive to capture this kind of stuff in a non-hacky way. E.g. for this issue in one of my own projects the config would probably be "put ```GMP_PREFIX=/opt/homebrew``` in src/config.mk". I don't think there's a good way to get cabal to do this kind of thing (but I'd be happy to wrong about that!) so it becomes a tradeoff between what's nice and what's reasonably expedient.

dev_setup.sh
marsella: I added some functions for checking which OS we support, since just `$OSTYPE` isn't enough to distinguish between, e.g., macos 12 and 14 due to having to check the hardware in a separate command. I am not sure if ubuntu 20 vs 22 would be caught by this either.

dev_setup.sh
marsella: There are compatible versions specified in the solvers repo, but I'm not sure if those are the minimum required or the latest version that's been tested. I also wasn't sure what to do in the case that there was an older version in the system; it seems rude to clobber it without checking with the user. I could check the version and print a warning if it's out of date?

dev_setup.sh
jtdaugherty: @sauclovian-g I'm not really sure what to say to that in response; while I don't want to question the merits of your approach and past experiences, I would still like my original request here to stand on the grounds that the environment variables being mentioned are, it would seem, a requirement for working with the project and so they strike me as the sort of thing you would put with the project. I'm also hoping we can get to some resolution on this to avoid confusion for @marsella if there's conflicting messages in this discussion.

dev_setup.sh
jtdaugherty: Hmm. Maybe the solvers repo ought to have its own scripting to manage the solver installation, so that there's one source of truth for that? This could them become 1) fetch solvers tarball, 2) unpack, 3) run a script from within that archive that Just Works. And then that other script would have to figure out what to do in the case of an older version already being installed (and yeah, I thought as you do, to warn about an older version rather than clobbering it).

dev_setup.sh
marsella: I added this but it's a little messy -- repeated runs of the script will re-append the same two lines repeatedly. Would it be reasonable to delete & re-create that file each time we run the setup script or do we want to account for the possibility that someone will separately put their own environment vars in there?

jtdaugherty: @marsella I'm not sure what you're referring to (I don't see any new commits). Can you say a bit more about what you mean by "re-append the same two lines" and "re-create that file"?

dev_setup.sh
marsella: I took another pass and tried to clean up control flow! Took out all the exits and tried to be more consistent throughout about giving a notice and moving on.

dev_setup.sh
marsella: I took this and @sauclovian-g 's similar suggestion and perhaps veered too far into modularization, but all the constants are now top level variables.

dev_setup.sh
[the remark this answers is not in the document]
marsella: Yeah, this feels a bit awkward; I was originally going to install all the solvers in the repo but they all have different commands to check if they're installed (can't just do `$solver --version`) and these are the only two needed to run the tests. I figured if users want other specific solvers they can double check the script and do the equivalent thing for themselves.

dev_setup.sh
marsella: Thanks for clarifying, I fixed to use this convention.

dev_setup.sh
marsella: Thanks for pointing this out! I think I got all of the bashisms out

dev_setup.sh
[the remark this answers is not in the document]
marsella: Fixed!

marsella:
Thank you for the feedback, I think I've addressed all of it, with some questions inline. This should now support macOS 12 as well as 14, but I am not sure how to test it since I don't have that OS and I don't think you can make a Docker container running macOS. 

Do you have a suggestion for how to test it, or is it OK to just add that it should support 12 but isn't tested in the docs?

dev_setup.sh
jtdaugherty: Looking at the latest changes, I see that `env.sh` is being *generated* by `dev_setup.sh`, so it sounds like above you're saying that is what's messy (appending to the generated file). I had assumed that `env.sh` would be static and put under version control, not generated.

dev_setup.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/sh
+
+# cryptol setup script
+#
+# Supported distribution(s):
+#  * macOS 14 (AArch64)
```
jtdaugherty, line 6: This mentions MacOS 14, but not 12, despite 12 being mentioned in comments elsewhere. Is that deliberate?

dev_setup.sh
```diff
@@ -0,0 +1,142 @@
...
+}
+
+# Installs the two solvers required to run the test suite for the repo.
+# Users may want to install other solvers, and indeed the what4 solvers repo
+# includes a half dozen other solvers that are compatible with cryptol. 
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
+
+        if [ ! is_macos ]; then
```
jtdaugherty, line 105: This OS check is done inside this function. Is there a reason it can't be done at the top level? I find it odd for it to be here because it halts the setup process, and that's the sort of check I tend to assume would be done before any real work begins, so that once work begins, assumptions like "we're on a supported platform" can be thought to hold for the rest of the script.

dev_setup.sh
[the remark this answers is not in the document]
jtdaugherty: Oh, interesting. So the solver installation here is only about getting the tests running? I ask because I could imagine that for getting an *installation* of cryptol running, a more robust solver installation process might be needed but this could be left as-is.

dev_setup.sh
jtdaugherty: I like the new changes; I think it can't hurt to factor those sorts of things out and it also results in other code being easier to refactor. Thanks!

jtdaugherty:
> Thank you for the feedback, I think I've addressed all of it, with some questions inline.

I resolved all the threads that I started that have been addressed in your latest changes. Thank you!

> This should now support macOS 12 as well as 14, but I am not sure how to test it since I don't have that OS and I don't think you can make a Docker container running macOS.
>
> Do you have a suggestion for how to test it, or is it OK to just add that it should support 12 but isn't tested in the docs?

My $0.02 is that until we have tested on macOS 12, we cannot say it is supported. Perhaps IT has a testing host with macOS 12 on it still? And then I must also ask: is it important to support macOS 12 specifically for some reason?

dev_setup.sh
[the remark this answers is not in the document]
jtdaugherty: One thought on development setup vs. installation is that I think that unifying how the solvers are installed and managed across both of those would be really important for ensuring that what we test on, what the user uses, and what we say we support, are all ensured to be overlapping. So that's more food for thought as other infrastructure gets touched here, since ISTM it would be really nice if there was a single source of truth for solver management that is easy to update, and then we can be confident that CI, development, and deployment environments are all using exactly the same things.

dev_setup.sh
sauclovian-g: @jtdaugherty we aren't in my world, was partly wandering off into the weeds and partly trying to sort out what we do need. It seems that the build system we have needs environment variables set, and changing that (if it's possible at all) is well beyond the scope of this ticket. But the general points about tree config remain.

ISTM that there are basically two choices:
1. check in a separate env.sh that probes the system on the fly and sets the environment variables (with two subcases, one where you run it with ```. ./env.sh``` and one where you run it with ```eval $(./env.sh)```)
2. have the setup script generate env.sh based on what it finds, akin to a configure script

of which (1) is possibly more robust but (2) is simpler and closer to what's already here. So in the interests of not dragging this out or making extra work, I think I'd suggest (2). I think the difference from what we already have is basically just writing out a new env.sh each time instead of appending...

sauclovian-g: however if @marsella is up for writing (1) I have no objection to that either...

I guess a third option might be to check in a separate env.sh for each environment but that has diminishing returns...

jtdaugherty: @sauclovian-g either option works for me and I think addresses the concern that I was trying to express, namely, just not to ask the user to do manually what could otherwise be automated to increase consistency.

dev_setup.sh
```diff
@@ -14,23 +14,25 @@ set -e
 
 HERE=$(cd `dirname $0`; pwd)
 LOG=$HERE/dev_setup.log
+ROOT=$HERE
+ENV=$ROOT/env.sh
```
sauclovian-g, line 18: Don't call this ```ENV```; that means something magic to the shell (it's a misfeature for doing something similar to .profile)

:-(

dev_setup.sh
sauclovian-g: Right

dev_setup.sh
```diff
@@ -0,0 +1,142 @@
...
+        # The output is not routed to log because the installer is interactive.
+        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
+    else
+        notice "Using existing ghcup installation"
+    fi
+
+}
+
+function is_macos {
+    [ "$OSTYPE" = "darwin"* ]
```
sauclovian-g, line 67: The * is a bashism. Also I don't think $OSTYPE is standard; it's not set in sh on a random Ubuntu VM I just checked. Probably what you want is to use ```$(uname -s)``` and test it for "Darwin" because that way you don't need to mask off a version number, plus uname is portable. (Or portable enough. It's missing on some things like AIX that we definitely don't care about.)

dev_setup.sh
sauclovian-g: I found one more, but I gave it its own comment so I'll mark this one done

sauclovian-g: Thanks!

dev_setup.sh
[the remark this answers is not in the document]
marsella: Yeah, I agree that it would be nice to push any interesting logic here into the what4_solvers repo to let it be called anywhere. I don't know how people actually use cryptol to say whether it makes sense to install one or more of them, which is why I defaulted to just getting the tests to run.

dev_setup.sh
sauclovian-g: I think the idea is to put something like
```
case $(uname -s) of
    Darwin) CRYPTOL_PLATFORM=macos;;
    Linux) if [ -f /etc/redhat_release ]; then CRYPTOL_PLATFORM=redhat else ...
    *) echo "Unsupported platform" 2>&1; exit 1;;
esac
```
at the top and then use the results of that probe in whatever other conditional logic is required later.

dev_setup.sh
marsella: Thank you for all this discussion! I took option 2.

> [...] the environment variables being mentioned are, it would seem, a requirement for working with the project

This particular instance is more about setting up Homebrew correctly, which is independent of this particular project and which others may have already done correctly or differently on their own machines. E.g. you can change the default prefix that brew uses, so I would not want to hard-code the environment variables in a file because it could bork someone's system that doesn't use the default Homebrew options, which aren't necessary to run this project. I think that's what @sauclovian-g's option 1 is getting at.

dev_setup.sh
marsella: I added a version check and a warning.

dev_setup.sh
marsella: I updated the docs throughout. The current state is
- Supported distribution is macOS 14
- I left in all the conditionals and checks for macOS 12
- I added a top-level check [per your next comment](https://github.com/GaloisInc/cryptol/pull/1674#discussion_r1622691534) to reject anyone running macOS 12

Hopefully the next dev that wanders in with macOS 12 can uncomment the top-level check, confirm that the rest of the script works, and not have to do any additional work.

dev_setup.sh
marsella: This is much nicer, thanks.

dev_setup.sh
marsella: oooops

dev_setup.sh
marsella: Oh, thanks. I used `uname` instead. I think there are some failure modes where e.g. someone is running macos 10 which has the same category and hardware as macos 12 but I'm (a) not sure how to detect exactly the right thing we need and (b) don't expect anyone is going to try to run this on older macOS versions or, like, an iPad.

marsella:
This is ready for re-review.

> on the scripts/ front, I think we probably do want a subdir to hold developer-facing utility material... but probably most of the stuff we'll be putting in it probably isn't scripts. So I dunno what to call it and probably it should wait until we have more stuff...

Sounds good to me, I'll leave this floating where it is.

> My $0.02 is that until we have tested on macOS 12, we cannot say it is supported. Perhaps IT has a testing host with macOS 12 on it still? And then I must also ask: is it important to support macOS 12 specifically for some reason?

We still have not formally decided what the set of officially supported OSes is for cryptol, but it will probably include macOS 12. I have not surveyed the folks at Galois who are currently working on or potentially ramping up on cryptol to know if any of them are actively using macOS 12 such that it would be worth testing and officially adding support at this time.

dev_setup.sh
```diff
@@ -9,19 +9,46 @@
...
 
 WHAT4_SOLVERS_SNAPSHOT="snapshot-20240212"
 WHAT4_SOLVERS_URL="https://github.com/GaloisInc/what4-solvers/releases/download/$WHAT4_SOLVERS_SNAPSHOT/"
 WHAT4_SOLVERS_MACOS_12="macos-12-X64-bin.zip"
 WHAT4_SOLVERS_MACOS_14="macos-14-ARM64-bin.zip"
 
+MACOS14="macos14"
+MACOS12="macos12"
+# Make sure we're running on a supported platform
+case $(uname -s) in
```
jtdaugherty, line 35: I recommend putting this whole `case` into a function, `supported_platform`, whose output is the platform string. Then reduce duplication of the echo/exit pairs so that you end up with something like
```bash
CRYPTOL_PLATFORM=$(supported_platform)

if [ -z "$CRYPTOL_PLATFORM" ]
then
    echo "Unsupported platform"
    exit 1
fi
```

jtdaugherty:
@marsella I added a couple more comments and resolved a bunch of others. Aside from the new ones I believe all of my older threads are now resolved. Thank you!

dev_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+    then
+        # This may or may not be the same directory that this script lives in.
+        mkdir -p `dirname $LOG`
+        echo "$@" >>$LOG
+        if ! "$@" >>$LOG 2>&1
+        then
+            echo
+            echo "An error occurred; please see $LOG"
+            echo "Here are the last 50 lines:"
+            tail -n 50 $LOG
```
kquick, line 68: I'd recommend shortening this tail and/or putting something indicating there was an error closer to the bottom of the output.  The concern here is that the user will just see the tail of the log (which may or may not contain error output) and not see the indication of an error because it is much further up (and possibly scrolled off the visible screen area).

dev_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+}
+
+function install_ghcup {
+    if ! ghcup --version &> /dev/null
+    then
+        notice "Installing ghcup, GHC, and cabal"
+        # Technically the installation only requires cabal, but it's
+        # recommended to get the whole GCH shebang in one package.
+        # The output is not routed to log because the installer is interactive.
+        curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
```
kquick, line 87: Piping a network download through `sh` makes me cringe, especially because it could happen as a hidden part of this script without user consent.  

Is this installer a short thing that doesn't change much?  This would allow having a (short list of) checksum values here in this script that verify the download before executing.  Or maybe just pre-informing the user of what is about to happen and getting their consent.

dev_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+        notice "Did not install GMP. This script only supports macOS 14"
+    fi
+}
+
+# Installs the two solvers required to run the test suite for the repo.
+# Users may want to install other solvers, and indeed the what4 solvers repo
+# includes a half dozen other solvers that are compatible with cryptol. 
+function install_what4_solvers {
+    if ! cvc4 --version &> /dev/null || ! cvc5 --version &> /dev/null; then
+        notice "Installing cvc4 and/or cvc5 solvers"
```
kquick, line 114: I'd recommend indicating where these are coming from (WHAT4_SOLVERS_URL) so the user is informed and knows these might be different locations.

dev_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+
+        solvers_dir=$(mktemp -d)
+        logged curl --proto '=https' --tlsv1.2 -sSfL -o "$solvers_dir/solvers.bin.zip" "$WHAT4_SOLVERS_URL$solvers_version"
+        cd $solvers_dir
+        logged unzip solvers.bin.zip
+        
+        # If we want to install more solvers by default, we can do so here,
+        # although we might need a different check than `--version`
+        for solver in cvc4 cvc5; do
+            if ! $solver --version &> /dev/null; then
```
kquick, line 130: Should this be `if ! ./$solver ...` since frequently `.` is explicitly not in PATH for security reasons?  This also helps ensure this operates on the just unpacked file and not something that might be in the PATH (which was verified in line 113, but is still a "bad smell").  Also, the notice in the following line should probably indicate where the solvers are being installed so the user can ensure they are in their path.  And I guess perhaps this whole `then` block might want to check if `/usr/local/bin` is in the path in case these are actually installed and it's just a path issue.

dev_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+                sudo mv $solver /usr/local/bin
+            fi
+        done
+
+        rm -r $solvers_dir
+    else
+        notice "Not installing cvc4 or cvc5 solvers because they already exist"
+
+        if ! (grep -q "version 1.8" <<< "$(cvc4 --version)"); then
+            notice "Your version of cvc4 is incorrect; expected 1.8"
```
kquick, line 142: I'd suggest "unexpected" rather than "incorrect" and adding the path where the corresponding solver was found.  

You might also want to let the user know that the setup is not marked as failure due to these version mismatches but that they might wish to re-run after uninstalling the existing versions  to get the suggested versions installed to avoid runtime cryptol issues.

And lastly, you might want to use vars (e.g. WHAT4_CVC5_VERSION) here that you set near line 30 so that it's close to the declared snapshot version used for future updates.

dev_setup.sh
sauclovian-g: ```uname -r``` prints the OS version (unfortunately on macOS it's not the MacOS version but the internal version of the kernel, or something, anyway it's 22.* for Ventura) and ```uname -m``` and ```uname -p``` print different variations of the machine type. Those are all portable (except e.g. AIX); there's also probably an apple-specific way to get "13" for Ventura or even the string "Ventura" but I don't know offhand what it is.

anyway info is available -- I suspect we don't actually need to care about the OS version though, just whether it's x86 or arm.

sauclovian-g: (I guess we do need the OS version as well as the platform to get the right solver bins)

dev_setup.sh
```diff
@@ -63,35 +91,18 @@ function install_ghcup {
...
 function install_gmp {
-    if [ is_macos ]; then
+    if [ $CRYPTOL_PLATFORM = $MACOS12 ] || [ $CRYPTOL_PLATFORM = $MACOS14 ]; then
         notice "Installing GMP via Homebrew, if it's not already installed"
         logged brew install gmp
 
         # `brew --prefix` is different on macOS 12 and macOS 14
-        echo "export CPATH=$(brew --prefix)/include" >> $ENV
-        echo "export LIBRARY_PATH=$(brew --prefix)/lib" >> $ENV
+        echo "export CPATH=$(brew --prefix)/include" >> $VAR_FILE
```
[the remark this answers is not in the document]
marsella: Oh yeah, in the interest of future-proofing against other variables getting added, I put a `truncate` call above so the file starts out empty when the script is run, and then this appends.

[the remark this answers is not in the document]
jtdaugherty: Okay, great.

dev_setup.sh
marsella: I looked but it seems that ghcup doesn't publish checksums; I could verify the gpg signatures but that would require adding some gpg keys to the user's toolchain and running the downloaded program anyway and adds a dependency on gpg, which doesn't seem ideal to me. I will add a prompt to get consent though.

jtdaugherty: Just a note: this script almost certainly needs to be runnable in an non-interactive setting when being used to set up the environment in a container. Consent would be interactive, but it might also be important to add a command-line flag like `--non-interactive` to give consent *and* install `ghcup` with defaults as a result.

dev_setup.sh
jtdaugherty: FWIW in scripts where I've used this (since this is based on my script), there is no error to see that isn't in the log since the console output of the script is exclusively non-error messages. Of course, whether 50 lines of log will be enough to show an error depends a lot on what you're doing, though. That's why (in my original script) the message tell you where to find the log in case the last 50 lines aren't useful enough to understand what happened.

dev_setup.sh
jtdaugherty: Also, isn't a user not consenting to install `ghcup` equivalent to not consenting to run this? Isn't the content of this script implicitly trusted? If there's a concern about running malicious code, you could also pre-fetch the script and put it under version control, and run that instead.

kquick: By "run this" I assume you mean "run dev_setup"; running a script I've downloaded and have read is different than something automatically downloading and running something else entirely that I don't necessarily even have visibility into or awareness that it's going to occur.  Even with https and certs, I know that *I* personally would want the opportunity to inspect something downloaded before just letting it run whatever commands it wanted.  The original intent is clearly benign and helpful, but we recently saw the exploit where innocuous looking changes accepted to a project had rather sinister effects.  Not that I'm claiming I would have caught those either, but I'm just suggesting we err on the side of safety and clarity.

Also, if we want this to run non-interactively that  presumably would require some flags or other input to the ghcup installation process as well; I haven't used ghcup so I'm based this on the preceding comment in the devscript.

jtdaugherty: > By "run this" I assume you mean "run dev_setup"

Yes, that's what I was thinking.

kquick: Good.

Also, I apologize for jumping in on this and don't mean to derail the process.  I've just had lots of experiences (both good and bad) in this area and wanted to help steer us towards the former.  So please feel free to ignore me if I'm just adding friction, and thank you again for putting this together, @marsella .

dev_setup.sh
marsella: Oh, this is confusing. This explicitly is trying to run the version on the path, _not_ the version in the local directory. The parent conditional says if we're missing either of these solvers, we need to download the whole set of binaries from what4_solvers. But if you're only missing one of them, we don't want to clobber the version you do have, so this double-checks for each solver whether it's already in the path and only installs it if it not.

I'm adding an `is_installed` method to replace this version check that will hopefully make it more obvious what's happening.

dev_setup.sh
marsella: These are good suggestions, I cleaned all this up.

dev_setup.sh
marsella: This is now superseded by the `supported_platform` method which I think doesn't use bashisms or un-portable methods!

dev_setup.sh
marsella: I popped an `[ERROR]` underneath the log anyway so that users who don't routinely read any terminal output beyond 3 lines away will know that something went wrong.

dev_setup.sh
marsella: > it might also be important to add a command-line flag

This makes sense; I would propose adding any CI-specific flags when we take on the issue to actually use this in CI rather than right now.

> I apologize for jumping in on this

It's no problem! Usually I'm the only one who uses my scripts, so it's nice to get a better sense of best practices.

dev_setup.sh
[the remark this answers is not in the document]
sauclovian-g: (this got done)

dev_setup.sh
sauclovian-g: I'll go look at the new stuff in a bit :-)

dev_setup.sh
sauclovian-g: One should really never curl | sh, but that's how you install a lot of things these days and pushing back on it isn't very effective IME :-(

I think you can install ghcup from homebrew instead (that's where mine seems to have come from) and while I suspect it just does the same thing there's at least other eyeballs on it. So maybe we should take that route since we're using homebrew for other stuff?

jtdaugherty: > This makes sense; I would propose adding any CI-specific flags when we take on the issue to actually use this in CI rather than right now.

Sounds good to me, something to keep in mind for later if it isn't relevant to your immediate needs. Maybe add a comment to the top of the script saying that it's assumed to be interactive in the mean time?

jtdaugherty:
I just went through and resolved any remaining threads of mine (all of which were addressed). Thanks again!

kquick (approved):

jtdaugherty (approved):

dev_setup.sh
```diff
@@ -0,0 +1,189 @@
...
+    esac
+}
+
+RED="\033[0;31m"
+function notice {
+    echo "[NOTICE] $*"
+}
+
+function is_installed {
+    command -v $* >/dev/null 2>&1
```
sauclovian-g, line 64: This probably wants "$@"

dev_setup.sh
```diff
@@ -0,0 +1,189 @@
...
+        ghcup_url="https://get-ghcup.haskell.org"
+        notice "ghcup not found; do you want to install from $ghcup_url"
+        read -p "Press Enter to continue or 'n' to skip: " consent
+        case $consent in
+            [Nn]* ) notice "Skipping ghcup installation"; return;;
+            * ) notice "Installing ghcup";;
+        esac
+
+        # Technically the installation only requires cabal, but it's
+        # recommended to get the whole GCH shebang in one package.
```
sauclovian-g, line 104: nitpick: GCH -> GHC

dev_setup.sh
```diff
@@ -0,0 +1,189 @@
...
+        # The output is not routed to log because the installer is interactive.
+        ghc_dir=$(mktemp -d)
+        logged curl --proto '=https' --tlsv1.2 -sSf -o "$ghc_dir/ghcup.bin" $ghcup_url
+        cd $ghc_dir
+        sh ghcup.bin
```
sauclovian-g, line 109: I don't think anything ever cds back. Might want to run it in a subshell so that the change is discarded when done. (That is, ```(cd $ghc_dir && sh ghcup.bin)```)

sauclovian-g (approved):
Couple more minor things I noticed, nothing else new

dev_setup.sh
marsella: The brew package only installs ghcup and not the whole toolchain but I was able to manually install the other parts. It does look much nicer now. I left a breadcrumb to the original install method in the docs in case the simple thing doesn't work for someone. As an added bonus, this removes all the interactivity!

dev_setup.sh
marsella: This is now gone but I had done the same thing in the what4 solvers bit. That's now fixed.

marsella:
Ready once again for your eyes 👀

dev_setup.sh
```diff
@@ -0,0 +1,210 @@
...
+            logged brew install ghcup
+            USED_BREW=true
+        else
+            notice "Did not install GHCup. This script only supports macOS 14"
+        fi
+
+        logged ghcup install ghc --set recommended
+        logged ghcup install cabal --set recommended
+        logged ghcup install stack --set recommended
+        logged ghcup install hls --set recommended
```
glguy, line 109: I think we should avoid referencing ghcup's *recommended* versions here. The recommended version will change over time and won't necessarily be in sync with the version we intended to target. We could list at the top of this script the versions of tools we require and then install exactly those versions with these commands. When it comes up invoking the versions of things we want it can be done explicitly by version.

For example when building with cabal we can do: `cabal configure -w ghc-9.6.5` to explicitly set the build to use the expected version of GHC. If we want to lock down the version of cabal we use we could also go further: `cabal-3.10.3.0 configure -w ghc-9.6.5`

glguy: As far as I know we don't use stack in our builds. We probably don't need to install it as part of this development environment.

jtdaugherty: +1 on locking down GHC and related tools here.

sauclovian-g: There are cabal configs for three GHC versions, though; which one? Or should we install all three by default?

jtdaugherty: Yeah... good question. ISTM having this script use those cabal configs (or whatever canonically defines which GHCs we need to support) to find out which GHC versions to install might be nice, in the spirit of having one source of truth for that.

PR #1682, opened 2024-06-06
Initial implementation of :check-docstrings

glguy:


src/Cryptol/REPL/Command.hs
```diff
@@ -177,21 +178,29 @@ instance Ord CommandDescr where
...
+  | ModNameArg  (String   -> REPL CommandResult)
+  | FilenameArg (FilePath -> REPL CommandResult)
+  | OptionArg   (String   -> REPL CommandResult)
+  | ShellArg    (String   -> REPL CommandResult)
+  | HelpArg     (String   -> REPL CommandResult)
+  | NoArg       (REPL CommandResult)
+
+data CommandResult = CommandResult
+  { crType :: Maybe String -- ^ type output for relevant commands
+  , crValue :: Maybe String -- ^ value output for relevan commands
```
sauclovian-g, line 194: typo :-)

RyanGlScott:
Could you write some documentation that describes how `:check-docstrings` works at a high level? The [Cryptol reference manual](https://github.com/GaloisInc/cryptol/tree/master/docs/RefMan) might be one place to put this documentation, but if there isn't a good place in the manual to put it, even including the comments as just Haddocks would be nice. In particular, I think we should make note of things such as:

* What syntax `:check-docstrings` is looking for?
* Can docstrings live in any type of comment, or can they only belong to certain types of code blocks?
* What are the conditions under which this command will succeed or fail?
* Are the docstring blocks run in a particular order? If not, the documentation should say that this is unspecified so that we are free to change the order.

Some test cases that exercise the functionality of `:check-docstrings` would also be appreciated.

sauclovian-g:
If there isn't a place in the manual to put it, feel free to stick it in an appendix marked something like "Additional notes to be merged", create another issue for sorting it out properly, and assign that to me. There should be a place for everything in the manual.

yav (approved):
Looks good to me.

PR #1684, opened 2024-06-11
Update dev-setup script to support Ubuntu 20.04 and 22.04

marsella:
This adds Ubuntu support to the dev-setup script. Makes progress toward, but does not complete, #1673.
- Adds some very minimalist Dockerfiles to use for testing
- Adds ubuntu support throughout the script
- Moves script + docker stuff into a `dev/` directory for containment

Due to the last item, the diff on here is unhelpful. Consider reviewing [without the first commit](https://github.com/GaloisInc/cryptol/pull/1684/files/8df304dc768d1f5eec6a294781127be5cfadbdd6..1952c8b4ff6198fdd46731eec6b21992dd67ee68), which will better show you the changes in the script.

Some questions for reviewers:
- In order to test the Ubuntu versions correctly from Apple hardware, you sometimes need to specify the platform to use when building and running the docker container (I had to do it the first time I built the container, but it seemed like subsequent builds remembered). Where should this documentation live?
```
$ docker build -t cryptol-ubuntu-22 --platform linux/amd64 -f ubuntu22.04.Dockerfile
```
- To keep things simple, I'm using the z3 version packaged in what4-solvers. The readme for that repo suggests using the latest version of z3 (4.13.0) but provides a much older version (4.8.something). Should I download the latest version via other means, instead?


dev/dev_setup.sh
```diff
@@ -88,29 +117,47 @@ function logged {
...
+        case $CRYPTOL_PLATFORM in
+            $MACOS12 | $MACOS14)
+                notice "Installing GHCup via Homebrew"
+                logged brew install ghcup && USED_BREW=true;;
+            $UBUNTU20 | $UBUNTU22)
+                notice "Installing any missing GHCup dependencies via apt-get"
+                logged apt-get install -y build-essential curl libffi-dev libffi7 libgmp-dev libgmp10 libncurses-dev libncurses5 libtinfo5
+
+                notice "Installing GHCup via direct download"
+                logged curl --proto '=https' --tlsv1.2 -sSfL -o ghcup "$GHCUP_URL/$(uname -m)-linux-ghcup"
```
marsella, line 147: note: I couldn't find a version of GHCup for linux that came via a normal package manager, hence the manual install of dependencies + downloading and running a binary straight from the internet 😬

dev/dev_setup.sh
```diff
@@ -120,97 +167,140 @@ function install_ghcup {
...
-    if ! is_installed z3; then
-        notice "Installing z3 solver via Homebrew"
-        logged brew install z3
-        USED_BREW=true
-    fi
+install_zlib() {
+    case $CRYPTOL_PLATFORM in
+        $MACOS12 | $MACOS14)
+            notice "Installing zlib via Homebrew, if it's not already installed"
+            logged brew install zlib && USED_BREW=true;;
```
marsella, line 186: note: I'm pretty sure this comes by default in all macs these days, but I added it here for completeness, on the off chance that somebody accidentally trashed their own version. Brew defaults to not linking it to anything to avoid borking with the system version.

dev/ubuntu20.04.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:20.04
+WORKDIR /home/
+
+RUN ["apt-get", "-y", "upgrade"]
+RUN ["apt-get", "-y", "update"]
+RUN ["apt-get", "install", "-y", "git"]
+RUN ["git", "clone", "https://github.com/GaloisInc/cryptol.git"]
+
+COPY dev_setup.sh ./cryptol/
```
marsella, line 9: todo: oops, this is a testing artifact. I will update this to run the setup script in the repo instead. Ditto below.

mccleeary-galois:
> This adds Ubuntu support to the dev-setup script. Makes progress toward, but does not complete, #1673.
> 
> * Adds some very minimalist Dockerfiles to use for testing
> * Adds ubuntu support throughout the script
> * Moves script + docker stuff into a `dev/` directory for containment
> 
> Due to the last item, the diff on here is unhelpful. Consider reviewing [without the first commit](https://github.com/GaloisInc/cryptol/pull/1684/files/8df304dc768d1f5eec6a294781127be5cfadbdd6..1952c8b4ff6198fdd46731eec6b21992dd67ee68), which will better show you the changes in the script.
> 
> Some questions for reviewers:
> 
> * In order to test the Ubuntu versions correctly from Apple hardware, you sometimes need to specify the platform to use when building and running the docker container (I had to do it the first time I built the container, but it seemed like subsequent builds remembered). Where should this documentation live?
> 
> ```
> $ docker build -t cryptol-ubuntu-22 --platform linux/amd64 -f ubuntu22.04.Dockerfile
> ```
> 
> * To keep things simple, I'm using the z3 version packaged in what4-solvers. The readme for that repo suggests using the latest version of z3 (4.13.0) but provides a much older version (4.8.something). Should I download the latest version via other means, instead?

We will want this type of documentation in the future README refactor in the dev setup part.

dev/dev_setup.sh
mccleeary-galois: This has been sadly the standard that we have done for most of our docker container/scripts thus far. I also couldn't find a better source.

RyanGlScott: To be fair, `ghcup` explicitly bills itself as an alternative to your system package manager that is specific to GHC. Both `ghcup` and your system package manager will, in some sense, download a binary straight from the Internet—it's just a matter of whether you trust the tool that's providing it.

dev/dev_setup.sh
```diff
@@ -0,0 +1,306 @@
...
+    exit 1
+fi
+
+update_submodules
+install_ghcup
+install_gmp
+install_zlib
+install_solvers
+put_brew_in_path
+set_ubuntu_language_encoding
```
kquick, line 303: I appreciate the way this structure summarizes the set of operations being performed.  It is a good "self-documenting" summary of what the code is doing for both the user of this script or someone working in an environment this script doesn't support and who wants to know what it does.  To support both of these, I'd recommend adding a reference to this portion in the preface comments; for example, add a "# Main operations" comment just above line 297 and then finish the paragraph in lines 10-13 with "For a list of the setup operations performed by this script, see the "Main Operations" section below (near the end of this file)."

dev/dev_setup.sh
```diff
@@ -0,0 +1,306 @@
...
+    echo "Unsupported platform; this script supports Ubuntu 20.04, Ubuntu 22.04, and macOS 14"
+    exit 1
+fi
+
+update_submodules
+install_ghcup
+install_gmp
+install_zlib
+install_solvers
+put_brew_in_path
```
kquick, line 302: At first I was confused, thinking "Shouldn't this be the first operation?  Most of the subsequent operations use `brew` (when on a Mac) so it seems like the brew path setup should precede them." but then I looked at the details and see this adds brew to the user's path in the env setup script.

I'm curious as to why that is done: if this script is responsible for setting up the user's dev environment by installing all the needed things, why would the user need access to `brew`?

Also, why isn't this setup needed for the `PATH` while the script operates?  It seems like the script is already assuming `brew` is in the `PATH`?

dev/dev_setup.sh
kquick, line 303: This should probably be moved to the Dockerfile and we shouldn't be overriding the local user's locale settings.  

Maybe it's worth warning the user if `LANG` or `LC_x` doesn't (case-insensitively?) contain "UTF-8" for their local environment, but other than that I think we shouldn't mess with the user's environment like this and just move it to the Dockerfile where we know it needs to be set to support UTF-8.

Also, the conventional wisdom is that you shouldn't set `LC_ALL`: it's intended as something the user can manually set as an override for all of the other `LC_x` settings.  Maybe we decide we're OK with that for the constrained Docker environment, but there's probably a more direct `LC_x` setting (or just set all except `LC_ALL`) that is still effective.

dev/dev_setup.sh
marsella: Oh, this is just confusingly named. This puts the install directories that brew uses by default into the path. We do assume that you already have brew installed / accessible when you call the script (I added a note re: this to the top of the file). This makes sure that, if the script had to install any packages via brew, they'll be available in your path.

Maybe `add_brew_prefix_to_path`? or `put_brew_packages_in_path`.

dev/dev_setup.sh
mccleeary-galois: I agree we should be careful with this, however another use case for this is to quickly set up VMs. Perhaps a warning + option to set language encoding? We can start handling optional flags for some things like this as well, but ideally we would have a 1 line of code (with possible flags) for setting up a VM as well.

dev/dev_setup.sh
```diff
@@ -66,17 +70,14 @@ supported_platform() {
         Linux)
```
sauclovian-g, line 70: You might want to distinguish between Debian-based and redhat-based distros here, since apt-get is Debian-only...

dev/ubuntu22.04.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:22.04
+WORKDIR /home/
+
+RUN ["apt-get", "-y", "upgrade"]
+RUN ["apt-get", "-y", "update"]
```
sauclovian-g, line 5: you might want to stuff build-essential in here (although I suppose the image probably already has it)

sauclovian-g (approved):
generally LGTM

dev/dev_setup.sh
kquick: Renaming would probably help.  Whatever you prefer, although the latter proposed name sounds the most descriptive to me.

dev/ubuntu22.04.Dockerfile
marsella: It does not come by default, I popped it into the setup script because it's on the list of 'things you need to download to get ghcup working'. I can also put it here, with a wary eye toward the possibility that we could iterate for a long time about what a reasonable base Dockerfile looks like and I don't want to fall down that rabbit hole right now 🐰

dev/dev_setup.sh
marsella: Thanks, I did not know this wisdom! I played around a bit and it seems like just setting `LANG` is enough to get the tests to pass. I'll modify this section to check and print a warning, and to add the `export LANG...` line as a comment in the generated var file. That way someone knows there's a problem, can easily uncomment the line if they don't care about changing their settings, but the default behavior lets them decide how to solve the problem.

I'll also update the dockerfile to set these.

dev/ubuntu20.04.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:20.04
+WORKDIR /home/
+
+RUN ["apt-get", "-y", "upgrade"]
+RUN ["apt-get", "-y", "update"]
+RUN ["apt-get", "install", "-y", "git"]
```
weaversa, line 6: Each `RUN` statement creates a layer in the Docker image. This can artificially increase the size of images. Official Docker best practice can be found here: https://docs.docker.com/build/building/best-practices/#run

weaversa: After re-reading the best practice, I see there are also caching issues that make chaining `apt-get` `RUN` commands necessary. Wow.

marsella: Thanks for flagging this, I've seen the pattern of chaining apt-get commands before but didn't realize why.

For what it's worth, I'm not going to do a full apt-get clean here because we do install additional packages in the setup script; if maximally optimizing for size of the image is important, we could install all those packages here and then delete the rest of the apt-get cache, but it seems like [it doesn't save a ton of space](https://gist.github.com/marvell/7c812736565928e602c4?permalink_comment_id=3463716#gistcomment-3463716) so I'm going to save that for later.

dev/ubuntu20.04.Dockerfile
marsella: Fixed this. It technically won't work for anyone testing before this is merged because the dev setup script isn't in main; you can test locally by adding `"-b", "add-ubuntu-dev-setup"` to the clone line.

dev/dev_setup.sh
marsella: This is nice, I added it.

dev/dev_setup.sh
marsella: ooh, nice. Fixed.

marsella:
Thanks, all! I think I've addressed all feedback. I'm iterating on making sure that the sourcing step of the dockerfiles works correctly, for some reason it's not as straightforward as I expected. But other comments are hopefully now resolve-able.

> We will want this type of documentation in the future README refactor in the dev setup part.

Cool. I added a note to the Dockerfiles for now so it doesn't get lost / forgotten.

dev/dev_setup.sh
```diff
@@ -0,0 +1,321 @@
...
+# Set of supported platforms:
+MACOS14="macos14"
+MACOS12="macos12" # actually, this isn't supported yet
+UBUNTU20="ubuntu-20.04"
+UBUNTU22="ubuntu-22.04"
+
+USED_BREW=false
+
+# Returns a string indicating the platform (from the set above), or empty if
+# a supported platform is not detected.
```
kquick, line 59: I would suggest returning a non-blank value here (e.g. maybe "unknown-platform" or something like that).  The reason is that blank results can trip up shell scripts unless special care is taken to guard against their presence (e.g. `case $blank_value in ...`  becomes `case in ...` and `if [ ubuntu -eq $BLANK ] ...` becomes `if [ ubuntu -eq ] ...`).  I don't know that to be the case herein, it's just a preventative measure.

marsella: When we call this function we check right away if it's blank and `exit` if so, so it shouldn't cause issues anywhere else.

dev/ubuntu22.04.Dockerfile
sauclovian-g: Probably better off in the setup script. Anyone who doesn't have it is going to want it.

dev/dev_setup.sh
```diff
@@ -279,13 +287,20 @@ put_brew_in_path() {
...
-            notice "Adding language environment variables to $VAR_FILE"
-            echo "export LANG=C.UTF-8" >> $VAR_FILE
-            echo "export LC_ALL=C.UTF-8" >> $VAR_FILE;;
-        *) ;;
+            if $LANG!= *"UTF-8"* || $LANG != *"utf-8"*; then
+                notice "Language environment variables are not set as expected."
+                notice "    You may need to set them to UTF-8."
+                notice "    Uncomment the LANG var in $VAR_FILE before sourcing to do so"
+                echo "# uncomment the following lines to fix decoding errors e.g. hGetContents" >> $VAR_FILE
+                echo "# export LANG=C.UTF-8" >> $VAR_FILE
```
sauclovian-g, line 295: I remember some time back I filed a bug report with I think Debian that there was no C.UTF-8 locale and they rejected it. That was years ago and they have probably grown up since, but maybe not.  (No need to do anything different here for now, but we might end up needing to revisit it sometime later. JFYI)

dev/dev_setup.sh
```diff
@@ -279,13 +287,20 @@ put_brew_in_path() {
 set_ubuntu_language_encoding() {
     case $CRYPTOL_PLATFORM in
         $UBUNTU20 | $UBUNTU22)
-            notice "Adding language environment variables to $VAR_FILE"
-            echo "export LANG=C.UTF-8" >> $VAR_FILE
-            echo "export LC_ALL=C.UTF-8" >> $VAR_FILE;;
-        *) ;;
+            if $LANG!= *"UTF-8"* || $LANG != *"utf-8"*; then
```
sauclovian-g: I don't think this conditional is going to work. Try
```
case x"$LANG" in
    *UTF-8*|*utf-8*) ;;
    *) notice ... ;;
```
(the x makes it not croak if the variable isn't set)

dev/ubuntu20.04.Dockerfile
```diff
@@ -1,10 +1,22 @@
...
+        build-essential \
+        git
 RUN ["git", "clone", "https://github.com/GaloisInc/cryptol.git"]
 
-COPY dev_setup.sh ./cryptol/
-CMD ["/bin/bash"]
\ No newline at end of file
+WORKDIR /home/cryptol
+RUN ["dev/dev_setup.sh"]
+RUN ["source", "dev/env.sh"]
```
sauclovian-g: This will, unfortunately, probably run a shell, load the environment settings into it, then exit and throw them away. I suspect we need to create a .profile or similar in the container that sources env.sh...

sauclovian-g: Oops, thanks gh. Maybe you can tell me that I'm making an outdated comment _before_ posting it? :-(

N/M.

marsella: Yeah. Actually it couldn't even run the source line, I kept getting errors like `"source": executable file not found in $PATH` and `source /home/cryptol/dev/env.sh: No such file or directory` despite trying to manually create it in the line before. I tried sourcing it with `. dev/env.sh` because I guess the default shell is dash and doesn't have source, but that didn't work, and I tried switching to a bash shell and then running the command, and running the command with normal `RUN source dev/env.sh` notation instead of the list like here, but it absolutely refused to ever find all the parts it needed. Hence the giving up and setting env vars by hand.

mccleeary-galois (approved):
LGTM we can always iterate in the future if needed.

PR #1701, opened 2024-07-10
Issue#1675

### LINKED ISSUE #1675: PrimeEC::ec_mult can produce incorrect results
```
Cryptol> :m PrimeEC
Loading module Cryptol
Loading module PrimeEC
PrimeEC> ec_mult`{13} 3 {x = 0, y = 0, z = 1}
{x = 0, y = 13, z = 1}
PrimeEC> :t ec_mult`{13}
ec_mult`{13} : Z 13 -> ProjectivePoint 13 -> ProjectivePoint 13
PrimeEC> ec_mult`{13} 3 {x = 0, y = 0, z = 1} : ProjectivePoint 13
{x = 0, y = 13, z = 1}
PrimeEC> {x = 0, y = 13, z = 1} : ProjectivePoint 13
[error] at <interactive>:334:13--334:15:
  • Unsolvable constraint:
      13 >= 14
        arising from
        at <interactive>:334:13--334:15
```

This is very strange behavior.

mccleeary-galois:
Closes #1675 

This seems to be the root cause bug, but I think there are some other improvements @marsella had in mind as well that we should go ahead and get in.

### DIFF
--- a/src/Cryptol/PrimeEC.hs
+++ b/src/Cryptol/PrimeEC.hs
@@ -294,7 +294,7 @@ ec_mult p d s
   | BN.bigNatIsZero (pz s) = zro
   | otherwise =
       case m of
-        0# -> panic "ec_mult" ["modulus too large", show (BN.bigNatToInteger (primeMod p))]
+        0# -> panic "ec_mult" ["integer with 0 width", show h]
         _  -> go m zro
 
  where
--- a/src/Cryptol/PrimeEC.hs
+++ b/src/Cryptol/PrimeEC.hs
@@ -202,13 +202,13 @@ ec_add p s t
 ec_sub :: PrimeModulus -> ProjectivePoint -> ProjectivePoint -> ProjectivePoint
 ec_sub p s t = ec_add p s u
   where u = case BN.bigNatSub (primeMod p) (py t) of
-              (# | y' #)    -> t{ py = BN.modBigNat y' (primeMod p) }
+              (# | y' #)    -> t{ py = y' `BN.bigNatRem` (primeMod p) }
               (# (# #) | #) -> panic "ec_sub" ["cooridnate not in reduced form!", show (BN.bigNatToInteger (py t))]
 {-# INLINE ec_sub #-}
 
 
 ec_negate :: PrimeModulus -> ProjectivePoint -> ProjectivePoint
-ec_negate p s = s{ py = BN.bigNatSubUnsafe (primeMod p) (py s) }
+ec_negate p s = s{ py = (BN.bigNatSubUnsafe (primeMod p) (py s))  `BN.bigNatRem` (primeMod p) }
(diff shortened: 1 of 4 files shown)


src/Cryptol/PrimeEC.hs
```diff
@@ -202,7 +202,7 @@ ec_add p s t
 ec_sub :: PrimeModulus -> ProjectivePoint -> ProjectivePoint -> ProjectivePoint
 ec_sub p s t = ec_add p s u
   where u = case BN.bigNatSub (primeMod p) (py t) of
-              (# | y' #)    -> t{ py = y' }
+              (# | y' #)    -> t{ py = BN.modBigNat y' (primeMod p) }
```
RyanGlScott, line 205: Would it suffice to call `` y' `bigNatRem` primeMod p `` instead? That way, you wouldn't need to introduce an additional `modBigNat` function.

RyanGlScott (commented):
The fix here looks quite plausible, although I am admittedly not an expert on elliptic curves. I do have one question about the implementation.

marsella (approved):
This will fix the reported bug, thank you!

I spent some time looking around in [the NIST document with EC algorithms](https://github.com/GaloisInc/cryptol/issues/1675#issuecomment-2147913578) you shared. In that doc, the subtraction algorithm matches what was here previously. I think the missing link is that the doc notes elsewhere that the three coordinates are all field elements, so the mod is implicit.

If we have users who depend on the elliptic curve operations in here, I would recommend making a separate issue to review the full file and bring it closer in line with a reference implementation. The two things I'd focus on are:
1. Making sure there are no other places where `BigNum`s are being used as normal numbers instead of the field elements they are supposed to represent.
2. Reviewing the names and conditions in this file compared to the reference. I noticed a few checks that are either missing or are in different places than I expected, so it might be worth confirming that this is fully correct.

tests/issues/issue1675/issue1675.icry
```diff
@@ -0,0 +1,5 @@
+:m PrimeEC
+ec_mult`{13} 3 {x = 0, y = 0, z = 1}
+:t ec_mult`{13}
+ec_mult`{13} 3 {x = 0, y = 0, z = 1} : ProjectivePoint 13
+{x = 0, y = 13, z = 1} : ProjectivePoint 13
```
yav, line 5: We probably don't need line (5), this is just type-checking stuff.

yav (approved):
Not an export, but the changes seem reasonable to me.

tests/issues/issue1675/issue1675.icry
mccleeary-galois: Removed in [dd11250](https://github.com/GaloisInc/cryptol/pull/1701/commits/dd112501ca1d3f08350fedf481267d3631832bf4)

PR #1706, opened 2024-07-22
Improve :check-docstrings

glguy:
* Handle more prefix-stripping cases for docstrings
* Catch REPL exceptions when checking docstrings
* Process unlabeled docstring codeblocks in :check-docstrings

### DIFF
--- a/docs/RefMan/BasicSyntax.rst
+++ b/docs/RefMan/BasicSyntax.rst
@@ -122,6 +122,7 @@ end-of-line can be dropped without affecting prefix detection.
 
 .. code-block:: cryptol
   :caption: Examples of docstrings
+
   /** Example documentation for x */
   x = 1
 
@@ -137,17 +138,28 @@ end-of-line can be dropped without affecting prefix detection.
   z = 1
 
 Test cases can be included in docstring comments that will be checked
-by the `:check-docstrings` command in the form of code blocks that are
-unlabeled or labeled with the `repl` language. This cases should be
+by the ``:check-docstrings`` command in the form of code blocks that are
+unlabeled or labeled with the ``repl`` language. This cases should be
 in the form of REPL commands. Success of the test is defined to be the
 success of all the REPL commands.
 
 .. code-block:: cryptol
   :caption: Example docstring with checked test
+
   /** This function 
    *
    * ```repl
    * :check f 10 == 20
+   * ```
+   *
+   * Also checked
+   * ```
+   * :safe f
+   * ```
+   *
+   * Not checked
+   * ```cpp
+   * int main() {}
    * ```
    */
   f : [8] -> [8]
--- a/docs/RefMan/BasicSyntax.rst
+++ b/docs/RefMan/BasicSyntax.rst
@@ -107,9 +107,51 @@ end of the line.  Block comments may be nested arbitrarily.
   // This is a line comment
   /* This is a /* Nested */ block comment */
 
-.. todo::
-  Document ``/** */``
+Documentation Comments
+----------------------
 
+Declarations in Cryptol can have documentation attached to them using special
+*docstring* comments. These attached comments are accessible in the REPL
+using the `:help` command.
+
+Syntactically, docstring comments are block comments that start with exactly
+two `*` characters: `/** ... */`. For lines after the first line a common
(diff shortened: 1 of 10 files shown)


docs/RefMan/BasicSyntax.rst
```diff
@@ -107,9 +107,51 @@ end of the line.  Block comments may be nested arbitrarily.
...
   /* This is a /* Nested */ block comment */
 
-.. todo::
-  Document ``/** */``
+Documentation Comments
+----------------------
 
+Declarations in Cryptol can have documentation attached to them using special
+*docstring* comments. These attached comments are accessible in the REPL
+using the `:help` command.
```
RyanGlScott, line 115: Given that this is ReStructured Text (rather than markdown), I believe inline text needs to be surrounded by double backticks:

```suggestion
using the ``:help`` command.
```

And similarly for other snippets of inline code below.

docs/RefMan/BasicSyntax.rst
```diff
@@ -107,9 +107,51 @@ end of the line.  Block comments may be nested arbitrarily.
...
+.. code-block:: cryptol
+  :caption: Example docstring with checked test
+  /** This function 
+   *
+   * ```repl
+   * :check f 10 == 20
+   * ```
+   */
+  f : [8] -> [8]
+  f x = 2 * x
```
RyanGlScott, line 154: Consider also including another docstring coment that uses plain `` ``` ... ``` `` (without `repl`), along with a comment saying that this will be checked, as well as another comment that uses something like `` ```c ... ``` ``, which is not checked.

src/Cryptol/Parser/ParserUtils.hs
```diff
@@ -1010,39 +1008,73 @@ mkDoc ltxt = ltxt { thing = docStr }
...
-
-  trimFront []                     = []
-  trimFront (l:ls)
-    | T.all commentChar l = ls
-    | otherwise           = T.dropWhile commentChar l : ls
-
-  dropPrefix []        = []
-  dropPrefix [t]       = [T.dropWhile commentChar t]
-  dropPrefix ts@(l:ls) =
+  whitespaceChar x = x `elem` (" \t" :: String)
```
RyanGlScott, line 1023: Why not use something like [`Data.Char.isSpace`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-Char.html#v:isSpace) to implement this?

src/Cryptol/REPL/Command.hs
```diff
@@ -2004,15 +2004,13 @@ moduleInfoCmd isFile name
...
+extractCodeBlocks raw = go [] (T.lines raw)
   where
-    cleanLines = T.dropWhile isSpace <$> T.lines raw
-
     go finished [] = Right (reverse finished)
     go finished (x:xs)
       | (ticks, lang) <- T.span ('`' ==) x
       , 3 <= T.length ticks
-      = if lang == "repl"
+      = if lang `elem` ["", "repl"]
```
RyanGlScott, line 2013: It might be worth leaving a comment here explaining what the meaning of the empty string `""` is in this context.

src/Cryptol/REPL/Command.hs
```diff
@@ -2068,14 +2066,23 @@ checkBlock = isolated . go
...
-                , srLog = logtxt
-                , srResult = result
-                }
-          pure (subresult : results)
+          (logtxt, eresult) <- captureLog (Cryptol.REPL.Monad.try (cmd 0 Nothing))
+          case eresult of
+            Left e -> do
+              let result = SubcommandResult
+                    { srInput = line
+                    , srLog = logtxt ++ show (pp e) ++ "\n"
```
RyanGlScott, line 2074: Why use a newline here in the `Left` case (but not in the `Right` case below)?

glguy: The newline is terminating the shown error `show (pp e) ++ "\n"` but nothing is being appended to the `Right` case

src/Cryptol/Parser/ParserUtils.hs
glguy: I think even letting tab get mixed in will create some really confusing situations but I'm happy to allow is space to really let people go wild

PR #1713, opened 2024-07-29
Enhance TypeChecker with exponent rules

### LINKED ISSUE #1489: Constraint fails to be proven
The following function definition is not accepted:
```
id : {k} (fin k, k > 0) => [2^^k] -> [2^^k]
id x = join(split`{2,2^^(k-1)}x)
```

The corresponding error is:
```
[error] at test.cry:2:1--2:33:
    in the definition of 'Main::id', at test.cry:2:1--2:3,
    we need to show that
      for any type k
      assuming
        • fin k
        • k > 0
        • 2 ^^ k == 2 * 2 ^^ (k - 1)
            arising from
            matching types
            at test.cry:2:8--2:12
        • 2 * 2 ^^ (k - 1) == 2 ^^ k
            arising from
            matching types
            at test.cry:2:31--2:32
```
I encountered this in version:
```
 ✗ cryptol --version
Cryptol 2.13.0.99
```

mccleeary-galois:
Closes #1489

### DIFF
--- a/src/Cryptol/TypeCheck/Solver/Numeric.hs
+++ b/src/Cryptol/TypeCheck/Solver/Numeric.hs
@@ -48,6 +48,7 @@ cryIsEqual ctxt t1 t2 =
     <|> tryCancelVar ctxt (=#=) t1 t2
     <|> tryLinearSolution t1 t2
     <|> tryLinearSolution t2 t1
+    <|> tryEqMulExp t1 t2
     <|> tryEqExp t1 t2
 
 -- | Try to solve @t1 /= t2@
@@ -237,8 +238,8 @@ tryCancelVar ctxt p t1 t2 =
 
 
 -- if (x >= 2) && x^a*x^b = x^c => a+b = c
-tryEqExp :: Type -> Type -> Match Solved
-tryEqExp x y = check x y <|> check y x
+tryEqMulExp :: Type -> Type -> Match Solved
+tryEqMulExp x y = check x y <|> check y x
   where 
     check i j =
       do  (m1,m2) <- aMul i
@@ -249,6 +250,19 @@ tryEqExp x y = check x y <|> check y x
           n <- aNat x_1
           guard (n >= 2)
           return $ SolvedIf [ tAdd a b =#= c ]
+
+-- if (x >= 2) && x^a = x^b => a = b
+tryEqExp :: Type -> Type -> Match Solved
+tryEqExp x y = check x y <|> check y x
+  where 
+    check i j =
+      do  
+          (x_1, a) <- (|^|) i
+          (x_2, b) <- (|^|) j
+          guard (x_1 == x_2)
+          n <- aNat x_1
+          guard (n >= 2)
+          return $ SolvedIf [ a =#= b ]
   
 -- min t1 t2 = t1 ~> t1 <= t2
 tryEqMin :: Type -> Type -> Match Solved
--- a/src/Cryptol/TypeCheck/Solver/Numeric.hs
+++ b/src/Cryptol/TypeCheck/Solver/Numeric.hs
@@ -48,6 +48,7 @@ cryIsEqual ctxt t1 t2 =
     <|> tryCancelVar ctxt (=#=) t1 t2
     <|> tryLinearSolution t1 t2
     <|> tryLinearSolution t2 t1
+    <|> tryEqExp t1 t2
 
 -- | Try to solve @t1 /= t2@
 cryIsNotEqual :: Ctxt -> Type -> Type -> Solved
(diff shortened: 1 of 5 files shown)


src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -137,6 +140,17 @@ tryGeqThanK _ t (Nat k) =
   -- XXX: K1 ^^ n >= K2
 
 
+-- (x >= 2 && x^a >= x^b) => a >= b
+tryGeqExp :: Ctxt -> Type -> Type -> Match Solved
+tryGeqExp _ x y = 
+      do  (x_1, a) <- (|^|) x
+          (x_2, b) <- (|^|) y
+          guard (x_1 == x_2)
+          n <- aNat x_1
```
yav, line 149: Small thing, but it might be better to move the `aNat` check earlier, perhaps right after the binding for `x_1`.  This should avoid the rest of the stuff (e.g., comparing the types) if we don't have a number in the base.

src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -137,6 +140,17 @@ tryGeqThanK _ t (Nat k) =
   -- XXX: K1 ^^ n >= K2
 
 
+-- (x >= 2 && x^a >= x^b) => a >= b
```
yav, line 143: The convention I've been using is to name things that only work for constants `K`, so maybe change the `x` to `K`.

src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -223,6 +237,33 @@ tryCancelVar ctxt p t1 t2 =
```
yav, line 239: I think instead of the `tryEqMulExp`, we should add a rewrite to types, to turn `x^a * x ^b` into `x ^ (a + b)`.  Then `tryEqExp` should handle this rule.

src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -223,6 +237,33 @@ tryCancelVar ctxt p t1 t2 =
...
+      do  (m1,m2) <- aMul i
+          (x_1, a) <- (|^|) m1
+          (x_2, b) <- (|^|) m2
+          (x_3, c) <- (|^|) j
+          guard (x_1 == x_2 && x_2 == x_3)
+          n <- aNat x_1
+          guard (n >= 2)
+          return $ SolvedIf [ tAdd a b =#= c ]
+
+-- if (x >= 2) && x^a = x^b => a = b
```
yav, line 254: Same comment as `tryGeqExp`:  use `K` instead of `x` in the comment, and move the `aNat` check as early as possible.

yav (changes requested):

src/Cryptol/TypeCheck/Solver/Numeric.hs
mccleeary-galois: Updated in [110a753](https://github.com/GaloisInc/cryptol/pull/1713/commits/110a753a94506829ac11402ccf4f0067bb18bebe)

src/Cryptol/TypeCheck/Solver/Numeric.hs
mccleeary-galois: Updated in [110a753](https://github.com/GaloisInc/cryptol/pull/1713/commits/110a753a94506829ac11402ccf4f0067bb18bebe)

src/Cryptol/TypeCheck/Solver/Numeric.hs
mccleeary-galois: Updated in [110a753](https://github.com/GaloisInc/cryptol/pull/1713/commits/110a753a94506829ac11402ccf4f0067bb18bebe)

src/Cryptol/TypeCheck/Solver/Numeric.hs
mccleeary-galois: Updated in [110a753](https://github.com/GaloisInc/cryptol/pull/1713/commits/110a753a94506829ac11402ccf4f0067bb18bebe)

yav (approved):

tests/issues/issue1489/issue1489.cry
```diff
@@ -0,0 +1,54 @@
+module ID where
+
+id : {k} (fin k, k > 0) => [2^^k] -> [2^^k]
```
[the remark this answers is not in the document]
mccleeary-galois: This was just an example of a function that used to bomb during compilation, the test is just making sure that we can compile these functions now.

Issue #1715, opened 2024-07-31
Reference evaluator crashes when using `Float`'s `Literal` instance


RyanGlScott:
Currently, the Cryptol reference evaluator can evaluate a `Float` value when using its `FLiteral` instance:

```
Float> :m Float
Loading module Cryptol
Loading module Float
Float> :eval 1.0 : Float64
0x1.0
```

But if you use `Float`'s `Literal` instance, then the reference evaluator crashes:

```
Float> :eval 1 : Float64
cryptol: You have encountered a bug in Cryptol's implementation.
*** Please create an issue at https://github.com/GaloisInc/cryptol/issues

%< --------------------------------------------------- 
  Revision:  65397a491bddc3b4e8f41053106ce8859387d662
  Branch:    release-3.1.0 (uncommited files present)
  Location:  [Reference Evaluator]literal
  Message:   TCon (TC TCFloat) [TCon (TC (TCNum 11)) [],TCon (TC (TCNum 53)) []] cannot represent literals
CallStack (from HasCallStack):
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.1.0-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/Eval/Reference.lhs:1804:19 in cryptol-3.1.0-inplace:Cryptol.Eval.Reference
%< ---------------------------------------------------
```

Fixing this should be a simple matter of adding another case to the [`literal`](https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/src/Cryptol/Eval/Reference.lhs#L1071-L1082) function.

Issue #1714, opened 2024-07-31
Reference evaluator does not implement many floating-point primitives


RyanGlScott:
Many of the primitives defined in [`lib/Float.cry`](https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/lib/Float.cry) are not implemented in the Cryptol reference evaluator, as seen in this example:

```
Float> :m Float
Loading module Cryptol
Loading module Float
Float> fpSqrt rne (25.0 : Float64)
0x5.0
Float> :eval fpSqrt rne (25.0 : Float64)
cryptol: You have encountered a bug in Cryptol's implementation.
*** Please create an issue at https://github.com/GaloisInc/cryptol/issues

%< --------------------------------------------------- 
  Revision:  65397a491bddc3b4e8f41053106ce8859387d662
  Branch:    release-3.1.0 (uncommited files present)
  Location:  [Reference Evaluator]evalPrim
  Message:   Unimplemented primitive
             Float::fpSqrt
CallStack (from HasCallStack):
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.1.0-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/Eval/Reference.lhs:1804:19 in cryptol-3.1.0-inplace:Cryptol.Eval.Reference
%< ---------------------------------------------------
```

We should fill out support for these primitives in this table:

https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/src/Cryptol/Eval/Reference.lhs#L1727-L1792

RyanGlScott:
Relatedly, the reference evaluator's `floatPrimTable` currently has an entry for `fpIsFinite`:

https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/src/Cryptol/Eval/Reference.lhs#L1755-L1759

There is no need for such an entry, however, as `fpIsFinite` isn't defined as a primitive:

https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/lib/Float.cry#L159-L161

As such, we should remove this entry.

PR #1716, opened 2024-07-31
Implement all floating-point primitives and instances in reference evaluator

RyanGlScott:
This PR is a collection of two commits which ensures that the reference evaluator's coverage of `Float` is on part with other Cryptol backends.

## Implement all floating-point primitives in reference evaluator

This ensures that all of the primitives defined in `lib/Float.cry` are supported in Cryptol's reference evaluator. Moreover, this removes the primitive table entry for `fpIsFinite`, as `fpIsFinite` is not defined as a primitive and therefore doesn't need special treatment.

I have added a `tests/regression/float_reference_eval.icry` test case which ensures that the reference evaluator does the expected thing on all floating-point operations.

Fixes https://github.com/GaloisInc/cryptol/issues/1714.

## Implement Float's Literal instance in the reference evaluator

Also extend the `float_reference_eval` test case to ensure that the reference evaluator covers all of `Float`'s instances.

Fixes https://github.com/GaloisInc/cryptol/issues/1715.


src/Cryptol/Eval/Reference.lhs
```diff
@@ -1752,16 +1763,55 @@ true for all supported precisions.
 >                           y <- fromVFloat <$> yv
 >                           pure (VBit (FP.bfCompare x y == EQ))
 >
->    , "fpIsFinite" ~> vFinPoly \_ -> pure $
->                      vFinPoly \_ -> pure $
+>    , "fpIsNaN"     ~> vFinPoly \_ -> pure $
```
mccleeary-galois, line 1766: What set of operations are we targeting here?

> Implement all floating-point primitives

Are we wanting to support all IEEE-754-2019 operations? We are getting close to that with this set (minus the recommended ones in the standard), or is this driven by another programming language target?

mccleeary-galois (approved):
This is certainly an improvement of set of supported operations from previous. I did want to have a discussion about what set we should be targeting fully in the future, but isn't necessary for getting this in.

src/Cryptol/Eval/Reference.lhs
RyanGlScott: By "primitives", I mean all of the Cryptol primitive functions listed in [`lib/Float.cry`](https://github.com/GaloisInc/cryptol/blob/25423ce4ee65337bb75c6922e269a6d359c150e5/lib/Float.cry). (Perhaps unsurprisingly, this list of primitives closely corresponds with the operations provided by SMT-LIB's floating-point theory.)

src/Cryptol/Eval/Reference.lhs
```diff
@@ -1707,13 +1708,23 @@ and integers.
 These functions capture the interactions with rationals.
 
 
-This just captures a common pattern for binary floating point primitives.
+These just capture common patterns for unary, binary, and trinary floating
```
yav, line 1711: I think the word for an operator with 3 arguments is "ternary"?

yav (approved):

src/Cryptol/Eval/Reference.lhs
RyanGlScott: You're right, I'm not sure why I though "trinary" sounded more natural. (It turns out that [trinary is a word](https://en.wiktionary.org/wiki/trinary), but much less common.) I'll change it to ternary (and rename the `*Tri*` functions to `*Tern*`).

PR #1731, opened 2024-08-14
Limit :check to return only locally defined properties

glguy:
This change aligns :check and friends with :check-docstrings. It ensures that these command check exactly the properties defined in the focused module and its locally defined submodules and any locally instantiated functors. This makes it more suitable for use while developing a module and for including in a module docstring.

### DIFF
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -505,7 +505,7 @@ qcCmd qcMode "" _pos _fnm =
                do let str = nameStr x
                   rPutStr $ "property " ++ str ++ " "
                   let texpr = T.EVar x
-                  let schema = M.ifDeclSig d
+                  let schema = T.dSignature d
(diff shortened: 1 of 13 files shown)


yav (commented):
It makes sense to me, but this is a significant change in behavior so we should make sure it won't break too much existing user code.   I do think it's useful to have this version of the command though, so if code would be broken, we could expose this functionality with a different command.

mccleeary-galois:
> This change aligns :check and friends with :check-docstrings. It ensures that these command check exactly the properties defined in the focused module and its locally defined submodules and any locally instantiated functors. This makes it more suitable for use while developing a module and for including in a module docstring.

Should note that this also changes the behavior of `:prove` and `:exhaust` in similar fashion.

weaversa:
LGreatTM! The current behavior of `:check` and friends is nonsensical. A change bringing the scope to 'focused module'-relevant properties is welcome.

PR #1733, opened 2024-08-19
Define `tIsNominal` and use it

### LINKED ISSUE #1732: Missing `tIsNominal` function
`Cryptol.TypeCheck.Type` includes a variety of helper functions for querying if a `Type` is of a particular shape, such as `tIsVar`, `tIsRec`, `tIsTuple`, and so on. One function that is conspicuously absent is a `tIsNominal :: Type -> Maybe (NominalType, [Type])` function, which would return `Just` if the `Type` is `TNominal`. I propose we add it. (This came up in a downstream project.)

RyanGlScott:
This defines a `tIsNominal :: Type -> Maybe (NominalType, [Type])` function. There is one spot in `Cryptol.TypeCheck.Infer` where this can more neatly replace a manual pattern match on `TNominal`, so I went ahead and did this as well.

Fixes #1732.


yav (approved):
Good idea!

PR #1735, opened 2024-08-20
Release 3.2.0

mccleeary-galois:


RyanGlScott (commented):
Thanks, @mccleeary-galois! A couple of process quibbles:

* In a separate commit, please bump the version numbers to `3.2.0.99`, which will be the next development version. That way, when we cut a release branch, we can include everything up to (but not including) the `3.2.0.99` commit.
* On a similar note, you've named the branch for this PR `release-3.2.0`, which is the naming convention we've adopted for release branches. You probably don't want to name this PR's branch `release-3.2.0`, since GitHub is configured to delete PR branches once they're merged. I'd recommend naming this PR's branch something like `prepare-release-3.2.0` instead to avoid this.

  Another reason to pick a name like `prepare-release-3.2.0` is that the CI is configured to run release-specific CI steps on any branch that matches the regex `release-**` (see [here](https://github.com/GaloisInc/cryptol/blob/cc3b3fb786801471b03f05a0dc2eee6fe1a9e84e/.github/workflows/ci.yml#L5)). This is desirable when you push the release branch, but not as desirable when you're just submitting a PR, since those extra CI steps will needlessly spend CI runner cycles. Giving it a name like `prepare-release-3.2.0` is an effective way of making the branch name not match that regex :)

PR #1739, opened 2024-08-27
Start on a mechanism to build Cryptol values from outside Cryptol

yav:
The plan here is to add a mechanism that allows the Cryptol interpreter to cooperate with an external function to build Cryptol values, which would help support more forms of FFI.

Our current FFI to C assumes a fixed data layout for values, and it is the job of Cryptol to parse the values back into Cryptol values, when results a returned by a foreign function.

The new approach exposes a "Cryptol value builder" object to the foreign function, so that the foreign function can build the Cryptol values itself.   The benefit of this is that we only need to impose representation constraints on primitve types (e.g., bool, integer, word), while the representation of compound values types (e.g., vectors, tuples, sums) is unrestricted.


src/Cryptol/Eval/FFI/Abstract.hs
```diff
@@ -0,0 +1,41 @@
...
+import Cryptol.Eval.Value
+import Cryptol.Backend.Monad
+import Cryptol.Backend.Concrete
+import Cryptol.Eval.FFI.Abstract.Export(exportSizes,exportValues)
+import Cryptol.Eval.FFI.Abstract.Call(runFFI)
+
+
+-- | Call a foreign function that follows the "abstract" calling convetion.
+callForeignAbstract ::
+  Name                           {- ^ Name of foregin function -} ->
```
RyanGlScott, line 19:
```suggestion
  Name                           {- ^ Name of foreign function -} ->
```

src/Cryptol/Eval/FFI/Abstract.hs
```diff
@@ -0,0 +1,41 @@
...
+  FFIFunType Type                {- ^ FFI type -} ->
+  ForeignImpl                    {- ^ Address of foreign worker -} ->
+  TypeEnv                        {- ^ Values for numeric type parametres -} ->
```
RyanGlScott, line 22:
```suggestion
  TypeEnv                        {- ^ Values for numeric type parameters -} ->
```

src/Cryptol/Eval/FFI/Abstract/Export.hs
```diff
@@ -0,0 +1,253 @@
...
+-- | Encoding of a word: buffer size, digits
+exportWord :: SWord Concrete -> [ExportVal] -> [ExportVal]
+exportWord (BV sz i) = \start ->
+  if sz <= 8 then EV8 (fromInteger i) : start else
+  if sz <= 64 then EV64 (fromInteger i) : start else
+  EVInteger i : EV64 size : start
+  where
+  !size = integerSize i
+
+-- | Export a sequence of values: tuples, recrods, sequnces.
```
RyanGlScott, line 142:
```suggestion
-- | Export a sequence of values: tuples, records, sequences.
```

src/Cryptol/Eval/FFI/Abstract/Export.hs
```diff
@@ -0,0 +1,253 @@
...
+cryStartExport ::
+    [ExportVal] {-| REVERSED.  Send these to foreign. -} ->
+    IO (StablePtr (IORef [ExportVal]))
+cryStartExport vs =
+  do ref <- newIORef (reverse vs)
+     newStablePtr ref
+
+-- | Get the next data item to export.
+-- We assume that this is the only place to manipulate the reference
+-- so there's not race condition. Note that it is also important
```
RyanGlScott, line 175:
```suggestion
-- so there's not a race condition. Note that it is also important
```

src/Cryptol/Eval/FFI/Abstract/Import.hs
```diff
@@ -0,0 +1,345 @@
+{-# Language OverloadedStrings #-}
+{-# Language BlockArguments #-}
+{-# Language LambdaCase #-}
+-- | Export builders for Cryptols values
```
RyanGlScott, line 4:
```suggestion
-- | Export builders for Cryptol values
```

src/Cryptol/Eval/FFI/Abstract/Import.hs
```diff
@@ -0,0 +1,345 @@
...
+import Cryptol.Backend.Monad
+import Cryptol.Backend.Concrete
+import Cryptol.Backend.SeqMap
+
+import Cryptol.Backend
+
+
+type Value = SEval Concrete (GenValue Concrete)
+
+-- | Imported of Cryptol values (aka "context")
```
RyanGlScott, line 49:
```suggestion
-- | Imported Cryptol values (aka \"context\")
```

src/Cryptol/Parser/ParserUtils.hs
```diff
@@ -895,9 +895,21 @@ mkPrimDecl :: Maybe (Located Text) -> LPName -> Schema PName -> [TopDecl PName]
...
+     fgn <- case mbCC of
+              Nothing -> pure ForeignC
+              Just cc ->
+                case thing cc of
+                  UnQual i
+                     | tx == "c" -> pure ForeignC
+                     | tx == "abstract" -> pure ForeignAbstract
+                     where tx = identText i
+                  _ -> errorMessage (srcRange cc)
+                          [ "Invalid calling convetion."
```
RyanGlScott, line 910:
```suggestion
                          [ "Invalid calling convention."
```

cryptol.cabal
```diff
@@ -98,6 +99,7 @@ library
   if flag(ffi)
     build-depends:     hgmp,
                        libffi            >= 0.2
+    include-dirs:      ./lib
```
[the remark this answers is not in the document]
RyanGlScott: Now that I look more closely, there is another reason to prefer putting this under `lib`: it's because the CI has special handling to ensure that the contents of `lib/` are copied to the `share/` subdirectory when packaging up a bindist. Arguably, `cry_ffi.h` belongs in `share/` as well, although there is a question of whether it should go alongside the other myriad files that currently inhabit `share/` (e.g., the prelude's `.cry` and `.z3` files).

For now, I think I will keep `cry_ffi.h` under `lib/` for consistency, but we may want to revisit this later.

src/Cryptol/Eval/FFI/GenHeader.hs
```diff
@@ -53,28 +54,31 @@ data ConvertResult
...
-  let params = snd $ mapAccumL renameParam Set.empty $
-        typeParams ++ inParams ++ outParams
-      renameParam names (C.Param u name) =
-        (Set.insert name' names, C.Param u name')
-        where name' = until (`Set.notMember` names) (++ "_") name
-  pure $ C.FunDecln Nothing retType (unpackIdent $ nameIdent fName) params
+convertFun :: (Name, FFI) -> GenHeaderM C.Decln
+convertFun (fName, cc) =
+  case cc of
+    CallAbstract {} -> undefined "XXX"
```
RyanGlScott, line 60: We should implement this `undefined`, as it currently causes `:generate-foreign-header` to crash when run on a file containing `foreign abstract` functions.

RyanGlScott (approved):
I'm pretty happy with the current state of this PR, and I think the design of `cry-ffi` is suitably flexible. I propose we land this.

Issue #1741, opened 2024-09-03
Simplify `inf - x` to `inf`


yav:
Since `-` requires its second argument to be finite, subtracting a finite number from `inf` should result in `inf`

RyanGlScott:
Hm, I'm surprised that this case of `nSub` doesn't cover this already:

https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/TypeCheck/Solver/InfNat.hs#L91

yav:
It will only do this if the second argument is a constant.

RyanGlScott:
Fixed in https://github.com/GaloisInc/cryptol/pull/1742.

PR #1742, opened 2024-09-03
Simplify `inf - x` to `inf`

yav:
This PR fixes issue #1741, adding an extra simplification rule to the type checker.

### DIFF
--- a/src/Cryptol/TypeCheck/SimpType.hs
+++ b/src/Cryptol/TypeCheck/SimpType.hs
@@ -113,6 +113,7 @@ tSub :: Type -> Type -> Type
 tSub x y
   | Just t <- tOp TCSub (op2 nSub) [x,y] = t
   | tIsInf y  = tError (tf2 TCSub x y)
+  | tIsInf x = x
   | Just 0 <- yNum = x
   | Just k <- yNum
   , TCon (TF TCAdd) [a,b] <- tNoUser x
--- a/src/Cryptol/TypeCheck/SimpType.hs
+++ b/src/Cryptol/TypeCheck/SimpType.hs
@@ -116,7 +116,7 @@ tSub x y
 
   | tIsInf x = x
     {- This assumes that `y` is finite and not error.  The first should
-       follow from the typing on `tSub`, which asserts that the second argumet
+       follow from the typing on `tSub`, which asserts that the second argument
        is finite and less than the first;  the second should have been handleed
        by the first equation above `tOp`. -}
 
--- a/src/Cryptol/TypeCheck/SimpType.hs
+++ b/src/Cryptol/TypeCheck/SimpType.hs
@@ -117,7 +117,7 @@ tSub x y
   | tIsInf x = x
     {- This assumes that `y` is finite and not error.  The first should
        follow from the typing on `tSub`, which asserts that the second argument
-       is finite and less than the first;  the second should have been handleed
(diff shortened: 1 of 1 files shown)


src/Cryptol/TypeCheck/SimpType.hs
```diff
@@ -113,6 +113,7 @@ tSub :: Type -> Type -> Type
 tSub x y
   | Just t <- tOp TCSub (op2 nSub) [x,y] = t
   | tIsInf y  = tError (tf2 TCSub x y)
+  | tIsInf x = x
```
RyanGlScott, line 116: Should this require that `y` be finite? (I'm thinking of what happens if you have `inf - inf`.)

RyanGlScott (commented):
Please make sure to:

* Add a test case where this simplification rule is important.
* Mention this change in the CHANGELOG, since this is a user-facing change.

src/Cryptol/TypeCheck/SimpType.hs
yav: I think that we should have already asserted that `y` is finite, it is part of the type of `-` (have a look in `Cryptol.cry`).  Also when we make `tSub` terms we should take care to preserve this invariant, although I think we have another tickets where this invariant was violated by a rewrite rule.

RyanGlScott: Ah, good point. Can you add a comment to `tSub` saying that it assumes the precondition that the second argument is finite? (It might also be worth adding similar comments to other functions in this module that assume preconditions, such as `tDiv` and `tMod`.)

src/Cryptol/TypeCheck/SimpType.hs
```diff
@@ -113,6 +113,7 @@ tSub :: Type -> Type -> Type
 tSub x y
   | Just t <- tOp TCSub (op2 nSub) [x,y] = t
   | tIsInf y  = tError (tf2 TCSub x y)
```
RyanGlScott, line 115: If we assume the precondition that the second argument (`y`) is finite, then this case must be unreachable, right?

src/Cryptol/TypeCheck/SimpType.hs
```diff
@@ -113,7 +113,13 @@ tSub :: Type -> Type -> Type
...
+
   | tIsInf x = x
+    {- This assumes that `y` is finite and not error.  The first should
+       follow from the typing on `tSub`, which asserts that the second argumet
```
RyanGlScott, line 119:
```suggestion
       follow from the typing on `tSub`, which asserts that the second argument
```

src/Cryptol/TypeCheck/SimpType.hs
```diff
@@ -113,7 +113,13 @@ tSub :: Type -> Type -> Type
...
+       is finite and less than the first;  the second should have been handleed
```
RyanGlScott, line 120:
```suggestion
       is finite and less than the first;  the second should have been handled
```

src/Cryptol/TypeCheck/SimpType.hs
```diff
@@ -113,7 +113,13 @@ tSub :: Type -> Type -> Type
...
+       by the first equation above `tOp`. -}
```
RyanGlScott, line 121: I'm not sure which equation this refers to. There is a use of `tOp` in the first equation, but there are no equations above it.

PR #1748, opened 2024-09-10
Update type of `take` in book

### LINKED ISSUE #1747: Update prelude `take` type in book
The cryptol book says the type of `take` is:
```
take : {front, back, a} (fin front) => [front + back]a -> [front]a
```

but it's actually 
```
take : {front, back, a} [front + back]a -> [front]a
```

This is very useful for operating over infinite sequences! We should update the book to match the implementation.

I checked all the other types for the "Sequences" category and they are all correct.

marsella:
Closes #1747.

I looked through all the other instances of `take` in the book. The other place with the type signature is correct (allows infinite front) and there are no examples or anything that claim that infinite fronts are disallowed. Also, all the other sequence types are correct.


RyanGlScott:
Looks entirely reasonable to me.

@yav, do we have a policy of needing to regenerate the [PDF copy of the book](https://github.com/GaloisInc/cryptol/blob/c73d1d8b9e0c8b4a274544a3bf14d42a4f1e2600/docs/ProgrammingCryptol.pdf) that is checked into version control whenever the book's source is updated?

RyanGlScott (approved):

PR #1751, opened 2024-09-17
Remove bitvectors width from the value tag.

### LINKED ISSUE #1740: Don't store widths of words twice
The cryptol value for words stored the length of the word and a `WordValue`:

https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/Eval/Value.hs#L127

But the `WordValue` also stored the length of the word:

https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/Backend/WordValue.hs#L91

The only case that does not have an explicit value is the `WordVal` which uses the type family `SWord` to pick a type based on representation.

In the `Concrete` backend this value is `BV`
https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/Backend/Concrete.hs#L135

which also has a length:

https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/Backend/Concrete.hs#L57

In fact, any instance of `SWord` must store its length, because the backend contains this method:

https://github.com/GaloisInc/cryptol/blob/09853251186d4590f3b093dbdf2b324c9a7e6ca0/src/Cryptol/Backend.hs#L298

I think we can remove the length from `VWord` and just get it from teh `WordValue`.

yav:
Fixes #1740.  We can do this because the values (WordVal) already has its width, so we don't need to store it again.

### DIFF
--- a/src/Cryptol/Backend.hs
+++ b/src/Cryptol/Backend.hs
@@ -38,6 +38,7 @@ module Cryptol.Backend
 import qualified Control.Exception as X
 import Control.Monad.IO.Class
 import Data.Kind (Type)
+import Data.Proxy(Proxy(..))
 
 import Cryptol.Backend.FloatHelpers (BF)
 import Cryptol.Backend.Monad
@@ -297,7 +298,7 @@ class MonadIO (SEval sym) => Backend sym where
   bitAsLit :: sym -> SBit sym -> Maybe Bool
 
   -- | The number of bits in a word value.
-  wordLen' :: f sym -> SWord sym -> Integer
+  wordLen' :: proxy sym -> SWord sym -> Integer
 
   -- | Determine if this symbolic word is a literal.
   --   If so, return the bit width and value.
@@ -817,5 +818,5 @@ type FPArith2 sym =
   SEval sym (SFloat sym)
 
 wordLen :: forall sym. Backend sym => sym -> SWord sym -> Integer
-wordLen _ x = wordLen' (Nothing :: Maybe sym) x
+wordLen _ x = wordLen' (Proxy :: Proxy sym) x
 {-# INLINE wordLen #-}
--- a/src/Cryptol/Backend.hs
+++ b/src/Cryptol/Backend.hs
@@ -1,7 +1,9 @@
 {-# Language FlexibleContexts #-}
 {-# Language TypeFamilies #-}
+{-# Language ScopedTypeVariables #-}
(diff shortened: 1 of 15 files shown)


src/Cryptol/Backend.hs
```diff
@@ -295,7 +297,7 @@ class MonadIO (SEval sym) => Backend sym where
   bitAsLit :: sym -> SBit sym -> Maybe Bool
 
   -- | The number of bits in a word value.
-  wordLen :: sym -> SWord sym -> Integer
+  wordLen' :: f sym -> SWord sym -> Integer
```
RyanGlScott, line 300: I wonder if we should rename `f` to `proxy` to signal the intent that this argument is never used at runtime.

src/Cryptol/Backend.hs
```diff
@@ -813,3 +815,7 @@ type FPArith2 sym =
   SFloat sym ->
   SFloat sym ->
   SEval sym (SFloat sym)
+
+wordLen :: forall sym. Backend sym => sym -> SWord sym -> Integer
+wordLen _ x = wordLen' (Nothing :: Maybe sym) x
```
RyanGlScott, line 820: Similarly, I wonder if we should pass `Proxy @sym` instead of `Nothing :: Maybe sym` here (and elsewhere in this PR where we use `Nothing :: Maybe sym`.

src/Cryptol/Eval/Generic.hs
```diff
@@ -1734,7 +1734,7 @@ errorV sym _ty msg =
 --   and return the associated character, if it is concrete.
 --   Otherwise, return a '?' character
 valueToChar :: Backend sym => sym -> GenValue sym -> SEval sym Char
-valueToChar sym (VWord 8 wval) =
+valueToChar sym (VWord wval) =
```
RyanGlScott, line 1737: This change appears to have changed the semantics of this function slightly, as it now matches _any_ word size instead of just 8-bit words. Perhaps the pattern guards below should check that the `wordValWidth` is 8?

RyanGlScott (commented):
Please make sure to mention this in the changelog. (If nothing else, the API changes will impact SAW.)

PR #1754, opened 2024-09-19
ci: add a check for the cryptol book pdf

marsella:
Follow up to #1748.

This adds a small CI job that will shout if you either
- update the ProgrammingCryptol PDF without changing the source code
- update the book source code without changing the PDF

@sauclovian-g Do you want this requirement to be documented in a README somewhere? Or maybe `CONTRIBUTING.md`?

### DIFF
--- a/.github/check_book_update.sh
+++ b/.github/check_book_update.sh
@@ -2,6 +2,13 @@
 #
 # Indicates whether the Programming Cryptol book PDF needs to be updated.
 #
+# Usage
+# This is primarily for use in CI. You can run it locally using the following:
+# > bash check_book_update.sh $(git diff --name-only --diff-filter ACDMRT master)
+#
+# If you are merging to a branch other than `master`, use that branch name
+# instead.
+#
 
 TEX_CHANGED=0
 PDF_CHANGED=0
--- a/.github/check_book_update.sh
+++ b/.github/check_book_update.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Indicates whether the Programming Cryptol book PDF needs to be updated.
+#
+
+TEX_CHANGED=0
+PDF_CHANGED=0
+
+# Examine the set of changed files to see if either the book source code
+# or the book PDF were changed.
+for fname in $@ ; do
+    case $fname in
+        docs/ProgrammingCryptol/*)
+            TEX_CHANGED=1
+            TEX_FILES="$TEX_FILES$fname " ;;
+        docs/ProgrammingCryptol.pdf)
+            PDF_CHANGED=1 ;;
+    esac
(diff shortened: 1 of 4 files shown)


.github/check_book_update.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Indicates whether the Programming Cryptol book PDF needs to be updated.
+#
```
RyanGlScott, line 4: It might be helpful to include a usage guide that explains how one might run this script locally. (The `git diff` invocation one needs is somewhat non-obvious.)

sauclovian-g:
Yeah, it would be good to document that. contributing.md is probably the right place (make a section about updating the documentation, basically should say something like any required documentation updates should be included with pull requests, these are the different pieces of documentation, this is how you build each one -- I'm happy to write all that if you don't want to but I don't have all the details at my fingertips)

sauclovian-g (commented):
Apart from the usage point Ryan made, LGTM.

This could be written to not depend on bash but there's probably no point.

.github/check_book_update.sh
marsella: Good call, I added.

marsella:
> contributing.md is probably the right place

I added an overview of the docs I could find. I didn't go super deep but I tried to cover everything I could see in the `docs/` directory.

> This could be written to not depend on bash

I thought you might say that 😁 I figured I had a decent argument in that it's primarily being run in CI and the ubuntu base image comes with bash by default so we're not going to have to negotiate with all our supported platforms.

RyanGlScott (approved):

sauclovian-g:
> > contributing.md is probably the right place
> 
> I added an overview of the docs I could find. I didn't go super deep but I tried to cover everything I could see in the `docs/` directory.

Thanks!

> > This could be written to not depend on bash
> 
> I thought you might say that 😁 I figured I had a decent argument in that it's primarily being run in CI and the ubuntu base image comes with bash by default so we're not going to have to negotiate with all our supported platforms.

I always do :-) shell scripting is an ancient art and bash feature bloat is not part of it. But in practice it doesn't really matter -- the important part is to avoid bashisms if you have ```#!/bin/sh```, because that really will fail and not just on BSD. Requiring that bash is installed is vaguely annoying... but de facto it's very difficult to avoid so in our context it isn't worth spending measurable amounts of time on.

There are two reasons to avoid using bash at all; one is that it's extremely bloaty (load image is about 1.3M on netbsd vs. around 200K for /bin/sh) and the other is that it's phenomenally slow compared to other sh implementations. But neither of these things actually matters in practice most of the time.

sauclovian-g (approved):

PR #1757, opened 2024-09-27
Allow doctests to span multiple lines with a trailing \ like the REPL already supports

glguy:
Additionally this allows empty doctest lines to be skipped without error

### DIFF
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -2065,7 +2065,9 @@ checkBlock ::
 checkBlock = isolated . go . continuedLines
   where
     go [] = pure []
-    go (line:block) =
+    go (line:block)
+      | T.all isSpace line = go block
+      | otherwise =
       case parseCommand (findNbCommand True) (T.unpack line) of
         Nothing -> do
           pure [SubcommandResult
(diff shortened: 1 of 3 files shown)


glguy:
Docstrings were meant to work like the REPL - this was an omission in the initial implementation

src/Cryptol/REPL/Command.hs
```diff
@@ -2104,6 +2106,19 @@ checkBlock = isolated . go
...
+continuedLines :: [T.Text] -> [T.Text]
+continuedLines xs =
+  case span ("\\" `T.isSuffixOf`) xs of
+    ([], []) -> []
+    (a, []) -> [concat' (map T.init a)] -- permissive
+    (a, b:bs) -> concat' (map T.init a ++ [b]) : continuedLines bs
+  where
+    -- concat that eats leading whitespace between elements
+    concat' [] = T.empty
+    concat' (y:ys) = T.concat (y : map (T.dropWhile isSpace) ys)
```
RyanGlScott, line 2119: What happens if you have an example like this?

```
/**
 ** ```repl
 ** :check f 1\
 ** 0
 ** ``` */
```

Does this get interpreted as `:check f 10` or `:check f 1 0`?

glguy: It gets interpreted as 10 which is consistent with C (as an example). This will be useful for long literals, too:

```
x = 0x0123456789abcdef\
      0123456789abcdef\
      0123456789abcdef
```

RyanGlScott: That makes sense—thanks for the explanation. It might be worth testing this property in the `T02` test case.

RyanGlScott (approved):

PR #1758, opened 2024-09-30
Correctly detect top-level parameterized modules in modules() API call

### LINKED ISSUE #1756: `Connection.modules` does not recognize parameterized module with explicit `interface`(s) as `parameterized`
```
modules$ for cry in *.cry; do echo "// $cry"; cat $cry; echo; done
// Functor.cry
module Functor where
  import interface I1
  import interface I2

// I1.cry
interface module I1 where
  type v: *

// I2.cry
interface module I2 where
  type w: *

modules$ export CRYPTOLPATH="$(pwd)"
modules$ python
>>> from pprint import pprint
>>> from cryptol import connect
>>> pprint(load_result := c.load_module('Functor').result())
[]
>>> pprint(modules := c.modules().result())
[{'module': 'Functor', 'parameterized': False},
 {'module': 'Cryptol', 'parameterized': False}]
```

glguy:
Fixes #1756

### DIFF
--- a/cryptol-remote-api/src/CryptolServer/Modules.hs
+++ b/cryptol-remote-api/src/CryptolServer/Modules.hs
@@ -47,11 +47,12 @@ visibleModulesDescr =
   Doc.Paragraph [Doc.Text "List the currently visible (i.e., in scope) module names."]
 
 visibleModules :: VisibleModulesParams -> CryptolCommand [VisibleModuleInfo]
-visibleModules _ = concatMap (moduleToInfos False) . loadedModules <$> getModuleEnv
+visibleModules _ = concatMap moduleToInfos . loadedModules <$> getModuleEnv
 
 
-moduleToInfos :: PP a => Bool -> T.ModuleG a -> [VisibleModuleInfo]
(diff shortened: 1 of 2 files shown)


cryptol-remote-api/src/CryptolServer/Modules.hs
```diff
@@ -52,14 +52,15 @@ visibleModules _ = concatMap (moduleToInfos False) . loadedModules <$> getModule
 
 moduleToInfos :: PP a => Bool -> T.ModuleG a -> [VisibleModuleInfo]
 moduleToInfos p m =
+  let p' = p || isParametrizedModule m in
```
RyanGlScott, line 55: Possibly dumb question: now that we explicitly check `isParameterizedModule m` on each recursive call, do we need to pass the `Bool` argument anymore? The only place where the `Bool` argument changes is on each recursive call to each of the `mFunctors` (where `True` is passed instead of `False`), but I would expect `isParametrizedModule` to always return `True` on nested functors anyway.

RyanGlScott (approved):

PR #1759, opened 2024-10-01
renderOneLine should render to one line

glguy:
Previously the implementation was using layoutCompat with the hope that this operation rendered to a single line. It doesn't. This function is only used in the obscure :dumptests command where rendering to a single line is important for the file format.


glguy (commented):
I don't know how to test this easily with our current test framework. It improves the formatting of random output, so simple gold testing isn't going to cut it.

RyanGlScott:
> It improves the formatting of random output, so simple gold testing isn't going to cut it.

Will the random output be deterministic if you call `:set-seed` beforehand?

src/Cryptol/REPL/Command.hs
```diff
@@ -351,6 +351,7 @@ commandList  =
              , "expression into a file. The first column in each line is"
              , "the expected output, and the remainder are the inputs. The"
              , "number of tests is determined by the \"tests\" option."
+             , "Use filename \"-\" to write tests to stdout."
```
RyanGlScott, line 354: This is a new feature, so it would be worth calling this out in the changelog.

glguy:
Link issue #1314

RyanGlScott (approved):

PR #1764, opened 2024-10-31
Basic syntax highlighting for Cryptol in Emacs

kquick:


RyanGlScott:
It's worth noting that we have been moving towards consolidating all of the syntax-highlighting utilities into centralized repos. For instance, we now have a https://github.com/GaloisInc/galois-vim-syntax repo that contains Vim syntax highlighters for Cryptol, Daedalus, and SAW. I wonder if we should do something similar for Emacs syntax highlighters.

mccleeary-galois:
> It's worth noting that we have been moving towards consolidating all of the syntax-highlighting utilities into centralized repos. For instance, we now have a https://github.com/GaloisInc/galois-vim-syntax repo that contains Vim syntax highlighters for Cryptol, Daedalus, and SAW. I wonder if we should do something similar for Emacs syntax highlighters.

That is a good idea for emacs as well, lets not stop this from going in at the moment though.

RyanGlScott:
Certainly, my comment should not be interpreted as a criticism, just that we should be mindful of this in the future if we add more Emacs syntax highlighters in the future. (I could be mistaken, but I believe that this one is the only Emacs one we have so far, so I think it's fine to store this in the `cryptol` repo for now.)

I'm not an Emacs user, so I don't have any valuable feedback to give on the technical contents of the PR itself.

kquick:
I had not known about that existing repo: thank you!  From a technical perspective, it is more complete (although a bit out of date for things like the keywords).

I'd propose we pause this MR, remove the vim syntax from this repo, and modify the README to refer to the vim-syntax repo.  Meanwhile, I'll propose an update MR to the thoughtpolice repo and if that repo is still supporting updates, we can update the cryptol README to reference it; otherwise I'll fork that one to a Galois emacs-syntax repo and we can reference that one.

yav:
@kquick should we close this PR or merge it in?

PR #1768, opened 2024-11-14
Fix doc-gen CI to not fail on the existsence of a non-mergeable PR.

kquick:
The doc generation run as part of CI attempts to build and deploy a version of the docs for each PR (which allows inspection of doc updates for a PR before it lands).  The docs build is done for the expected marge result (`merge_commit_sha`) of the PR, but if the PR has a merge conflict with the main branch, then the value for this field will be `null`, and thus this CI script will fail.

This change modifies the doc generation CI process to skip PR's that are in this state.


sauclovian-g (approved):
LGTM

PR #1769, opened 2024-11-17
Add CVE checks for Haskell & Python packages

mtullsen:
- Per the title.
- For more detail, see (readme)[cve-reports/README.md]


cve-reports/hs-cves.json
mccleeary-galois, line 1: Let's not commit these into the repo.

cve-reports/bin/cabal-audit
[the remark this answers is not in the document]
mtullsen: accidental! Thanks

cve-reports/hs-cves.json
mtullsen: A-OK.

PR #1774, opened 2024-11-19
Fix `freeVars` calculation for `TNominal`

### LINKED ISSUE #1773: Incorrect `FreeVars Type` case for newtypes and enums
While auditing the code in `Cryptol.IR.FreeVars` recently, @yav and I noticed a bug in the `FreeVars Type` instance involving `TNominal` (i.e., newtypes and enums):

https://github.com/GaloisInc/cryptol/blob/9821aa9deb889b92f12a3363c70e284da87a5091/src/Cryptol/IR/FreeVars.hs#L162

There are two things wrong here:

1. This collects the free variables of `nt` (a `NominalType`), but the `FreeVars NominalType` does not return the name of the nominal type (intentionally so). As such, if you had `newtype N = { x : Bit }`, then `freeVars N` would not include `N`, which is a pretty serious bug.
2. Less seriously, the code here overapproximates the set of free variables by collecting all of the free variables from the _definition_ of `nt`. For example, `freeVars N` will include `Bit` because the `FreeVars NominalType` instance includes the free variables of `N`'s fields, even though the _type_ `N` doesn't directly mention `Bit`.

To fix both of these issues, we should modify the instance so that it only includes `ntName nt` (but no other parts of `nt`) in the `Deps` that this returns. 

In practice, these issues are unlikely to cause anything bad to happen within Cryptol itself. This is because the `FreeVars` instances are primarily used for sorting declarations in dependency order, but we only sort top-level functions, not nominal type declarations. Still, it would be good to fix this so that Cryptol API users _could_ sort nominal type declarations in dependency order if they wanted to.

RyanGlScott:
Rather than computing the free variables of a `NominalType` definition, we only include the name of the `NominalType` itself as a free variable.

Fixes #1773.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@
 * Fix a bug in which splitting a `[0]` value into type `[inf][0]` would panic.
   ([#1749](https://github.com/GaloisInc/cryptol/issues/1749))
 
+* Fix a bug in which the free variables of types mentioning newtypes or enums
+  were incorrectly computed.
+  ([#1773](https://github.com/GaloisInc/cryptol/issues/1773))
+
 ## New features
 
 * REPL command `:dumptests <FILE> <EXPR>` updated to write to stdout when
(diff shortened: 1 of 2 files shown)


yav (approved):
LGTM

Issue #1777, opened 2024-11-27
Make `isNumeric` include `prime` constraints


RyanGlScott:
`Cryptol.TypeCheck.Solver.SMT` includes a [`checkUnsolvable`](https://github.com/GaloisInc/cryptol/blob/4485d702e3c0b863afa22a6aafc696180c9bb877/src/Cryptol/TypeCheck/Solver/SMT.hs#L210-L223) function, which can be used to check if a set of Cryptol-related constraints are unsatisfiable. One thing that `checkUnsolvable` does is to only filter out the _numeric_ constraints—that is, constraints that talk about types of kind `#`. Numeric constraints are classified by the `isNumeric` function:

https://github.com/GaloisInc/cryptol/blob/4485d702e3c0b863afa22a6aafc696180c9bb877/src/Cryptol/TypeCheck/Solver/SMT.hs#L349-L353

Curiously, `isNumeric` does not consider `prime n` constraints, even though `prime : # -> Prop`. While this doesn't impact anything in Cryptol, since Cryptol itself doesn't use `checkUnsolvable`, it _does_ impact external tools that use `checkUnsolvable`. One unfortunate consequence of this is that external tools would mistakenly conclude that a constraint like `prime inf` is satisfiable, even though it is not.

We should add `prime` to `isNumeric`. Doing so will require extending `lib/CryptolTC.z3` with an SMT-LIB definition of `prime`, or at least a partial definition. (See [`cryExp`](https://github.com/GaloisInc/cryptol/blob/4485d702e3c0b863afa22a6aafc696180c9bb877/src/Cryptol/TypeCheck/Solver/SMT.hs#L349-L353) for an example of a partial definition of exponentiation from which we can take inspiration.)

PR #1778, opened 2024-11-27
Add `prime` to `isNumeric`, define `cryPrime` SMT-LIB function 

RyanGlScott:
Fixes https://github.com/GaloisInc/cryptol/issues/1777.

### DIFF
--- a/lib/CryptolTC.z3
+++ b/lib/CryptolTC.z3
@@ -75,6 +75,23 @@
   (cryBool true)
 )
 
+
+(declare-fun cryPrimeUnknown (Int) Bool)
+
+(define-fun cryPrimeTable ((x Int)) Bool
+  (or (= x 2)
+  (or (= x 3)
+  (or (= x 5)
+  (or (= x 7)
+  (cryPrimeUnknown x)))))
+)
+
+(define-fun cryPrime ((x InfNat)) MaybeBool
+  (ite (isErr x) cryErrProp
+    (cryBool (and (isFin x) (cryPrimeTable (value x)))))
+)
+
+
 ; ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lib/CryptolTC.z3
```diff
@@ -75,6 +75,24 @@
...
+
+(declare-fun cryPrimeUnknown (Int) Bool)
+
+(define-fun cryPrimeTable ((x Int)) Bool
+  (or (= x 2)
+  (or (= x 2)
+  (or (= x 3)
+  (or (= x 5)
+  (or (= x 7)
+  (cryPrimeUnknown x))))))
```
yav, line 87: I wonder if we should bother with this at all (i.e., maybe just leave it as uninterpreterd?)  I suspect in the cases where this is used, the primes are typically much larger...   Also we have 2 twice.

yav (approved):

lib/CryptolTC.z3
RyanGlScott: > I wonder if we should bother with this at all (i.e., maybe just leave it as uninterpreterd?)

A good question. I only bothered with this because `cryExp` also goes through the effort of defining a table, so I followed suit here:

https://github.com/GaloisInc/cryptol/blob/4485d702e3c0b863afa22a6aafc696180c9bb877/lib/CryptolTC.z3#L276-L281

If we decide to remove the table for `cryPrime`, we probably should remove it for `cryExp` as well. (I'm not really sure how to test whether these are useful or not.)

> Also we have 2 twice.

Oops! I'll fix that.

RyanGlScott: Fair enough, you've convinced me. I've removed `cryPrimeTable`.

PR #1781, opened 2024-12-06
Fix tab completion for `:t` and `:check`

### LINKED ISSUE #1780: Tab completion on check and friends
Right now, the Cryptol repl supports tab completion within the `:prove` and `:exhaust` commands, but not for `:check` or `:type`. It would be a very nice quality-of-life improvement to add completion!

Related issue #781.

RyanGlScott:
If the user types `:<CMD> ...` and `<CMD>` matches a defined command exactly, then commit to that command's tab completion. (Previously, the tab completion logic would consider all possible commands for which `<CMD>` is a prefix, but this is silly when you have an exact match with an existing command.)

Fixes #1780.


cryptol/REPL/Haskeline.hs
```diff
@@ -223,22 +223,32 @@ canDisplayColor = io (hSupportsANSI stdout)
...
 
+  completeCommand ::
+    String -> String -> CommandDescr -> REPL (String, [Completion])
+  completeCommand cmd rest c
+    | null rest && not (any isSpace l') = do
+      return (l, cmdComp cmd c)
+    | otherwise = do
+      (rest',cs) <- cmdArgument (cBody c) (reverse (sanitize rest),r)
+      return (unwords [rest', reverse cmd],cs)
+
```
sauclovian-g, line 251: This completes ```:check```<tab> to ```:check ``` as well as handling the completions for :check's arguments. So if you type ```:ch<tab>``` it completes as far as ```:check``` but then another tab doesn't tell you what the remaining expansions are but just completes ```:check```. This seems not entirely desirable.

I think it should check ```null rest && not (any isSpace l')``` first and only go on to look at ```findCommandExact``` if that isn't true. (And in that case I think it wants to only look at ```findCommandExact```...)

RyanGlScott: Yes, a fair point. On a related note, it would be nice to add some unit tests for this code, since there are quite a few interesting cases here that are tedious to check manually.

RyanGlScott: I've pushed an additional commit which adds some scaffolding for Cryptol API unit tests, and I've filled out the test cases that work thus far. I haven't gotten around to actually implementing your suggested fix, however. I'm not sure when I will get back around to working on this, so if someone would like to finish this PR in the interim, please feel free to push to the branch.

RyanGlScott: I've pushed a commit which fixes the bugs that @sauclovian-g noticed, and I've added even more regression tests. PTAL.

sauclovian-g (approved):
LGTM

PR #1783, opened 2024-12-09
Add a `tcSmtFile` REPL option

### LINKED ISSUE #1782: Offer a `tcSmtFile` option for writing typechecker SMT solver output to a file
There are two situations in which Cryptol can interact with an SMT solver:

1. When running the `:prove` command. This SMT solver is controlled by the `prover` option.
2. When typechecking code. This SMT solver is controlled by the `tcSolver` option.

With (1), one has the ability to write the SMT solver interactions to a file by enabling the `smtFile` option. There is no such option for (2), however, which means that the only way to record typechecker-related SMT solver interactions is to enable `tcDebug` and copy-paste the output that gets printed to `stdout`, which will look something like this:

```
[send->] (set-option :print-success true )
[<-recv] success
[send->] (set-option :produce-models true )
[<-recv] success
[send->] (set-option :global-decls false )
[<-recv] success
[send->] (declare-datatypes () ((InfNat (mk-infnat (value Int ) (isFin Bool ) (isErr Bool ) ) ) ) )
[<-recv] success
...
```

But this is not a valid SMT-LIB file, so one then has to scrub out the `[send->] ` prefixes and remove the `[<-recv] ` lines entirely. All told, this is quite tedious.

To avoid this tedium, Cryptol should offer a `tcSmtFile` option, where `tcSmtFile` is to `tcSolver` as `smtFile` is to `prover`. Moreover, it should format the typechecker-related SMT solver interactions such that it is a valid SMT-LIB file, e.g.,

```
[send->] (set-option :print-success true )
; success
[send->] (set-option :produce-models true )
; success
[send->] (set-option :global-decls false )
; success
[send->] (declare-datatypes () ((InfNat (mk-infnat (value Int ) (isFin Bool ) (isErr Bool ) ) ) ) )
; success
...
```

RyanGlScott:
`tcSmtFile` is to `tcSolver` as `smtFile` is to `prover`. The `tcSmtFile` option allows users to record typechecker-related SMT solver interactions to a file.

This requires using a more recent `simple-smt` version to bring in the changes from https://github.com/yav/simple-smt/pull/25 and https://github.com/yav/simple-smt/pull/27, which are needed to plumb the relevant information down to the `simple-smt` API.

Fixes #1782.


src/Cryptol/TypeCheck/Solver/SMT.hs
```diff
@@ -80,13 +84,24 @@ setupSolver s cfg = do
...
 startSolver onExit sCfg =
-   do logger <- if (solverVerbose sCfg) > 0 then SMT.newLogger 0
-
-                                     else return quietLogger
+   do (logger, mbLoggerHdl) <-
+        if (solverVerbose sCfg) > 0
+        then do loggerHdl <-
+                  case solverSmtFile sCfg of
+                    Nothing   -> pure stdout
+                    Just file -> openFile file WriteMode
```
yav, line 92: I think we need a way to close this file.  Maybe here we can return a `Maybe (IO ())` which will be `Just (hClose hande)` and this should be added to `onExit`?

src/Cryptol/TypeCheck/Solver/SMT.hs
```diff
@@ -100,7 +115,9 @@ startSolver onExit sCfg =
 
 -- | Shut down a solver instance
 stopSolver :: Solver -> IO ()
-stopSolver s = void $ SMT.stop (solver s)
+stopSolver s =
+  do void $ SMT.stop (solver s)
+     mapM_ hClose (loggerHandle s)
```
yav, line 120: If we make the change above, then we don't need to close the handle here.  I think keeping all the solver cleanup stuff in a single place is probably a good idea.

src/Cryptol/TypeCheck/Solver/SMT.hs
RyanGlScott: Oh dear, I completely overlooked `onExit` here. I'll fold the `Handle`-closing logic into that per your suggestion.

PR #1787, opened 2025-01-06
Add Bitwuzla support

### LINKED ISSUE #1786: Support Bitwuzla
We should add support for the [Bitwuzla](https://bitwuzla.github.io/) SMT solver in Cryptol, which offers efficient reasoning over bit-vectors, floating-points, arrays and uninterpreted functions. Both What4 and SBV now support Bitwuzla, so we should be able to add Bitwuzla support in Cryptol relatively easily.

RyanGlScott:
Fixes #1786.


Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN    ! $(cryptol -c ":s prover=yices" | tail -n +2 | grep -q .) \
     && ! $(cryptol -c ":s prover=cvc4" | tail -n +2 | grep -q .) \
     && ! $(cryptol -c ":s prover=cvc5" | tail -n +2 | grep -q .) \
     && ! $(cryptol -c ":s prover=abc" | tail -n +2 | grep -q .) \
-    # && ! $(cryptol -c ":s prover=boolector" | tail -n +2 | grep -q .) \
+    && ! $(cryptol -c ":s prover=bitvector" | tail -n +2 | grep -q .) \
```
sauclovian-g, line 49: typo patrol: don't think you mean "bitvector" here :-)

sauclovian-g (approved):

sauclovian-g:
(also did you mean the test to use cvc5 and not bitwuzla?)

Dockerfile
RyanGlScott: Indeed, this should be `bitwuzla`. (That reminds me—I should make sure that the Dockerfile continues to work after these changes.)

RyanGlScott:
Oops, I indeed meant to use `bitwuzla` instead of `cvc5` in that test case. Fixed.

PR #1791, opened 2025-01-23
Update modNameChunksText to match modNameToText behavior

glguy:


glguy (commented):
PR made to address the comment at https://github.com/GaloisInc/saw-script/pull/2174#issuecomment-2604755467

I chatted with Iavor and this seems to be what the code should be doing.

RyanGlScott (approved):
I've updated GaloisInc/saw-script#2174, and [the integration tests now pass](https://github.com/GaloisInc/saw-script/actions/runs/12937231324/job/36087221031)! Thank you so much for fixing this.

PR #1792, opened 2025-01-30
Check the repository owner in CI tasks

sauclovian-g:
I discovered recently after updating a fork I'd had sitting around that its periodic CI runs on master were failing.

Apparently it isn't possible to have scheduled runs that are skipped in forks (see https://github.com/orgs/community/discussions/16109 for an open/ignored feature request) but it is at least possible to have them not break.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,7 @@ jobs:
       - if: runner.os == 'Windows'
         run: .github/wix.ps1
 
-      - if: runner.os == 'Windows' && github.event.pull_request.head.repo.fork == false
+      - if: runner.os == 'Windows' && github.event.pull_request.head.repo.fork == false && github.repository_owner == 'GaloisInc'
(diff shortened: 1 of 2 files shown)


RyanGlScott (approved):
This is great! We should also port this over to the `crucible` and `saw-script` CI, both of which do signing and Docker builds.

kquick (commented):
Is there a reason to check `github.event.pull_request.head.repo.fork` if we are testing `github.repository_owner`?  Does the latter subsume the former?

It would also be good to add some comments in the yaml file explaining the purpose of using `github.repository_owner` over a `pull_request` test (i.e. the scheduled tasks, which do not have a pull request context).

sauclovian-g:
That's a good question. I think it won't if we create our own fork within the GH organization (or at least, it would make sense that way) ... it is not _completely_ implausible that we might do that sometime.

Re the comments I'll make another changeset...

PR #1795, opened 2025-02-07
`ChaChaPolyCryptolIETF`: Return an `Option` in `AEAD_CHACHA20_POLY1305_DECRYPT`

RyanGlScott:
This copies over changes made to the upstream `cryptol-specs` version of the `ChaChaPolyCryptolIETF` spec, up to and including a change to make the `AEAD_CHACHA20_POLY1305_DECRYPT` function return an `Option` (as done in GaloisInc/cryptol-specs#240). This is in pursuit of the eventual goal of purging uses of the `(a, Bit)` Cryptol anti-pattern in favor of returning `Option` types, as described in GaloisInc/cryptol-specs#239.)
    
This checks off one box in GaloisInc/cryptol-specs#239.

### DIFF
--- a/examples/ChaChaPolyCryptolIETF.md
+++ b/examples/ChaChaPolyCryptolIETF.md
@@ -337,7 +337,7 @@ diagonal round needs to have an inverse permutation calculated, which we do
 here:
 
 ```cryptol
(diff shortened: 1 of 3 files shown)


marsella (approved):
Looks good to me! I did not build it separately, assuming that you pasted the version from `cryptol-specs` and added the note.

PR #1801, opened 2025-02-12
Implement a smart constructor for EPropGuard

glguy:
This allows us to simplify away trivially satisfied type-case in a single place. In particular this helps later compiler passes that work on specialized instances of polymorphic functions.


RyanGlScott (commented):
Nice!

Perhaps it is worth adding a comment in the Haddocks for `EPropGuards` recommending that users construct it using the `ePropGuards` constructor rather than using `EPropGuards` directly?

RyanGlScott (approved):

PR #1803, opened 2025-02-13
Typechecker: Simplify `width(2^^n)` to `n + 1`

### LINKED ISSUE #1802: `cryptol-specs` regression due to changes in #1800
The changes in #1800 unintentionally caused Cryptol to no longer be able to typecheck the code in the SPHINCS+ spec, as seen in https://github.com/GaloisInc/cryptol-specs/pull/244. Here is a minimized example of a program that works with Cryptol 3.2.0, but does not work with a nightly version of Cryptol:

```
module Bug where

parameter
  type loglogw : #
  type constraint (loglogw >= 1, loglogw <= 3)

type logw = 2^^loglogw
type w = 2^^logw

csum : [32][logw] -> [logw+5]
csum msg = sum [`w - 1 - zext msgi | msgi <- msg]
```

With a nightly Cryptol, the error message is:

```
[error] at Bug.cry:1:1--11:50:
    when checking the module's parameters,
    we need to show that
      for any type Bug::`parameter` interface of Bug::loglogw
      assuming
        • Bug::`parameter` interface of Bug::loglogw >= 1
        • Bug::`parameter` interface of Bug::loglogw <= 3
        • 5 + 2 ^^ Bug::`parameter` interface of Bug::loglogw >=
            width (2 ^^ 2 ^^ Bug::`parameter` interface of Bug::loglogw)
            arising from
            at Bug.cry:11:17--11:35
```

The error message here is somewhat dense (see #1560), but the root of the issue is that Cryptol is trying to prove that `5 + 2^^x >= width (2^^(2^^x))`, but it cannot. Arguably, the fact that it typechecked before was a fluke (as Cryptol was incorrectly simplifying `2^^(2^^x)` to `2^^(2*x)`; see #1799), but it is still unfortunate that this program has regressed.

Although Cryptol's reasoning about exponentiation is incomplete in general, there is fortunately a relatively simple tweak we could make to fix this regression. We could add a simplification rule that allows simplifying types of the form `width(2^^n)` to `n + 1`. This would simplify the constraint above to `5 + 2^^x >= 2^^x + 1`, which Cryptol can easily solve for.

RyanGlScott:
This simplification ends up being critical in order to typecheck the code in the SPHINCS+ spec in `cryptol-specs`, as Cryptol's reasoning about exponentiation is otherwise not complete enough.

Fixes #1802.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,9 @@
 * Add a REPL option `tcSmtFile` that allows writing typechecker-related SMT
   solver interactions to a file.
 
+* The typechecker can now simplify types of the form `width (2^^n)` to `n + 1`.
+  ([#1802](https://github.com/GaloisInc/cryptol/issues/1802))
+
 # 3.2.0 -- 2024-08-20
 
 ## Language changes
--- a/src/Cryptol/TypeCheck/SimpType.hs
+++ b/src/Cryptol/TypeCheck/SimpType.hs
@@ -320,6 +320,10 @@ tWidth x
   , TCon (TF TCExp) [p,q] <- tNoUser a
   , Just 2 <- tIsNum p = q
 
+  -- width (2^n) = n + 1
+  | TCon (TF TCExp) [m,n] <- tNoUser x
+  , Just 2 <- tIsNum m = tf2 TCAdd n (tNum (1 :: Int))
+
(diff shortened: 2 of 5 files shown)


marsella (approved):
Thank you for addressing so quickly!

PR #1804, opened 2025-02-15
Use smart constructor instead of VSeq

### LINKED ISSUE #1437: Make `GenValue` abstract

danmatichuk:
Resolves #1437. Renames VSeq constructor to
VSeqCtor and replaces it with a view-only pattern.

All uses of VSeq to construct sequences now use
either mkSeq or mkFinSeq smart constructors. These enforce the invariant that a VSeq must not be
a sequence of bits.


src/Cryptol/Eval/Value.hs
```diff
@@ -122,15 +132,22 @@ data GenValue sym
   | VInteger !(SInteger sym)                   -- ^ @ Integer @ or @ Z n @
   | VRational !(SRational sym)                 -- ^ @ Rational @
   | VFloat !(SFloat sym)
-  | VSeq !Integer !(SeqMap sym (GenValue sym)) -- ^ @ [n]a   @
-                                               --   Invariant: VSeq is never a sequence of bits
+  | VSeqCtor !Integer !(SeqMap sym (GenValue sym)) -- ^ @ [n]a   @
+                                                   --   Invariant: VSeq is never a sequence of bits
```
RyanGlScott, line 136: The Haddocks here still refer to `VSeq`, even though the data constructor has been renamed to `VSeqCtor`. We should update the Haddocks to be in sync with the implementation.

Also, it would be worth adding a comment that `VSeqCtor` is intentionally not exported in an effort to preserve the invariant described in the comment, and that all `VSeqCtor` values must be created through a smart constructor such as `mkSeq` or `mkFinSeq`.

src/Cryptol/Eval/Value.hs
```diff
@@ -122,15 +132,22 @@ data GenValue sym
...
   | VPoly CallStack (TValue -> SEval sym (GenValue sym))   -- ^ polymorphic values (kind *)
   | VNumPoly CallStack (Nat' -> SEval sym (GenValue sym))  -- ^ polymorphic values (kind #)
  deriving Generic
 
+pattern VSeq :: Integer -> SeqMap sym (GenValue sym) -> GenValue sym
+pattern VSeq len vals <- (VSeqCtor len vals)
+
+{-# COMPLETE VRecord, VTuple, VEnum, VBit, VInteger,
+             VRational, VFloat, VWord, VStream,
+             VFun, VPoly, VNumPoly, VSeq #-}
```
RyanGlScott, line 149: Worth leaving a comment here that this should include all of `GenValue`'s data constructors _except_ for `VSeqCtor`, which has instead been swapped out for `VSeq`. That way, if we ever need to add another `GenValue` data constructor later, it will be more obvious that we need to update this `COMPLETE` set (and how to do so).

src/Cryptol/Eval/Value.hs
```diff
@@ -122,15 +132,22 @@ data GenValue sym
...
+                                                   --   Invariant: VSeq is never a sequence of bits
   | VWord !(WordValue sym)                     -- ^ @ [n]Bit @
   | VStream !(SeqMap sym (GenValue sym))       -- ^ @ [inf]a @
   | VFun  CallStack (SEval sym (GenValue sym) -> SEval sym (GenValue sym)) -- ^ functions
   | VPoly CallStack (TValue -> SEval sym (GenValue sym))   -- ^ polymorphic values (kind *)
   | VNumPoly CallStack (Nat' -> SEval sym (GenValue sym))  -- ^ polymorphic values (kind #)
  deriving Generic
 
+pattern VSeq :: Integer -> SeqMap sym (GenValue sym) -> GenValue sym
+pattern VSeq len vals <- (VSeqCtor len vals)
```
RyanGlScott, line 145: Redundant parentheses:

```suggestion
pattern VSeq len vals <- VSeqCtor len vals
```

src/Cryptol/Eval/Value.hs
```diff
@@ -22,10 +22,19 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE PatternSynonyms #-}
```
RyanGlScott, line 25: Minor style suggestion: it would be worth alphabetizing this `LANGUAGE` pragma with respect to the other pragmas. (Similarly for all other places where `LANGUAGE PatternSynonyms` is introduced in this patch.)

src/Cryptol/Eval/Value.hs
```diff
@@ -22,10 +22,19 @@
...
-    GenValue(..), ConValue
+    GenValue 
+      (VRecord, VTuple
+      , VEnum, VBit
+      , VInteger, VRational
+      , VFloat, VWord
+      , VStream, VFun
+      , VPoly, VNumPoly)
+  , ConValue
+  , pattern VSeq
```
RyanGlScott, line 37: I'd recommend writing the export list like this:

```suggestion
    GenValue 
      (VRecord, VTuple
      , VEnum, VBit
      , VInteger, VRational
      , VFloat, VWord
      , VStream, VFun
      , VPoly, VNumPoly
      , VSeq) -- pattern synonym
  , ConValue
```

That way, if you write `import Cryptol.Eval.Value (GenValue(..))` elsewhere, then the pattern synonym will automatically be in scope. This will also avoid needing to enable `LANGUAGE PatternSynonyms` in quite so many places.

src/Cryptol/Eval/Value.hs
```diff
@@ -375,9 +392,18 @@ mkSeq :: Backend sym => sym -> Nat' -> TValue -> SeqMap sym (GenValue sym) -> SE
 mkSeq sym len elty vals = case len of
   Nat n
     | isTBit elty -> VWord <$> bitmapWordVal sym n (fromVBit <$> vals)
-    | otherwise   -> pure $ VSeq n vals
+    | otherwise   -> pure $ VSeqCtor n vals
   Inf             -> pure $ VStream vals
 
+mkFinSeq :: Backend sym => sym -> Integer -> [GenValue sym] -> GenValue sym
```
RyanGlScott, line 398: Please add some Haddocks for this function. What does it construct, and where would it be useful to use this function? (I _think_ `mkFinSeq sym n vs` meant to abstract over the pattern of writing `VSeq n (finiteSeqMap sym vs)`, but it would be helpful to say as much.)

src/Cryptol/Eval/Value.hs
```diff
@@ -375,9 +392,18 @@ mkSeq :: Backend sym => sym -> Nat' -> TValue -> SeqMap sym (GenValue sym) -> SE
...
+mkFinSeq sym n xs = 
+  if any isBit xs then
+    panic "mkFinSeq" ["Bit type not supported"]
```
RyanGlScott, line 401: I have mixed feelings about this code, as it means that it is possible to shoot yourself in the foot by accidentally calling this function on a list of `Bit` values. On the other hand, I do see why it exists: `mkSeq` is monadic, and it is convenient to have a pure function in places that do not have a monad available.

Perhaps we should keep this function, but we should rename it to something like `unsafeFinSeq` to reflect the fact that some care is required in order to use this function correctly. Moreover, we could also add comments next to each call site that justify why it is safe to use the function in those contexts.

src/Cryptol/Eval/What4.hs
```diff
@@ -177,6 +177,7 @@ suiteBPrims :: W4.IsSymExprBuilder sym => What4 sym -> Map.Map PrimIdent (Prim (
 suiteBPrims sym = Map.fromList $ [ (suiteBPrim n, v) | (n,v) <- prims ]
  where
  (~>) = (,)
+ vSeq n w = mkSeq sym (Nat n) (TVSeq w TVBit)
```
RyanGlScott, line 180: There are several parts of the code where we redefine a helper function like this one. Perhaps it would be worth giving this code a home in `Cryptol.Eval.Value` and exporting it?

src/Cryptol/Eval/Value.hs
danmatichuk: One idea would be to introduce a 'newtype' that represents non-VBit values and take a list of those. Then push the "unsafe" step into a separate function that converts GenValue into NonVBitValue.

RyanGlScott: > One idea would be to introduce a 'newtype' that represents non-VBit values and take a list of those.

If we go that route, then I'd prefer having a newtype over a list of `GenValue`s and call it `FinSeq`.

I don't have a strong opinion on whether we abstract this with a newtype or not—as long as we justify each use of the unsafe function that constructs a finite sequence, then I'm happy.

src/Cryptol/Eval/What4.hs
danmatichuk: Is there a good way to indicate the role of each Integer parameter? With the definition nearby it's clear what role each plays, but pulling it out into `Cryptol.Eval.Value` it's a bit obscure to see a function like `wordSeq :: Backend sym => sym -> Integer -> Integer -> SEval sym (Genvalue sym)`.

RyanGlScott: I would advise to just add Haddock comments for each `Integer`:

```hs
wordSeq ::
  Backend sym =>
  sym ->
  -- | The length of the sequence.
  Integer ->
  -- | The word size of the element type.
  Integer ->
  ...
```

src/Cryptol/Eval/Value.hs
danmatichuk: I made another pass with a `FinSeq` newtype. Let me know what you think.

danmatichuk: This is an alternative approach using pattern synonyms: https://github.com/GaloisInc/cryptol/tree/dm/issue-1437-alt

src/Cryptol/Eval/Value.hs
```diff
@@ -369,15 +397,55 @@ ilam sym f =
...
+--   if any values are a 'VBit'.
+toFinSeq :: [GenValue sym] -> Maybe (FinSeq sym)
+toFinSeq xs = FinSeq <$> mapM go xs
+  where
+    go x = case x of
+      VBit _ -> Nothing
+      _ -> Just x
+
+-- | Wrap a 'GenValue' list as a 'FinSeq'. Raises a runtime
+--   error if any values are a 'VBit'
```
RyanGlScott, line 415:
```suggestion
--   error if any values are a 'VBit'.
```

src/Cryptol/Eval/Value.hs
```diff
@@ -369,15 +397,55 @@ ilam sym f =
                      Nat i -> f i
                      Inf   -> panic "ilam" [ "Unexpected `inf`" ])
 
+-- | A finite sequence of non-VBit values. Used in 'finSeq' to
+--   safely construct a 'VSeq'.
+newtype FinSeq sym = FinSeq { fromFinSeq :: [GenValue sym] }
+  deriving Generic
```
RyanGlScott, line 403: Please remove this `Generic` instance, as this leaks the implementation details of `FinSeq`. (Since `FinSeq` is exported abstractly without its newtype constructor, I don't think this was intended.)

RyanGlScott (commented):
Thanks! This is looking quite nice.

Could you add an entry to the `CHANGES.md` file that describes the API changes that this introduces?

src/Cryptol/Eval/What4.hs
kquick: Alternatives to just Haddocks to indicate the usage of the arguments are: use a `newtype` wrapper around the `Integer`, or use the `tagged` package (https://hackage.haskell.org/package/tagged).  Both have some (small) impact on clients... which could be considered a good thing.

RyanGlScott: True, although I suspect that there are many other places where we'd want to use such a newtype beyond `wordSeq`.

For now, I'm inclined to keep things simple and use two `Integer`s as arguments. If we encounter bugs later that are caused by confusing these two arguments, then we can consider introducing newtype(s) to distinguish them.

CHANGES.md
```diff
@@ -1,9 +1,15 @@
 # next -- TBA
 
+
```
RyanGlScott, line 3: This newline change is unnecessary. Can you remove it?

RyanGlScott (approved):
LGTM, aside from one very minor thing.

PR #1809, opened 2025-02-19
check if what4 adapter is present before setting timeout

### LINKED ISSUE #1807: `proverTimeout` is broken for What4-based solvers
Attempting to use a non-zero `proverTimeout` with any What4-based SMT solver will cause Cryptol to throw an exception. Example:

```
$ ~/Software/cryptol-3.2.0/bin/cryptol
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻  
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃  
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.2.0 (1bcb75c)

Loading module Cryptol
Cryptol> :set prover=w4-z3
Cryptol> :set proverTimeout=5
Cryptol> :prove \(x : [8]) -> x == x

What4 exception:
Failed to get solver.cvc4.timeout: not found
```

I suspect that the culprit is the following code:

https://github.com/GaloisInc/cryptol/blob/ba0f8d988d25b6db129392b4863bf13df173b196/src/Cryptol/Symbolic/What4.hs#L775-L782

As seen above, we don't want to unconditionally set the timeout configuration settings for _all_ SMT solvers, as each solver has its own settings. Instead, I think we need to do something solver-specific here.

danmatichuk:
fixes #1807

### DIFF
--- a/src/Cryptol/Symbolic/What4.hs
+++ b/src/Cryptol/Symbolic/What4.hs
@@ -145,6 +145,36 @@ doW4Eval sym m =
        W4Error err  -> liftIO (X.throwIO err)
        W4Result p x -> pure (p,x)
 
+-- | Wraps a 'W4.ConfigOption' to provide a consistent interface for setting
+--   solver goal timeouts (see 'configTimeoutMilliSeconds', 'configTimeoutSeconds', 'setConfigTimeout').
+data ConfigTimeout = 
+        ConfigTimeout 
+          (W4.ConfigOption W4.BaseIntegerType)
+          -- | Translate a 'W4.SolverGoalTimeout' into the integer backing this config option.
+          (W4.SolverGoalTimeout -> Integer)
+
+-- | Construct a 'ConfigTimeout' from a 'W4.ConfigOption', where the given
+--   option configures a solver-specific timeout measured in milliseconds.
+configTimeoutMilliSeconds :: W4.ConfigOption W4.BaseIntegerType -> ConfigTimeout
+configTimeoutMilliSeconds opt = ConfigTimeout opt W4.getGoalTimeoutInMilliSeconds
+
+-- | Construct a 'ConfigTimeout' from a 'W4.ConfigOption', where the given
+--   option configures a solver-specific timeout measured in seconds. 
+--   Rounds up to at least 1 second when setting a non-zero timeout 
+--   (see 'W4.getGoalTimeoutInSeconds').
+configTimeoutSeconds :: W4.ConfigOption W4.BaseIntegerType -> ConfigTimeout
+configTimeoutSeconds opt = ConfigTimeout opt W4.getGoalTimeoutInSeconds
+
+setConfigTimeout :: 
+  W4.IsExprBuilder sym =>
+  ConfigTimeout ->
+  W4.SolverGoalTimeout ->
+  sym ->
+  IO ()
+setConfigTimeout (ConfigTimeout opt toOpt) t sym = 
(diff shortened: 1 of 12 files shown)


src/Cryptol/Symbolic/What4.hs
```diff
@@ -766,17 +777,17 @@ varShapeToConcrete evalFn v =
...
-    symCfg sym W4.cvc5Timeout (1000 * s)
-    symCfg sym W4.boolectorTimeout (1000 * s)
-    symCfg sym W4.yicesGoalTimeout s -- N.B. yices takes seconds
+setTimeout :: W4.IsExprBuilder sym => W4ProverConfig -> Integer -> sym -> IO ()
+setTimeout cfg s sym =
+ do symCfg cfg sym "Z3" W4.z3Timeout (1000 * s)
+    symCfg cfg sym "CVC4" W4.cvc4Timeout (1000 * s)
+    symCfg cfg sym "CVC5" W4.cvc5Timeout (1000 * s)
+    symCfg cfg sym "Boolector" W4.boolectorTimeout (1000 * s)
+    symCfg cfg sym "Yices" W4.yicesGoalTimeout s -- N.B. yices takes seconds
```
RyanGlScott, line 792: Instead of having to enumerate all of the solvers and their timeout options, I wonder if we should do the following:

1. Add a new field to [`AnOnlineAdapter`](https://github.com/GaloisInc/cryptol/blob/5e746cc637c490014980c790458a8c1908401fe2/src/Cryptol/Symbolic/What4.hs#L152-L156) (of type `ConfigOption BaseIntegerType`) to represent the timeout option for a particular solver.
2. In `setTimeout`, match on `cfg` and apply the timeout option for each solver that is contained in the `W4ProverConfig`.

This approach has the advantage that we wouldn't need to remember to keep the implementation of this function in sync whenever Cryptol adds support for a new SMT solver. (In fact, I suspect that this function is _already_ out of sync, since this function doesn't mention the recently-added Bitwuzla solver anywhere.)

RyanGlScott:
Are you able to reproduce [this test failure from CI](https://github.com/GaloisInc/cryptol/actions/runs/13419296547/job/37488712641?pr=1809#step:10:190) locally? If so, do you know what might be causing it?

```
      issue_1807.icry: [Failed]
6c6,8
< Solver returned UNKNOWN
---
> 
> What4 exception:
> fd:18: hFlush: resource vanished (Broken pipe)
```

(I do not use macOS, so I am unable to reproduce it myself.)

danmatichuk:
I've seen this before and it's definitely an issue with What4 itself. It's highly nondeterministic from what I can tell. 

I'll see if I can reproduce it on an up to date What4 build and raise a corresponding issue there. I believe it's limited to the cvc solvers, so we can remove those from the test for now.

src/Cryptol/Symbolic/What4.hs
danmatichuk: This makes sense to me, although I think we'll need an auxiliary config option for yices to account for the second/millisecond discrepancy vs. the rest of the solvers.

danmatichuk:
Currently failing as a result of https://github.com/GaloisInc/what4-solvers/issues/58. Should work with an updated build of yices.

RyanGlScott:
I've published a new `what4-solvers` snapshot ([`snapshot-20250227`](https://github.com/GaloisInc/what4-solvers/releases/tag/snapshot-20250227)) that includes a fix for GaloisInc/what4-solvers#58. I've also pushed a commit that makes use of it—let's see what happens.

PR #1811, opened 2025-02-20
Change the naming convention for anonymous modules

### LINKED ISSUE #1810: User namable anonymous modules
Some constructs in the module system introduce implicitly defined modules/interfaces.   At the moment these have pretty long names, which are not valid Cryptol identifiers.  On the one hand, this is good because it avoids collisions with user-defined names.

On the other hand, this is a problem for other tools integrating with Cryptol: for example if we want to reason about a Cryptol specification, we may want to refer to things defined in these implicit modules, but we can't do it, because we have no convenient way to name the anonymous modules.

To solve this we should introduce a different naming scheme for implicit modules introduce by the module system.

yav:
Fixes #1810.  We use shorter and more readable names, which are valid identifiers.  The reference manual documents the naming convention.


docs/RefMan/Modules.rst
```diff
@@ -13,6 +13,9 @@ contain at most one top-level module.
   f : [8]
   f = 10
 
+Module names should avoid using two consecutive underscored, because
```
RyanGlScott, line 16:
```suggestion
Module names should avoid using two consecutive underscores, because
```

src/Cryptol/Parser/ParserUtils.hs
```diff
@@ -1394,18 +1393,18 @@ class MkAnon t where
   mkAnon    :: AnonThing -> t -> t
   toImpName :: t -> ImpName PName
 
-data AnonThing = AnonArg | AnonIfaceMod
+data AnonThing = AnonArg Int Int | AnonIfaceMod
```
RyanGlScott, line 1396: Please leave Haddocks saying what these two `Int`s are.

src/Cryptol/Utils/Ident.hs
```diff
@@ -170,22 +170,24 @@ instance NFData ModName
 
 -- | Change a normal module name to a module name to be used for an
 -- anonnymous argument.
-modNameArg :: ModName -> ModName
-modNameArg (ModName m fl) =
+modNameArg :: Int -> Int -> ModName -> ModName
```
RyanGlScott, line 173: Similarly, leave Haddocks describing what these two `Int`s are.

src/Cryptol/Utils/Ident.hs
```diff
@@ -364,17 +366,17 @@ modParamIdent (Ident x a t) =
 
 -- | Make an anonymous identifier for the module corresponding to
 -- a `where` block in a functor instantiation.
-identAnonArg :: Ident -> Ident
-identAnonArg (Ident b _ txt) = Ident b AnonModArgName txt
+identAnonArg :: Int -> Int -> Ident
```
RyanGlScott, line 369: Similarly, leave Haddocks describing what these two `Int`s are.

src/Cryptol/Utils/Ident.hs
```diff
@@ -364,17 +366,17 @@ modParamIdent (Ident x a t) =
...
 
 -- | Make an anonymous identifier for the interface corresponding to
 -- a `parameter` declaration.
 identAnonIfaceMod :: Ident -> Ident
 identAnonIfaceMod (Ident b _ txt) = Ident b AnonIfaceModName txt
 
 -- | Make an anonymous identifier for an instantiation in an import.
-identAnontInstImport :: Ident -> Ident
-identAnontInstImport (Ident b _ txt) = Ident b AnonInstImport txt
+identAnonInstImport :: Int -> Int -> Ident
```
RyanGlScott, line 378: Similarly, leave Haddocks describing what these two `Int`s are.

src/Cryptol/Utils/Ident.hs
```diff
@@ -58,7 +58,7 @@ module Cryptol.Utils.Ident
   , modParamIdent
```
RyanGlScott, line 58: Is this `modParamIdent` function actually used anywhere? I ask since it creates an `Ident` whose text begins with `"module parameter"`, which seems like the sort of verbose identifier that this patch is working to fix.

src/Cryptol/Utils/Ident.hs
```diff
@@ -394,10 +397,15 @@ instance NFData MaybeAnon
...
-    NormalName       -> txt
-    AnonModArgName   -> "`where` argument of " <> txt
-    AnonIfaceModName -> "`parameter` interface of " <> txt
-    AnonInstImport   -> txt
+    NormalName -> txt
+    AnonModArgName l c
+      | T.null txt -> "where_at__" <> suff l c
+      | otherwise  -> txt <> "__where"
+    AnonIfaceModName    -> txt <> "__parameter"
+    AnonInstImport l c  -> "import_at__" <> suff l c
```
RyanGlScott, line 405: Worth leaving a comment here for future developers to remind them to update the comments in the reference manual if this code is ever changed.

tests/mono-binds/test02.icry.stdout
```diff
@@ -17,6 +17,7 @@ test02::test =
 
 
 
+
```
RyanGlScott, line 20: It's unfortunate that there is so much extra whitespace printed here (and in other test cases) between lines. This is a problem that has always existed, but it's even more noticeable now because...

src/Cryptol/TypeCheck/AST.hs
```diff
@@ -532,10 +532,14 @@ instance PP n => PP (WithNames (ModuleG n)) where
          , vcat (map pp' mDecls)
 
          , vcat (map pp (Map.elems mFunctors))
+
+         , vcat (map ppSig (Map.toList mSignatures))
```
RyanGlScott, line 536: ...calling `vcat` here when there are no `mSignatures` will result in an extra newline being printed. Does `prettyprinter` offer a combinator that will collapse the newlines to avoid this problem?

tests/suiteb/aes-mct-ecb.icry.stdout
```diff
@@ -14,9 +14,4 @@ Q.E.D.
 property aes192_MCT_vectors_decrypt_correct Using exhaustive testing.
 Testing... Passed 1 tests.
 Q.E.D.
-property aes256_MCT_vectors_encrypt_correct Using exhaustive testing.
-Testing... Passed 1 tests.
-Q.E.D.
-property aes256_MCT_vectors_decrypt_correct Using exhaustive testing.
-Testing... Passed 1 tests.
-Q.E.D.
+property aes256_MCT_vectors_encrypt_correct Ctrl-C
```
RyanGlScott, line 17: This is an unusual change. What happened here?

src/Cryptol/Utils/Ident.hs
yav: It appears to be unused in the Cryptol code base, so I'll delete it.

PR #1816, opened 2025-02-27
add special cases to handle setting timeouts for SBV-based solvers

### LINKED ISSUE #1808: `proverTimeout` is broken for various SBV-based solvers
Cryptol correctly propagates the `proverTimeout` setting to some SBV-based solvers, but not all of them. The `sbv-bitwuzla`, `sbv-abc`, and `sbv-z3` solvers appear to do this correctly:

```
Cryptol> :set proverTimeout=1
Cryptol> :set prover=sbv-bitwuzla
Cryptol> :prove \(x : [8]) -> x == x
Q.E.D.
(Total Elapsed Time: 0.004s, using "Bitwuzla")
Cryptol> :set prover=sbv-abc
Cryptol> :prove \(x : [8]) -> x == x
Q.E.D.
(Total Elapsed Time: 0.032s, using "ABC")
Cryptol> :set prover=sbv-z3
Cryptol> :prove \(x : [8]) -> x == x
Q.E.D.
(Total Elapsed Time: 0.022s, using "Z3")
```

On the other hand, the `sbv-cvc4`, `sbv-cvc5`, `sbv-boolector`, and `sbv-yices` solvers all throw an exception when attempting to use the timeout:

```
Cryptol> :set proverTimeout=1
Cryptol> :set prover=sbv-cvc4
Cryptol> :prove \(x : [8]) -> x == x

SBV exception:

*** Data.SBV: Unexpected non-success response from CVC4:
***
***    Sent      : (set-option :timeout 1000)
***    Expected  : success
***    Received  : unsupported
***
***    Stderr    : CVC4 interrupted by SIGTERM.
***    Exit code : ExitFailure (-6)
***    Executable: /home/ryanscott/Software/what4-solvers/cvc4
***    Options   : --lang smt --incremental --interactive --no-interactive-prompt --model-witness-value
***
***    Reason    : Backend solver reports it does not support this option.
***                Check the spelling, and if correct please report this as a
***                bug/feature request with the solver!

Cryptol> :set prover=sbv-cvc5
Cryptol> :prove \(x : [8]) -> x == x

SBV exception:

*** Data.SBV: Unexpected non-success response from CVC5:
***
***    Sent      : (set-option :timeout 1000)
***    Expected  : success
***    Received  : unsupported
***
***    Stderr    : cvc5 interrupted by SIGTERM.
***    Exit code : ExitFailure (-15)
***    Executable: /home/ryanscott/Software/what4-solvers/cvc5
***    Options   : --lang smt --incremental --nl-cov
***
***    Reason    : Backend solver reports it does not support this option.
***                Check the spelling, and if correct please report this as a
***                bug/feature request with the solver!

Cryptol> :set prover=sbv-boolector
Cryptol> :prove \(x : [8]) -> x == x

SBV exception:

*** Data.SBV: fd:19: hGetLine: end of file:
***
***    Sent      : (set-option :timeout 1000)
***
***    Executable: /home/ryanscott/Software/what4-solvers/boolector
***    Options   : --smt2 -m --output-format=smt2 --no-exit-codes --incremental
***
***    Hint      : Solver process prematurely ended communication.
***                It is likely it was terminated because of a seg-fault.
***                Run with 'transcript=Just "bad.smt2"' option, and feed
***                the generated "bad.smt2" file directly to the solver
***                outside of SBV for further information.

Cryptol> :set prover=sbv-yices
Cryptol> :prove \(x : [8]) -> x == x

SBV exception:

*** Data.SBV: Unexpected non-success response from Yices:
***
***    Sent      : (set-option :timeout 1000)
***    Expected  : success
***    Received  : unsupported
***
***    Exit code : ExitFailure (-15)
***    Executable: /home/ryanscott/Software/what4-solvers/yices-smt2
***    Options   : --incremental
***
***    Reason    : Backend solver reports it does not support this option.
***                Check the spelling, and if correct please report this as a
***                bug/feature request with the solver!
```

The problem is that the code in Cryptol attempts to propagate `proverTimeout` information to all SBV-based solvers by sending the `(set-option :timeout <NUM>)` command:

https://github.com/GaloisInc/cryptol/blob/ba0f8d988d25b6db129392b4863bf13df173b196/src/Cryptol/Symbolic/SBV.hs#L110-L119

But as seen above, not all solvers understand `(set-option :timeout <NUM>)`. For instance, `cvc5` configures a timeout by way of its `--tlimit=<MILLISECONDS>` command-line option.

We should figure out if SBV has a better way to configure timeout information that does the right thing for all solvers and do that instead.

danmatichuk:
fixes #1808

* ~Yices still does not work due to unexpected behavior after a timeout is triggered (see: https://github.com/LeventErkok/sbv/issues/735)~ Workaround using deadman timer

* the special cases for CVC4/5 are addressed in the 11.1.5 release for SBV (which is not compatible with cryptol as it requires GHC 9.8)

### DIFF
--- a/src/Cryptol/Symbolic/SBV.hs
+++ b/src/Cryptol/Symbolic/SBV.hs
@@ -110,11 +110,36 @@ proverConfigs =
 setTimeoutSecs ::
   Int {- ^ seconds -} ->
   SBV.SMTConfig -> SBV.SMTConfig
-setTimeoutSecs s cfg = cfg
-  { SBV.solverSetOptions =
-    SBV.OptionKeyword ":timeout" [show (toInteger s * 1000)] :
-    filter (not . isTimeout) (SBV.solverSetOptions cfg) }
+setTimeoutSecs s cfg = case SBV.name (SBV.solver cfg) of
+  SBV.Yices -> cfg'
+  -- FIXME: although this correctly sets the timeout for Yices,
+  -- SBV crashes due to an unexpected response from Yices after
+  -- a timeout is hit.
+  -- see:
+  --   https://github.com/LeventErkok/sbv/issues/735
+  --   https://github.com/SRI-CSL/yices2/issues/547
+    { SBV.extraArgs =
+      ["--timeout", show (toInteger s)] ++
+      SBV.extraArgs cfg' }
+  -- NOTE: in sbv 11.1.5 these special cases are handled
+  -- by a new 'SetTimeOut' constructor for 'SMTOption'
+  SBV.CVC4 -> cfg'
+    { SBV.extraArgs =
+      ["--tlimit-per", show (toInteger s * 1000)] ++
+      SBV.extraArgs cfg' }
+  SBV.CVC5 -> cfg'
+    { SBV.extraArgs =
+      ["--tlimit-per", show (toInteger s * 1000)] ++
+      SBV.extraArgs cfg' }
+  _ -> cfg'
+    { SBV.solverSetOptions =
+      SBV.OptionKeyword ":timeout" [show (toInteger s * 1000)] :
+      (SBV.solverSetOptions cfg') }
   where
+    cfg' = cfg
+      { SBV.solverSetOptions =
+        filter (not . isTimeout) (SBV.solverSetOptions cfg) }
+
     isTimeout (SBV.OptionKeyword k _) = k == ":timeout"
     isTimeout _ = False
 
--- a/src/Cryptol/Symbolic/SBV.hs
+++ b/src/Cryptol/Symbolic/SBV.hs
@@ -29,6 +29,7 @@ module Cryptol.Symbolic.SBV
 
 
 import Control.Applicative
+import Control.Concurrent (threadDelay)
 import Control.Concurrent.Async
 import Control.Concurrent.MVar
 import Control.Monad.IO.Class
@@ -42,7 +43,7 @@ import qualified Data.Vector as Vector
 
 import LibBF(bfNaN)
 
-import qualified Data.SBV as SBV (sObserve, symbolicEnv)
+import qualified Data.SBV as SBV (sObserve, symbolicEnv, SMTReasonUnknown (..))
 import qualified Data.SBV.Internals as SBV (SBV(..))
 import qualified Data.SBV.Dynamic as SBV
 import qualified Data.SBV.Trans.Control as SBV (SMTOption(..))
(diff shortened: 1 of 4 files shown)


RyanGlScott:
> the special cases for CVC4/5 are addressed in the 11.1.5 release for SBV (which is not compatible with cryptol as it requires GHC 9.8)

Interesting. As a stopgap measure, would it be worth allowing Cryptol to build against `sbv-11.1.5`, but not requiring it as the minimum? That way, folks who are able to use GHC 9.8 or later will be able to benefit from the latest changes in SBV, but without requiring us to drop support for pre-9.8 GHCs quite yet.

src/Cryptol/Symbolic/SBV.hs
```diff
@@ -110,11 +111,39 @@ proverConfigs =
...
+  -- and instead rely on addDeadmanTimer to kill Yices when the
+  -- timeout is hit.
+  -- see:
+  --   https://github.com/LeventErkok/sbv/issues/735
+  --   https://github.com/SRI-CSL/yices2/issues/547
+    {- { SBV.extraArgs =
+      ["--timeout", show (toInteger s)] ++
+      SBV.extraArgs cfg' } -}
+  -- NOTE: in sbv 11.1.5 these special cases are handled
+  -- by a new 'SetTimeOut' constructor for 'SMTOption'
```
RyanGlScott, line 129: Is this comment meant to apply to the Yices timeout above, or the CVC{4,5} timeouts below? If it's the latter, it would be worth adding a newline between this comment and the Yices-specific comments above to make this more obvious.

Also, it would be worth leaving a comment saying that once Cryptol requires `sbv-11.1.5` as the minimum, then we can drop the CVC{4,5}-related special cases below.

src/Cryptol/Symbolic/SBV.hs
```diff
@@ -250,6 +279,25 @@ runMultiProvers pc lPutStrLn provers callSolver processResult e = do
               do forM_ others (\a -> X.throwTo (asyncThreadId a) ExitSuccess)
                  return r
 
+-- | Wrap a solver call with a given timeout. If the timeout is reached, then
+--   the default result is returned and the solver call is terminated.
+addDeadmanTimer ::
+  Int {- ^ timeout in seconds -} ->
+  (SBV.SMTConfig -> res) {- ^ result to give when a timeout is hit -} ->
+  (SBV.SMTConfig -> SBV.Symbolic SBV.SVal -> IO res) ->
```
RyanGlScott, line 287: Add a comment like:

```suggestion
  (SBV.SMTConfig -> SBV.Symbolic SBV.SVal -> IO res) { ^ call the SMT solver -} ->
```

src/Cryptol/Symbolic/SBV.hs
```diff
@@ -250,6 +279,25 @@ runMultiProvers pc lPutStrLn provers callSolver processResult e = do
               do forM_ others (\a -> X.throwTo (asyncThreadId a) ExitSuccess)
                  return r
 
+-- | Wrap a solver call with a given timeout. If the timeout is reached, then
+--   the default result is returned and the solver call is terminated.
```
RyanGlScott, line 283: As I understand it, this code is only necessary to work around an issue in the way SBV interacts with Yices. Is that correct? If so, worth stating as much in the Haddocks.

RyanGlScott (commented):
Please mention this in the changelog!

src/Cryptol/Symbolic/SBV.hs
danmatichuk: It is, but at the moment it's actually wrapping all solver calls regardless, it just happens to be strictly necessary with Yices.

In general there's no harm doing this, it's just an additional measure to ensure that the timeout is respected even if the solver misbehaves somehow.

I can note all of this in the Haddocks.

RyanGlScott:
> Regardless I think this is probably outside the scope of this specific issue for the moment, unless we want to put in some preprocessor flags to conditionally use the new sbv API.

Ah, I wasn't aware that this required using a different API. In that case, no need to do this as part of this PR.

RyanGlScott (approved):

PR #1819, opened 2025-02-28
improve exhaustiveness check by attempting each guard as the consequent

### LINKED ISSUE #1796: Incomplete reasoning about exponentiation with respect to exhaustiveness checking
In certain cases, Cryptol's exhaustiveness checker for numeric constraint guards incorrectly deems sets of guards to be inexhaustive when they involve exponentiation. Here is a program with some relevant examples:

```
module Bug where

exhaustive1 : {w} [2 ^^ (2 ^^ w)] -> [2 ^^ (2 ^^ w)]
exhaustive1 x
  | (2 ^^ (2 ^^ w)) < 2 => x
  | (2 ^^ (2 ^^ w)) == 2 => ~ x
  | (2 ^^ (2 ^^ w)) > 2 => x

// This should be exhaustive, as it simply reorders the guards found in the
// `exhaustive1` function, but Cryptol deems it non-exhaustive.
nonExhaustive1 : {w} [2 ^^ (2 ^^ w)] -> [2 ^^ (2 ^^ w)]
nonExhaustive1 x
  | (2 ^^ (2 ^^ w)) == 2 => ~ x
  | (2 ^^ (2 ^^ w)) < 2 => x
  | (2 ^^ (2 ^^ w)) > 2 => x

// This should be exhaustive, as it simply reorders the guards found in the
// `exhaustive1` function, but Cryptol deems it non-exhaustive.
nonExhaustive2 : {w} [2 ^^ (2 ^^ w)] -> [2 ^^ (2 ^^ w)]
nonExhaustive2 x
  | (2 ^^ (2 ^^ w)) == 2 => ~ x
  | (2 ^^ (2 ^^ w)) > 2 => x
  | (2 ^^ (2 ^^ w)) < 2 => x

// Note that `(2 ^^ w) >= 1`, which means that there is no need for a
// `(2 ^^ w) < 1` guard below.
exhaustive2 : {w} [2 ^^ w] -> [2 ^^ w]
exhaustive2 x
  | (2 ^^ w) == 1 => ~ x
  | (2 ^^ w) > 1 => x

// This should be exhaustive, as it is a slight variation on the same idea used
// in `exhaustive2`—namely, that `(2 ^^ (2 ^^ w)) >= 2`. Despite this, Cryptol
// deems this function non-exhaustive.
nonExhaustive3 : {w} [2 ^^ (2 ^^ w)] -> [2 ^^ (2 ^^ w)]
nonExhaustive3 x
  | (2 ^^ (2 ^^ w)) == 2 => ~ x
  | (2 ^^ (2 ^^ w)) > 2 => x
```

I would expect all of the functions above to be marked as exhaustive, but Cryptol does not do so for the functions prefixed with `nonExhaustive*`:

```
[warning] at Bug.cry:36:1--36:15:
  Could not prove that the constraint guards used in defining Bug::nonExhaustive3 were exhaustive.
[warning] at Bug.cry:20:1--20:15:
  Could not prove that the constraint guards used in defining Bug::nonExhaustive2 were exhaustive.
[warning] at Bug.cry:12:1--12:15:
  Could not prove that the constraint guards used in defining Bug::nonExhaustive1 were exhaustive.
```

My understanding is that the typechecker's reasoning about the behavior of exponentiation is necessarily incomplete, but it does feel like we could do better in these specific examples.

danmatichuk:
fixes #1796 

this comes at the cost of increasing the time to
fail if the guards are truly non-exhaustive

### DIFF
--- a/src/Cryptol/TypeCheck/Infer.hs
+++ b/src/Cryptol/TypeCheck/Infer.hs
@@ -1340,19 +1340,22 @@ checkExhaustive name as asmps guards =
     (_, []) -> Nothing
     (ys,x:xs') -> Just (x, ys ++ xs')
 
-  -- if starting with the longest guard fails, re-try in descending order
+  -- if starting with the longest guard fails with a 'ProofUnknown' result,
+  -- then re-try with the next guard in descending order.
+  -- NB: in the worst case this is quadratic in the number of guard predicates, but
+  -- in practice we expect this number to be low
   go [] _ = pure False -- XXX: we should check the asmps are unsatisfiable
   go goals i = case pluck i goals of
     Just (goalp, rest) ->
       do ok <- doGoals (theAlts rest) (map toGoal goalp)
          case ok of
-           Just True -> pure True
-           Just False -> pure False
-           Nothing -> go goals (i+1)
+           ProofSuccess -> pure True
+           ProofFail -> pure False
+           ProofUnknown -> go goals (i+1)
     Nothing -> pure False
 
   cmpByLonger props1 props2 = compare (length props2) (length props1)
-                                          -- reversed, so that longets is first
+                                          -- reversed, so that longest is first
 
   theAlts :: [[Prop]] -> [[Prop]]
   theAlts = map concat . sequence . map chooseNeg
@@ -1368,13 +1371,13 @@ checkExhaustive name as asmps guards =
   -- Try to validate all cases
   doGoals todo gs =
     case todo of
-      []     -> pure $ Just True
+      []     -> pure ProofSuccess
       alt : more ->
         do ok <- canProve (asmps ++ alt) gs
            case ok of
-             Just True -> doGoals more gs
-             Just False -> pure $ Just False
-             Nothing -> pure Nothing
+             ProofSuccess -> doGoals more gs
+             ProofFail -> pure ProofFail
+             ProofUnknown -> pure ProofUnknown
 
   toGoal :: Prop -> Goal
   toGoal prop =
@@ -1390,13 +1393,22 @@ checkExhaustive name as asmps guards =
     UnsolvedGoals{} -> True
     _ -> False
 
-  canProve :: [Prop] -> [Goal] -> InferM (Maybe Bool)
+  canProve :: [Prop] -> [Goal] -> InferM ProofResult
   canProve asmps' goals =
     do res <- tryProveImplication (Just (thing name)) as asmps' goals
        case res of
-         Left errs | all maybeSolvable errs -> return Nothing
-         Left{} -> return $ Just False
-         Right{} -> return $ Just True
+         Left errs | all maybeSolvable errs -> return ProofUnknown
+         Left{} -> return ProofFail
+         Right{} -> return ProofSuccess
+
+data ProofResult = 
+    ProofSuccess
+    -- ^ Proof attempt was successful.
+  | ProofFail
+    -- ^ Proof attempt failed, and goal is most likely not provable.
+  | ProofUnknown
+    -- ^ Proof attempt failed due to incomplete heuristics, goal may
+    -- still be provable.
 
 {- | Generate type-checked syntax for the code in a PropGuard. For example,
 consider the following (pre–type-checked) syntax for a guard:
--- a/src/Cryptol/TypeCheck/Infer.hs
+++ b/src/Cryptol/TypeCheck/Infer.hs
@@ -1320,7 +1320,17 @@ The implications were derive by the following general algorithm:
   we need to consider a branch for each disjunct --- one branch gets the
   assumption @~B1@ and another branch gets the assumption @~B2@. Each
   branch's implications need to be proven independently.
-
+- If the solver fails to prove any subgoal, but with an error indicating
+  the subgoal may be provable (i.e. the attempt was terminated due to some partial
+  heuristic), then pick the next-longest guard as the RHS and re-start. e.g.:
+  @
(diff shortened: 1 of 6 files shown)


RyanGlScott:
Can you elaborate a bit further on how your approach works? It's not obvious to me at a glance why checking the guards in a different order is desirable (or even sound).

danmatichuk:
> Can you elaborate a bit further on how your approach works? It's not obvious to me at a glance why checking the guards in a different order is desirable (or even sound).

The algorithm for checking exhaustiveness involves picking one guard as the consequent, and then proving that negating all of the other guards imply the consequent.

i.e for:
```
f | (B1, B2) => ...
  | (C1, C2, C3) => ...
```

The overall goal is to prove that the predicates cover all possible cases, e.g.:
prove that `(B1 /\ B2) \/ (C1 /\ C2 /\ C3)` always holds.

We can rewrite this into either:
* `~(B1 /\ B2) => (C1 /\ C2 /\ C3)`
* `~(C1 /\ C2 /\ C3) => (B1 /\ B2)`

These can be proven by then expanding into a collection of implications that can be solved independently:

Either:
```
~B1 => C1 /\ C2 /\ C3
~B2 => C1 /\ C2 /\ C3
```
or
```
~C1 => B1 /\ B2
~C2 => B1 /\ B2
~C3 => B1 /\ B2
```

The choice of which set of goals to solve is somewhat arbitrary, but currently the "longest" guard (most individual predicates) is chosen in order to reduce the overall number of goals that need to be solved. However, because the solver used is heuristic here, the choice of which guard to use as the consequent can affect whether or not the proof succeeds.

danmatichuk:
I've modified the description of the algorithm in the comment to include this change now.

RyanGlScott:
Ah, I missed the description of the algorithm (that you have since amended)—my bad!

If I understand you correctly, the reason that Cryptol was deeming the examples from https://github.com/GaloisInc/cryptol/issues/1796 to be non-exhaustive was because:

* The algorithm just so happened (by an unlucky fluke of the previous heuristics) to solve for a guard for which the answer was "unknown", causing the algorithm to conservatively give up.
* If the algorithm had instead tried processing the guards in a different order, then it would have made further progress.

Is that right?

danmatichuk:
> Ah, I missed the description of the algorithm (that you have since amended)—my bad!
> 
> If I understand you correctly, the reason that Cryptol was deeming the examples from #1796 to be non-exhaustive was because:
> 
> * The algorithm just so happened (by an unlucky fluke of the previous heuristics) to solve for a guard for which the answer was "unknown", causing the algorithm to conservatively give up.
> * If the algorithm had instead tried processing the guards in a different order, then it would have made further progress.
> 
> Is that right?

Yes, more precisely it only matters which guard is picked as the "longest" to serve as the RHS, where "longest" is just the number of individual predicates (separated by commas, not conjuncts, afaik). Since the guards in those examples are all the same length by this measure, it ends up just using whichever is first in the source code.

So the change doesn't try all permutations of guards, it just tries picking a different guard for "longest" if the solver gives an "unknown" result.

src/Cryptol/TypeCheck/Solve.hs
```diff
@@ -290,10 +290,10 @@ proveImplication dedupErrs lnam as ps gs =
 -- | Tries to prove an implication. If proved, then returns `Right (m_su ::
 -- InferM Subst)` where `m_su` is an `InferM` computation that results in the
```
RyanGlScott, line 291: This claims that it returns `Right (m_su :: InferM Subst)`, which doesn't seem right. I wonder if this is a leftover from a previous implementation. I'd favor removing the mention of the substitution from the Haddocks and instead changing it to something like "return `Right warnings`, where `warnings` records all the warnings invoked during proving".

src/Cryptol/TypeCheck/Solve.hs
```diff
@@ -290,10 +290,10 @@ proveImplication dedupErrs lnam as ps gs =
...
 -- solution substitution, and records any warning invoked during proving. If not
--- proved, then returns `Left (m_err :: InferM ())`, which records all errors
+-- proved, then returns `Left (m_err :: [Error])`, which records all errors
 -- invoked during proving.
```
RyanGlScott, line 294: Perhaps just:

```suggestion
-- proved, then returns `Left errs`, where `errs` records all errors
-- invoked during proving.
```

(Mentioning the `[Error]` type is a bit redundant, given that it's already mentioned in the type signature.)

src/Cryptol/TypeCheck/Infer.hs
```diff
@@ -1368,13 +1371,13 @@ checkExhaustive name as asmps guards =
...
+      []     -> pure ProofSuccess
       alt : more ->
         do ok <- canProve (asmps ++ alt) gs
            case ok of
-             Just True -> doGoals more gs
-             Just False -> pure $ Just False
-             Nothing -> pure Nothing
+             ProofSuccess -> doGoals more gs
+             ProofFail -> pure ProofFail
+             ProofUnknown -> pure ProofFail
```
RyanGlScott, line 1380: This changed a `Nothing` value to `ProofFail`. Shouldn't this be `ProofUnknown`?

danmatichuk: You're right, fixed.

RyanGlScott: My hope is that fixing this will also fix the test suite failures seen in the most recent CI pipeline. I'll approve this under the assumption that that is actually the case.

RyanGlScott (approved):

Issue #1744, opened 2024-09-04
Some position handling issues in the typechecker


sauclovian-g:
The following file produces five type errors. The real problem is that the ```Key``` argument of ```once``` is commented out. (When I originally hit this, it was just missing entirely, by accident.)

It would probably be nice if the typechecker had an explicit message for "your function has too many parameters for its type signature", which it doesn't. That's what's really going on.

However, the curious bit is that the first output error is reported with a position starting at line 1, column 1. The ending position, however, appears to be the intended location. It's odd that the position can get spliced like this, and suggests something maybe worth looking into.

I assume this error is getting printed first because the errors get sorted by starting line, so the fact that it's first (which is distracting for the user, since normally if there's a pile of errors the first one is the real issue and the rest are fallout) is a side effect and not its own problem.

Also though, the position of the second error, which is reporting that the return type of ```once``` is wrong and reflects the real issue, is reported as the entirety of the function rather than, say, the return value. This is not ideal, especially in a large function.

What's here is cut down from a much larger example so it doesn't mean anything. It's a fold over a list of pairs (has to be a list of tuples to tickle the problem, I think) that updates a newtype.

Bug.cry:
```
module Bug where

import Array

type Widget = [3]
type Wombat = [4]
type Key = [5]
newtype Info = {
   stuff: Array [5] [6]
}

once: Info -> /*Key ->*/ Widget -> Wombat -> Info
once info key widget wombat = info'
   where
      info' = Info {
         stuff = arrayUpdate info.stuff key 0
      }

iter: {n} n < 10 => Info -> Key -> [n] (Widget, Wombat) -> Info
iter info key entries = infos ! 0
   where
      infos = [info] # [once info' key widget wombat
                           | info' <- infos
                           | (widget, wombat) <- entries]

process: {n} n < 10 => Info -> [n] (Widget, Wombat) -> Info
process info entries = iter info 0 entries
```

Output (from current head, which is also 3.2.0):
```
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻  
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃  
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.2.0.99 (0985325)
https://cryptol.net  :? for help

Loading module Cryptol
Loading module Array
Loading module Bug
[error] at Bug.cry:1:1--24:37:
  Type mismatch:
    Expected type: 3
[...]

PR #1821, opened 2025-03-04
add location to selectors when doing a 'noPat' pass

danmatichuk:
partial fix for #1744

### DIFF
--- a/src/Cryptol/Parser/NoPat.hs
+++ b/src/Cryptol/Parser/NoPat.hs
@@ -68,7 +68,7 @@ simpleBind x e = Bind { bName = x, bParams = []
 
 sel :: Pattern PName -> PName -> Selector -> Bind PName
 sel p x s = let (a,ts) = splitSimpleP p
-            in simpleBind a (foldl ETyped (ESel (EVar x) s) ts)
+            in simpleBind a (at p (foldl ETyped (ESel (EVar x) s) ts))
 
 -- | Given a pattern, transform it into a simple pattern and a set of bindings.
 -- Simple patterns may only contain variables and type annotations.


sauclovian-g (approved):
All seems plausible, but I've never looked at most of this code before so take it for what it's worth...

src/Cryptol/ModuleSystem/Renamer.hs
```diff
@@ -1073,12 +1073,15 @@ instance Rename Bind where
...
          do mbSig <- traverse renameSchema (bSignature b)
             shadowNames (fst `fmap` mbSig) $
-              do (patEnv,pats') <- renamePats (bParams b)
+              do (patEnv,pats') <- renamePats (bindParams b)
                  -- NOTE: renamePats will generate warnings,
                  -- so we don't need to trigger them again here.
                  e' <- shadowNames' CheckNone patEnv (rnLocated rename (bDef b))
+                 let bParams' = case bParams b of
+                       PatternParams _ -> PatternParams pats'
+                       DroppedParams rng i -> DroppedParams rng i
```
RyanGlScott, line 1082: I'd consider introducing a `Rename` instance for `BindParams`, as this would allow you to encapsulate the logic of renaming them in that instance.

src/Cryptol/Parser/AST.hs
```diff
@@ -495,6 +496,45 @@ data Bind name = Bind
...
+bindHeaderLoc :: Bind name -> Maybe Range
+bindHeaderLoc b = getLoc (bName b, (bSignature b, bParams b))
+
+-- | An empty 'BindParams' (i.e. zero parameters).
+--   Note that 'dropParams' should be used instead of this
+--   when rewriting an existing 'Bind' to have no parameters.
+noParams :: BindParams name
+noParams = PatternParams []
+
+data BindParams name =
```
RyanGlScott, line 525: If I understand, this data type really only exists for the sake of improving error messages. Is that right? If so, it would be worth stating that in a comment.

src/Cryptol/TypeCheck/Error.hs
```diff
@@ -109,6 +111,10 @@ data Error    = KindMismatch (Maybe TypeSource) Kind Kind
               | RecursiveTypeDecls [Name]
                 -- ^ The type synonym declarations are recursive
 
+              | TooManyParams Name Type Int Int
+                -- ^ A binder implies more arguments than
+                --   its type signature
```
RyanGlScott, line 116: Please leave a comment saying what these fields represent. In particular, it's not obvious which `Int` is the expected number of arguments, and which `Int` is the actual number of arguments.

src/Cryptol/ModuleSystem/Renamer.hs
danmatichuk: I think this isn't completely straightforward, because we need to also pull out the new naming environment. I can pull this out into a separate function `renameBindParams`, and put it next to the similar functions `renamePats` etc.

RyanGlScott (approved):

PR #1827, opened 2025-03-20
Bump parameterized-utils to new release 2.1.10.0

sauclovian-g:


cabal.GHC-9.4.8.config
```diff
@@ -201,7 +201,7 @@ constraints: any.BoundedChan ==1.0.3.0,
              any.os-string ==2.0.7,
              any.panic ==0.4.0.1,
              any.parallel ==3.2.2.0,
-             any.parameterized-utils ==2.1.9.0,
+             any.parameterized-utils ==2.1.10.0,
```
RyanGlScott, line 204: I suspect that you will also need to bump the `what4` version after it is released to Hackage.

sauclovian-g:
Ugh, do I need to rev the cache to get it to find the new release upstream? If so maybe this should wait until there's a what4 release as well...

sauclovian-g:
I put back the temporary reference because we'll want it for test-building the what4 candidate. And I cancelled all the resulting CI runs because they're a waste of cycles

RyanGlScott:
> Ugh, do I need to rev the cache to get it to find the new release upstream?

At the very least, you'll need to bump the `index-state`s in the `cabal.GHC-*.config` files, as their dates are too old to contain the soon-to-be-uploaded-to-Hackage version of `parameterized-utils` that you want.

sauclovian-g:
It seems publishing the release candidate made it go away, so that link doesn't work any more.

RyanGlScott:
Indeed, I think you'd want https://hackage.haskell.org/package/parameterized-utils-2.1.10.0/parameterized-utils-2.1.10.0.tar.gz instead (whose URL no longer contains "`candidate`")

sauclovian-g:
Humph. Now it fails injecting the download output even with -v0. I guess it is just not going to work with the download.

sauclovian-g:
ok, it finally succeeded, I'm going to clean up the mess

sauclovian-g:
For the record, what's actually in this branch is:
- fix some missing cabal -v0 when testing the exercises
- bump parameterized-utils and what4 in the cabal freeze files, and also the index date
- adjust the version constraints on what4 to accept the new version

Temporarily it also included references to Hackage candidate uploads for the new releases, which then blew up because of the missing cabal -v0 (and then also either more cases I haven't found, or a cabal bug where downloading these produces extra output anyway) because the output from those extra downloads leaked into test runs. All of the candidate references have been removed again.

But we should be ready to go now. Finally.

PR #1830, opened 2025-03-26
CI: Support Ubuntu 24.04, drop 20.04 support

RyanGlScott:
GitHub Actions is removing its Ubuntu 20.04 runners soon (https://github.com/actions/runner-images/issues/11101), so this patch removes CI support for Ubuntu 20.04 and makes Ubuntu 24.04 the latest supported Ubuntu version.

### DIFF
--- a/cryptol-remote-api/Dockerfile
+++ b/cryptol-remote-api/Dockerfile
@@ -11,7 +11,7 @@ ARG GHCVER_BOOTSTRAP="9.0.2"
 # may need to update ALEXVER and HAPPYVER as well.
 ARG ALEXVER="3.4.0.1"
 ARG HAPPYVER="1.20.1.1"
-FROM ubuntu:22.04 AS toolchain
+FROM ubuntu:24.04 AS toolchain
 ARG PORTABILITY=false
 RUN apt-get update && \
     apt-get install -y \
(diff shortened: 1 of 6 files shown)


RyanGlScott:
Here is a manually triggered job that demonstrates that the changes to the Dockerfile work as intended: https://github.com/GaloisInc/cryptol/actions/runs/14090326568

cryptol-remote-api/Dockerfile
```diff
@@ -59,7 +59,7 @@ RUN if ${PORTABILITY}; then \
 
 FROM toolchain AS build
 
-RUN useradd -u 1000 -m cryptol
+RUN useradd -m cryptol
```
sauclovian-g, line 62: Did you mean to include this change in here? (seems unrelated, is all)

sauclovian-g (approved):
LGTM

cryptol-remote-api/Dockerfile
RyanGlScott: This change is intentional, and is in fact required in order to make the Docker image build with `ubuntu:24.04` as the base image. See the commit message in 2fdef73839a4cef5c28ea691c6fbefa456f0514f.

sauclovian-g: Just making sure, sorry about the noise.

Issue #1836, opened 2025-04-02
Generalize Docker images to work on both x86-64 and ARM64


RyanGlScott:
Currently, the `cryptol` and `cryptol-remote-api` Docker images are only available for x86-64 (specifically, `linux/amd64`). Given the popularity of ARM64 machines these days, we would also like to support ARM64. This will involve two subtasks:

* [ ] Generalize the logic in the Dockerfiles to support downloading x86-64 or ARM64 binaries whenever appropriate.
* [ ] Generalize the CI to deploy both x86-64 and ARM64 Docker images.

PR #1838, opened 2025-04-03
Dockerfiles: Support both amd64 and arm64 builds

### LINKED ISSUE #1837: `cryptol-remote-api/Dockerfile`: Allow building with `podman`
Currently, `podman` fails to build the `cryptol-remote-api` `Dockerfile` due to these lines:

https://github.com/GaloisInc/cryptol/blob/fe250a589c3a7f3d41c0893fe344694e1a7d81d9/cryptol-remote-api/Dockerfile#L60-L70

This is because the `build` stage references the `GHCVER` argument without first redeclaring `ARG GHCVER` after the `FROM` line. `docker buildx` will inherit the value of `GHCVER` from the `toolchain` stage, but `podman` is stricter and will not do this, causing `GHCVER` to expand to the empty string in the `COPY cabal.GHC-${GHCVER}.config cabal.project.freeze`. The fix is simple: just add an `ARG GHCVER` line after the `FROM toolchain AS build` line.

There is a similar issue related to the `PORTABILITY` `ARG` here:

https://github.com/GaloisInc/cryptol/blob/fe250a589c3a7f3d41c0893fe344694e1a7d81d9/cryptol-remote-api/Dockerfile#L72

RyanGlScott:
Ensure that the `what4-solvers` and `ghcup` binaries correspond to the appropriate architecture when downloading them. Addresses one half of https://github.com/GaloisInc/cryptol/issues/1836.

While testing this with `podman` on Apple silicon, I discovered #1837, so I also included a fix for that in a separate commit. As such, this PR also fixes #1837.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,19 @@ USER cryptol
 WORKDIR /cryptol
 RUN mkdir -p rootfs/usr/local/bin
 WORKDIR /cryptol/rootfs/usr/local/bin
+ARG TARGETPLATFORM
 # The URL here is based on the same logic used to specify BIN_ZIP_FILE in
-# `.github/workflow/ci.yml`, but specialized to x86-64 Ubuntu.
-RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20250326/ubuntu-24.04-X64-bin.zip"
+# `.github/workflow/ci.yml`, but specialized to Ubuntu.
+RUN case ${TARGETPLATFORM} in \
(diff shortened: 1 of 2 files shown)


sauclovian-g (approved):
Seems plausible, don't remember how `ARG` works though fwiw (if I ever knew)

RyanGlScott:
Manually triggered CI job demonstrating that the Docker builds still work on x86-64 after these changes: https://github.com/GaloisInc/cryptol/actions/runs/14263039121

(I verified locally that the ARM64 builds work.)

PR #1844, opened 2025-04-15
Add :saw notebook command

### LINKED ISSUE #1835: Add `:saw` command to REPL and docstrings
This command should take a file argument, which should be a SAWScript file, and run it with `saw`, returning success or failure.  We probably also need some settings to configure how to find `saw`.

The command should work from docstrings, the intended use case is that when in additional to running tests and proving things with Cryptol, we can also invoke a SAW script to presumably verify something about the program.

peterohanley-galois:
The command succeeds when `saw` returns 0. Saw output is suppressed because `saw` adds timestamps and absolute paths to all output.

Saw is launched directly, so there is no concern about shell commands escaping the notebook.

Fixes #1835

### DIFF
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -1531,8 +1531,8 @@ helpCmd cmd
 runShellCmd :: String -> REPL CommandResult
 runShellCmd cmd
   = io $ do h <- Process.runCommand cmd
-            _ <- waitForProcess h
-            return emptyCommandResult -- This could check for exit code 0
+            e <- waitForProcess h
+            pure emptyCommandResult { crSuccess = e == ExitSuccess }
 
 cdCmd :: FilePath -> REPL CommandResult
 cdCmd f | null f = do rPutStrLn $ "[error] :cd requires a path argument"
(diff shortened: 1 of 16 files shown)


peterohanley-galois:
The tests are failing because `saw` isn't in the path, but it's not built as part of CI for good reason.

yav:
@peterohanley-galois we can't really have the CI start failing if we merge this, but I agree that we don't want to depend on `saw`.  Could you make the tests run some other script, pretending to be `saw`?

src/Cryptol/REPL/Command.hs
```diff
@@ -318,6 +318,9 @@ nbCommandList  =
   , CommandDescr [ ":check-docstrings" ] [] (ModNameArg checkDocStringsCmd)
       "Run the REPL code blocks in the module's docstring comments"
       ""
+  , CommandDescr [ ":saw" ] [] (FilenameArg sawCmd)
+    "Load a given SAW file"
```
yav, line 322: Maybe mention `CRYPTOL_SAW` here

yav (approved):
Overall looks good, thanks!   Could you do these fixes before merging:
  * do something the command line flags (see comment below, please disregard this if I am confused about the flag situation, I haven't used those functions in a while)
  * update CHANGES to note that we have this new feature
  * update the docs---I don't think we have anything in the manual about individual commands, but maybe say a bit more in the docs for `:saw`, otherwise users would have not way to know about `CRYPTOL_SAW`

RyanGlScott (commented):
Please add an entry to the `CHANGES.md` file which describes these changes.

peterohanley-galois:
For the mock saw I added a sawCmd user option because I don't see a way to get an environment variable into the testing environment, but I'm concerned this creates a shell command hole.

PR #1848, opened 2025-04-28
Generate `:exhaust` for simple properties.

### LINKED ISSUE #1842: Improve project checking to auto validate easy properties
Currently, validating a project checks the docstrings on all definitions in a project.   Notably this *does not* validate any `property`s in project, unless they are explicitly listed in a docstring with a specific validation strategy (as we wouldn't know *how* to validate the property).

However, it is pretty common to have properties that a just a fixed test case (i.e., they are a value of type `Bool`).   There really is no ambiguity on how to check these---we should simply evaluate them and see if the answer is true.

We should modify project validation to automatically check such properties.  The semantics should be that a if `p` is a property of type bool, project validation should behave as if one had written a docstring containing `:exhaust p` (in this case "exhausting" all possible inputs is trivial as the property has no inputs).    In particular, we should remember to cache the outcome of validating the property so we won't need to re-evaluate it if nothing's changed.

The easiest way to implement this is probably to modify the pass that collects what commands need be executed to add the implicit `:exhaust p` commands, as described above.

ChrisEPhifer:
Closes #1842.

This works by adding `Text` of the form:

````md
```
:exhaust p
```
````

To the list of docstrings extracted from a module's `Decl`s, when `p` is a `Bit`-typed declaration that has the `property` `Pragma`.

I wasn't set up to run every test in the regression suite (particularly those requiring CVC4/CVC5/Bitwuzla/Yices); I've marked this as a draft for now, and will use CI to determine which (if any) additional golden/`.stdout` files I need to update.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Language changes
 
+* `Bit` properties (e.g. `property p = True`) will be checked with `:exhaust`,
+  unless their docstrings contain code blocks understood by `:check-docstrings`.
+  ([#1842](https://github.com/GaloisInc/cryptol/issues/1842))
+
 * `foreign` function declarations now support an optional calling convention
   keyword. See the [manual
   section](https://galoisinc.github.io/cryptol/master/FFI.html#calling-conventions)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,9 @@
 
 ## Language changes
 
-* `Bit` properties (e.g. `property p = True`) will be checked with `:exhaust`,
-  unless their docstrings contain code blocks understood by `:check-docstrings`.
+* When running the `:check-docstrings` command, `Bit` properties (e.g. `property
+  p = True`) will be checked with `:exhaust`, unless their docstrings contain
+  code blocks understood by `:check-docstrings`.
   ([#1842](https://github.com/GaloisInc/cryptol/issues/1842))
 
 * `foreign` function declarations now support an optional calling convention
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -1993,61 +1993,6 @@ moduleInfoCmd isFile name
        rPutStrLn "}"
        pure emptyCommandResult
 
--- | Extract the code blocks from the body of a docstring.
---
--- A single code block starts with at least 3 backticks followed by an
--- optional language specifier of "cryptol". This allowed other kinds
--- of code blocks to be included (and ignored) in docstrings. Longer
--- backtick sequences can be used when a code block needs to be able to
--- contain backtick sequences.
---
--- @
--- /**
---  * A docstring example
---  *
---  * ```cryptol
---  * :check example
---  * ```
---  */
--- @
-extractCodeBlocks :: T.Text -> [[T.Text]]
-extractCodeBlocks raw = go [] (T.lines raw)
-  where
-    go finished [] = reverse finished
-    go finished (x:xs)
-      | (spaces, x1) <- T.span (' ' ==) x
-      , (ticks, x2) <- T.span ('`' ==) x1
-      , 3 <= T.length ticks
-      , not (T.elem '`' x2)
-      , let info = T.strip x2
-      = if info `elem` ["", "repl"] -- supported languages "unlabeled" and repl
-        then keep finished (T.length spaces) (T.length ticks) [] xs
-        else skip finished ticks xs
-      | otherwise = go finished xs
-
-    -- process a code block that we're keeping
-    keep finished _ _ acc [] = go (reverse acc : finished) [] -- unterminated code fences are defined to be terminated by github
-    keep finished indentLen ticksLen acc (x:xs)
-      | x1 <- T.dropWhile (' ' ==) x
-      , (ticks, x2) <- T.span ('`' ==) x1
-      , ticksLen <= T.length ticks
-      , T.all (' ' ==) x2
-      = go (reverse acc : finished) xs
-
-      | otherwise =
-        let x' =
-              case T.span (' ' ==) x of
-                (spaces, x1)
-                  | T.length spaces <= indentLen -> x1
-                  | otherwise -> T.drop indentLen x
-        in keep finished indentLen ticksLen (x' : acc) xs
-
-    -- process a code block that we're skipping
-    skip finished _ [] = go finished []
-    skip finished close (x:xs)
-      | close == x = go finished xs
-      | otherwise = skip finished close xs
-
 data SubcommandResult = SubcommandResult
   { srInput :: T.Text
   , srResult :: CommandResult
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by RyanGlScott** [comment]
I think I'm missing something: what is the `:ensure` command mentioned in the PR title? Did you mean to say "`:exhaust`", or is `:ensure` something different entirely?

**Comment by ChrisEPhifer** [comment]
> I think I'm missing something: what is the `:ensure` command mentioned in the PR title? Did you mean to say "`:exhaust`", or is `:ensure` something different entirely?

... Oops. You're right, sorry. Let's not think too hard about what time it is on the west coast...

File: tests/docstrings/T05.icry.stdout [hunk 1]
```diff
@@ -12,4 +12,8 @@ Checking module T05
       Using exhaustive testing.
       Testing... Passed 1 tests.
       Q.E.D.
-Successes: 2, No fences: 1, Failures: 0
+    :exhaust p
+      Using exhaustive testing.
+      Testing... Passed 1 tests.
+      Q.E.D.
```
RyanGlScott, line 18: This example already has an explicit `:exhaust` pragma in the docstrings:

https://github.com/GaloisInc/cryptol/blob/86ee4b688c2e55493ae8cbb13979666654e97d0f/tests/docstrings/T05.md?plain=1#L13-L15

As such, this runs `:exhaust p` twice (you can see the output of the other `:exhaust p` invocation above). Is it possible to detect examples like this to avoid redundantly running `:exhaust` again?

tests/docstrings/T10.icry.stdout
```diff
@@ -22,12 +22,32 @@ Checking module T10
   Docstrings on T10::where_at__30_11::N
   Docstrings on T10::F1::N
   Docstrings on T10::M::q
+    :exhaust q
```
RyanGlScott, line 25: Relatedly, I wonder if this output should say something like "`:exhaust q (implicit)`" or some such to indicate that this is running due to the implicit behavior of how `:check-docstrings` works, rather than an explicit directive in the docstrings themselves.

RyanGlScott (commented):
This looks reasonable at a quick glance. I think it would be worth mentioning the fact that Cryptol implicitly `:exhaust`s `Bool`-typed properties somewhere in the [Documentation Comments](https://galoisinc.github.io/cryptol/master/BasicSyntax.html#documentation-comments) section of the Cryptol user manual, however.

tests/docstrings/T05.icry.stdout
ChrisEPhifer: Good idea. I can poke around at the code consuming the gathered-up docstrings / see about de-duplicating what's extracted there, rather than trying to detect this based on the docstring text itself.

ChrisEPhifer: Oh sure! If that's an acceptable adjustment to what we originally wanted, it's definitely an easy way to avoid duplicates without any kind of janky cleanup after gathering up / extracting all the code blocks.

I'll adjust what I had queued up re: changes to the reference manual and implement this check instead.

ChrisEPhifer: Actually -- We don't know about the structure of the docstring code blocks until the extraction is done, so we'll have to move some of this into the `REPL.Command` module.

I was hoping to just replace `<>` with `<|>` where I'm adding the new doctring test, but this is too aggressive, as it'll skip adding the `:exhaust` if there are _any_ docstring comments, even if there are no other code fences. That would have been cool :)

yav (approved):
Looks good to me, but let's do the fix where we only add the `:exhaust` if there's no user-defined code blocks.

We should also update the docs, and the CHANGES files before merging.

tests/docstrings/T05.icry.stdout
ChrisEPhifer: 6af7930d

tests/docstrings/T10.icry.stdout
ChrisEPhifer: 9cb4dde6

ChrisEPhifer:
> Looks good to me, but let's do the fix where we only add the `:exhaust` if there's no user-defined code blocks.
> 
> We should also update the docs, and the CHANGES files before merging.

e26fcf7e and ea496601

CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## Language changes
 
+* `Bit` properties (e.g. `property p = True`) will be checked with `:exhaust`,
+  unless their docstrings contain code blocks understood by `:check-docstrings`.
+  ([#1842](https://github.com/GaloisInc/cryptol/issues/1842))
```
RyanGlScott, line 7: When I first read this sentence, I was under the false impression that _all_ `Bit` properties would be checked with `:exhaust`. I think this should mention "when running `:check-docstrings`" upfront, as that is an important caveat. Something like:

```suggestion
* When running the `:check-docstrings` command, `Bit` properties (e.g. `property p = True`) will be checked with `:exhaust`,
  unless their docstrings contain code blocks understood by `:check-docstrings`.
  ([#1842](https://github.com/GaloisInc/cryptol/issues/1842))
```

RyanGlScott (approved):
LGTM, aside from some minor quibbles.

src/Cryptol/REPL/Command.hs
```diff
@@ -1993,61 +1993,6 @@ moduleInfoCmd isFile name
        rPutStrLn "}"
        pure emptyCommandResult
```
yav, line 1995: Why did `extractCodeBlocks` get moved to `Cryptol.TypeCheck.AST`?  I don't think that's a good place for it, as we already have too much stuff in that module.  As far as I can see, it's only called from `Command`, so keeping it there is probably OK.  If it does need to be moved for some reason, let's see what's a better place for it.

yav (changes requested):
I don't think `Crypto.TypeCheck.AST` is a good home for `extractBlocks`, let's find a better place.

src/Cryptol/REPL/Command.hs
ChrisEPhifer: OK, no problem -- I didn't love it either, I was just trying to make minimally-invasive changes that made sense in the existing module hierarchy, but I'm happy to add this new module.

ChrisEPhifer: dd8f8daa

CHANGES.md
ChrisEPhifer: aee7bfbc

src/Cryptol/TypeCheck/AST.hs
```diff
@@ -621,3 +622,65 @@ gatherModuleDocstrings nameToModule m =
...
+      if (null . extractCodeBlocks . fromMaybe "" . dDoc) d &&
+         (tIsBit . sType . dSignature) d &&
+         PragmaProperty `elem` dPragmas d
+      then Just $ "```\n" <> ":exhaust " <> (identText . nameIdent) (dName d) <> " // implicit" <> "\n```"
+      else Nothing
+
+-- | Extract the code blocks from the body of a docstring.
+--
+-- A single code block starts with at least 3 backticks followed by an
+-- optional language specifier of "cryptol". This allowed other kinds
```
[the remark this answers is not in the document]
ChrisEPhifer: faf00ff6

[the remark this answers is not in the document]
ChrisEPhifer: d041d04c

ChrisEPhifer (commented):
Alright, I've moved the docstring functions to a new module `Cryptol.TypeCheck.Docstrings`, and addressed the the small Haddock typos / lacking clarity Ryan caught in `CHANGES.md`.

yav (approved):
Looks good, thanks!

PR #1856, opened 2025-05-19
ModuleSystem: partially internalize renaming failure

### LINKED ISSUE #1822: Panic on multiple nested modules with the same name
Here's an example:
```cryptol
module Test where                                                                
                                                                                    
submodule S where                                                                   
  submodule E where                                                              
    x = 0x02                                                                     
                                                                                 
submodule S where    
```
This causes a panic at `lookupResolved`.

peterohanley-galois:
It doesn't panic but still fails. I don't know enough about the intended workings to say if this is correct.

There is a lot of churn from including the explicit possibility of failure in the `ResolvedLocal` type but I haven't carried this principle all the way through.

@yav, in the meeting today it sounded like you wanted the possibility of failure pushed into types completely, which I interpret as removing `mkFakeName` and similar and using explicit failure terms added to the types that are removed at the end of renaming, is this what you'd like done?

Resolves #1822


yav:
I am not sure I understand the question, but my preference would be not do big refactorings if possible.   The purpose of `mkFakeName` is to be a placeholder in situation where we've detected an error, but we don't want to abort execution immediately, and continue analyzing code in the hope of finding additional errors.   The idea is that such names should not be used for anything, and should not persists after the renamer has run.

The reason they are defined to `panic` is that we'd like to make sure that they are not used.  In the past we've had bugs where these end up being used because of the way code is structured.  So far we've managed to fix such situations, but it is usually a bit clunky.

I haven't looked in detail at what's happening in this case, but if it is too hard to just work around, some alternative solutions could be:
  * Change the code to stop after the first error.  This has the drawback of only reporting one error at a time, but it is likely to significantly simplify things, including removing the need for fake names.
  * Make fake names into normal names (i.e., not a panic), but with some marker, so we know that they are fake.   This would make it possible for code to notice them, and just do something sensible instead of exploding.   The drawback is that if we make a mistake they may end up flowing into the following passes, instead of causing a panic.

I am a bit busy with other stuff at the moment, but if you need more specific input on the issue here, we could do a quick meeting to come up with a plan.

sauclovian-g:
I would say the important bit is making sure that execution aborts at the end of the pass if any error has been issued. If the pass doesn't produce a result, the result can't have placeholder values :-)

(I'm in general more or less always in favor of continuing until some reasonable checkpoint, like the end of the pass, when an error happens... to the point that when laying out the error support for a new compiler generally there won't be a one-stop way to print a user-facing error and exit.)

peterohanley-galois:
> yav: my preference would be not do big refactorings if possible.

That's what's in the current smaller PR.

> sauclovian: I would say the important bit is making sure that execution aborts at the end of the pass if any error has been issued.

It's present.

src/Cryptol/ModuleSystem/Renamer/Monad.hs
```diff
@@ -50,7 +52,11 @@ data RenamerInfo = RenamerInfo
     -- ^ External modules
   }
 
-newtype RenameM a = RenameM { unRenameM :: ReaderT RO (StateT RW Lift) a }
+-- The ExceptionT here is for bailing when a fake value like mkFakeName is
+-- encountered. These values have already had an error recorded and are being
+-- processed for best-effor error recording so we do not need to return any
```
sauclovian-g, line 57: typo (`effor`) fwiw

src/Cryptol/ModuleSystem/Renamer/Monad.hs
```diff
@@ -207,12 +218,14 @@ setResolvedLocals mp (RenameM m) =
...
   do mp <- RenameM (roResolvedModules <$> ask)
-     pure case Map.lookup nm mp of
-            Just r  -> r
-
-            -- XXX: could this happen because we couldn't resolve a module?
-            Nothing -> panic "lookupResolved"
-                        [ "Missing module: " ++ show nm ]
+     case Map.lookup nm mp of
+       Just r -> pure r
+       -- XXX: could this happen because we couldn't resolve a module?
```
yav, line 223: Let's remove the `XXX`: this PR has answered and question in the affirmative :)

PR #1863, opened 2025-05-29
Language server

yav:
Start on a language server for Cryptol


marsella:
I'm excited about this! Here's some initial feedback. I realized after I wrote this up that perhaps my color scheme is different from yours, so maybe disregard the specific colors:
- Multiline comments like this are only are green on the first line (one or two stars at the top)
```
/**
 *
 */
 ``` 
- the `CRYPTOLPATH` is behaving funny, lots of import failures. 
	- When I'm in `cryptol-generated-specs/deps/cryptol-specs`, it has the path as
		- Searched paths: 
		  `/Users/marcella/mtv/cryptol-generated-specs/deps/cryptol-specs/Primitive/Symmetric/MAC/HMAC/Tests`
		  `/Users/marcella/mtv/cryptol-generated-specs`
		  `/Users/marcella/.cryptol`
		  `/Users/marcella/mtv/cryptol/language-server/share/cryptol`
		- Instead of the actual `CRYPTOLPATH=/Users/marcella/mtv/cryptol-specs` (which I get when I open a vscode terminal)
	- For me, the practical experience is that I have different versions of cryptol-specs in different folders and want to use the local one in each case -- e.g. in cryptol-specs itself I want the local version; when I'm dealing with refinements and things in cryptol-generated-specs I want the `deps/` version in there
	- Can I set this on a per-folder basis in VSCode? Or based on the workspace I have open? Even if not, it would be nice to make this a setting within the extension.
- Is `F12` supposed to jump to a definition? It doesn't work across files for me (or across qualified imports) but that's likely because of the path problem
	- It sometimes fails on local variables within files e.g. in AES/Specification.cry. Maybe due to other failures? Maybe due to trying to jump from a property? Maybe due to it being a functor?
- No type completion on constants. I would like this for when I'm trying to type a random value of a specific length.
- Some inconsistent coloring for syntax highlighting. E.g. in `Primitive/Symmetric/Cipher/Block/PRESENT.cry`:
	- In the definition of `sbox`, the first constant in the array is brown and the rest are green
	- in `roundEncrypt`'s type definition, the first `State` in the tuple is brown and the second is blue
	- in `PrEncrypt`'s type definition, `PRKeySize` is blue and `PRBlockSize` is brown
	- in `PrEncrypt_128`, there's some funky coloring across the board, in all the definitions
	- In `CipherInterface`, `KeySize` is blue? black? where it's declared and brown where it's used. Is that intentional?

podhrmic:
@yav your last commit successfully builds the language server in the CI, but it is not yet bundled in the packaged artifacts.

Issue #1832, opened 2025-03-28
Different behavior in Cryptol REPL vs. Python bindings involving Cryptol project and nested module names


RyanGlScott:
To reproduce this issue, you will need a sufficiently new version of Cryptol (3.3.0 or newer should work). You will also need to install the `cryptol` Python bindings.

Next, create a `project-flat-module-hierarchy` directory:

```
$ tree project-flat-module-hierarchy/ 
project-flat-module-hierarchy/
├── A.cry
└── cryproject.toml

0 directories, 2 files
```

With the following file contents:

```
// project-flat-module-hierarchy/A.cry
module A where

/**
 * ```
 * :check \x -> f x == x
 * ```
 */
f : [8] -> [8]
f x = x
```
```toml
# project-flat-module-hierarchy/cryproject.toml
modules = ["A.cry"]
```

Finally, create the following Python script in `project-flat-module-hierarchy`'s parent directory (i.e., this script should live next to the `project-flat-module-hierarchy` directory):

```python
# load_project.py
import json
import cryptol
from cryptol.single_connection import *

def main():
    connect(verify=False)
    print(json.dumps(load_project('project-flat-module-hierarchy', 'untested'), indent=2))

if __name__ == "__main__":
    main()
```

Observe that both the Cryptol REPL and the Python bindings are able to load this project successfully:

```
# Cryptol REPL
$ rm -rf project-flat-module-hierarchy/.cryproject
$ ~/Software/cryptol-3.3.0/bin/cryptol -p project-flat-module-hierarchy
Scanning A.cry
Scanning /home/ryanscott/Software/cryptol-3.3.0/share/cryptol/Cryptol.cry
Loading module Cryptol
Loading module A
Checking module A
  Docstrings on A
  Docstrings on A::f
    :check \x -> f x == x
      Using random testing.
      Testing... Passed 100 tests.
      Expected test coverage: 32.39% (83 of 256 values)
  Successes: 1, No fences: 1, Failures: 0
<snip>

# Python bindings
$ rm -rf project-flat-module-hierarchy/.cryproject
$ CRYPTOL_SERVER=~/Software/cryptol-3.3.0/bin/cryptol-remote-api poetry run python load_project.py 
{
  "scan_status": [
    {
      "file": "/home/ryanscott/Documents/Hacking/Haskell/cryptol/cryptol-remote-api/python/project-flat-module-hierarchy/A.cry",
      "scanned": {
        "changed": true,
[...]

PR #1866, opened 2025-06-05
Doctest support for Python API

### LINKED ISSUE #1865: Loading projects from the Python API does not work correctly
There's a few issues that prevent running doctests from the Cryptol API, namely:
  * The names of the scanned modules are incorrect, making it impossible to test them
  * There's a bug in loading projects that prevents us from finding the file (see #1832)
  * We produce too much output for testing commands (we don't need the progress bar output)
  * We don't have a good example of how to run doctests with the Python API

yav:
Fixes #1865, #1832

This PR contains a collection of fixes to the support for running docstings tests from the Python API, using projects and caching.  Besides the bugfixes, the PR adds a detailed Python script with an example of how to use the API to run tests with caching.


weaversa:
@yav Would you be willing to throw [junit xml](https://pypi.org/project/junitxml/) into your example script, to support CI results?

cryptol-remote-api/src/CryptolServer/Utils.hs
```diff
@@ -0,0 +1,51 @@
...
+
+import Data.Text(Text)
+import qualified Data.Text as Text
+import qualified Data.ByteString as BS
+import Data.Word
+import Data.Bits
+
+import Data.Aeson
+
+newtype BS = BS BS.ByteString
```
RyanGlScott, line 11: What is this type? Why do we need it separately from `ByteString`? Including some Haddocks would be helpful.

PR #1873, opened 2025-06-24
Attempt to build the language server in the CI

podhrmic:
I need to get a language server binary for Ubuntu 22.04, this seemed to be the easier way to get it


yav:
Thanks for noticing this!  I just pushed a change to `language-server` branch to add the equivalent change, to keep the PR dependencies simple, so I think we can close this.

PR #1880, opened 2025-07-16
Allow `;` between case branches

glguy:
This change removes an unnecessary shift/reduce error in the parser and removes previously unreachable grammar productions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,10 @@
   section](https://galoisinc.github.io/cryptol/master/FFI.html#the-abstract-calling-convention)
   for more information.
 
+* Allow an explicit `;` separator between `case` branches. This change removes the
+  unreachable code in the grammar for `case` and `where` expressions with explicit
+  `{` and `}` that was never reachable due to the way the layout rule worked.
+
 ## Bug fixes
 
 * Allow changing the `tcSolver` setting to non-Z3 solvers (e.g., CVC5) without
--- a/src/Cryptol/Parser.y
+++ b/src/Cryptol/Parser.y
@@ -44,15 +44,6 @@ import Cryptol.Utils.RecordMap(RecordMap)
 import Paths_cryptol
 }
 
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -17,6 +17,10 @@
   section](https://galoisinc.github.io/cryptol/master/FFI.html#the-abstract-calling-convention)
   for more information.
 
+* Allow an explicit `;` separator between `case` branches. This change removes the
+  unreachable code in the grammar for `case` and `where` expressions with explicit
+  `{` and `}` that was never reachable due to the way the layout rule worked.
```
RyanGlScott, line 22: A follow-up question is: do we also want to support the ability to write things like `case x of { ...; ... }` with explicit braces? I am inclined to answer "yes" to this question, given that GHC also supports this, and we are largely stealing GHC's syntactic quirks. (This is not something that we need to do as part of this PR, given that the code path that supported parsing `case x of { ...; ... }` never worked in the first place, but we should open a follow-up issue about this.)

RyanGlScott (approved):

sauclovian-g:
Am I right in thinking the grammar for those bits is fine and the problem with them not working is entirely upstream in the layout handling? (That is, I don't see a problem with them in the grammar... but I could easily be missing something)

yav:
The challenge is that we use `{ }` for both records and blocks, and the two clash when you write record patterns.    Specifically when you see `where {` you don't know if the `{` is supposed to start a declaration block or a record pattern in a pattern binding.

At present we don't support record patterns in `case` so, in theory, we could make this work, by adjusting the layout rule to only start a layout block if `of` is not followed by `{`.     Otoh, this will not work for `where` and having the two be inconsistent is odd.  Also, we may want to add support for fancier patterns in the future.

This is a bit of an unfortunate state but things have been like that for quite some time, and it hasn't been that much of an issue, so I think it's fine to stick to just using layout.   We could always revisit the decision if we really need some way to write non-layout blocks.

PR #1886, opened 2025-08-01
Get library building with FFI disabled

glguy:
Being able to turn off the FFI is super-handy for being able to load cryptol in ghci and Haskell-language-server.

Obviously it should build with the feature turned off (if we're going to keep the feature). We might want an additional feature that *specifically* disabled the foreign exports just for when we're developing anything that's not the FFI itself.

This PR just gets things building again. We might want to solve some of the problems differently than implemented here.

### DIFF
--- a/src/Cryptol/Eval/FFI.hs
+++ b/src/Cryptol/Eval/FFI.hs
@@ -10,11 +10,17 @@ module Cryptol.Eval.FFI
   ) where
 
 import Cryptol.Eval.FFI.ForeignSrc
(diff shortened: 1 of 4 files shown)


RyanGlScott:
Ah! Now that you mention it, I _have_ run into this error, but I forgot that I applied a workaround by putting this in my `cabal.project.local` file:

```
package cryptol
  ghc-options: -fobject-code
```

RyanGlScott (approved):
In any case, let's go ahead and land this.

PR #1888, opened 2025-08-08
Add an RME-based prover

glguy:
This brings the RME tactic from SAW directly into Cryptol


sauclovian-g:
We should really not paste a second copy of the whole thing, even though it's not that large...

glguy:
> We should really not paste a second copy of the whole thing, even though it's not that large...

I agree. If saw-script wants to add a new submodule that's one option. The other is that saw script can switch to using this mini library out of the Cryptol dep it has.

RyanGlScott:
I would be fine with moving `rme` to this repo and then switching SAW to depend on this repo's version of the library instead.

sauclovian-g:
(I am fine with either making the rme library its own thing or moving it to the Cryptol library)

glguy:
I've created a new repository for this package: https://github.com/GaloisInc/rme

It's marked *private* at the moment, but only until we're agreed to go forward with it.

RyanGlScott (approved):
Thanks! This looks good to me, aside from the following:

* Make sure to advertise the existence of the new `w4-rme` prover in the changelog.
* The CI failure is likely caused by `rme-what4` having overly restrictive lower version bounds on `base`, which should be easy to rectify.

PR #1892, opened 2025-08-15
Fix conversion from values back to expressions

### LINKED ISSUE #1891: Convering a value to an expression creates malformed terms
The problem is in `toExpr` in `Cryptol.Eval.Concrete`---in some cases we forgot to add `TPRoofApp` to match the constraints on polymorphic functions, which results in malformed `Expr`s

yav:
Fixes #1891

We were missing a `EProofApp` constructor when making numeric literals

### DIFF
--- a/src/Cryptol/Eval/Concrete.hs
+++ b/src/Cryptol/Eval/Concrete.hs
@@ -119,11 +119,11 @@ toExpr prims t0 v0 = findOne (go t0 v0)
            ETuple <$> (zipWithM go ts =<< lift (sequence tvs))
       (TVBit, VBit b) ->
         pure (prim (if b then "True" else "False"))
-      (TVInteger, VInteger i) -> pure (numlit i tInteger)
+      (TVInteger, VInteger i) -> pure (intlit i)
       (TVIntMod n, VInteger i) -> pure (numlit i (tIntMod (tNum n)))
       (TVRational, VRational (SRational n d)) ->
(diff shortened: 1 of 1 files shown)


src/Cryptol/Eval/Concrete.hs
```diff
@@ -144,6 +141,15 @@ toExpr prims t0 v0 = findOne (go t0 v0)
       (_,VNumPoly{}) -> mzero
       _ -> mismatch
     where
+      intlit n
+        | n < 0 = EApp (EProofApp (ETApp (prim "negate") tInteger))
+                       (numlit (-n) tInteger)
+        | otherwise = numlit n tInteger
+
+      numlit i t =
+        EProofApp (ETApp (ETApp (prim "number") (tNum (i::Integer))) t)
```
RyanGlScott, line 150: Probably worth a comment that `numlit i t` has the precondition that `i` should be a non-negative integer (hence why `intlit` has a special case for negative integers to satisfy the precondition).

RyanGlScott (commented):
GitHub won't let me highlight it in a review comment, but there is one more occurrence of `prim "number"` in `toExpr.go`:

https://github.com/GaloisInc/cryptol/blob/e9360454f24f840ecff59db4be528b488c84cd4c/src/Cryptol/Eval/Concrete.hs#L137-L139

Should this be changed to `numlit` as well?

RyanGlScott (approved):

PR #1895, opened 2025-08-26
Use `rme{,-what4}` dependencies from Hackage

RyanGlScott:
This brings the treatment of `rme` in line with how `what4` (also a Hackage dependency) is handled.


glguy (approved):
This seems simpler; thanks!

PR #1899, opened 2025-09-05
CI: Ensure that Docker image jobs clone repo before running `.github/ci.sh`

RyanGlScott:
A follow-up to #1897, which accidentally broke the Docker-related CI jobs. This ensures that Docker-related CI jobs always clone the `cryptol` repo before running the `.github/ci.sh` script to ensure that the script is actually available to run on a CI runner.


RyanGlScott:
Manually triggered CI workflow demonstrating that the Docker-related jobs now pass: https://github.com/GaloisInc/cryptol/actions/runs/17491231133

PR #1904, opened 2025-09-16
Make validation commands use the currently focused module

### LINKED ISSUE #1903: Module-level docstrings behave differently across `--project` flag and the interpreter
Module-level docstring behave differently in the interpreter compared to using the project flag. In the interpreter, they run. With the project flag, I get a "There are no properties in scope" error.

Consider the following code
```
/**
 * ```repl
 * :prove
 * ```
 */
module Potato where

property yukon = True
```
with a project file `cryproject.toml`:
```
modules = [
    "Potato.cry"
]
```

I get the following output on the command line:
```
% cryptol --project .
Scanning Potato.cry
Scanning Cryptol
Loading module Cryptol
Loading module Potato
Checking module Potato
  Docstrings on Potato
    :prove
      There are no properties in scope.
  Docstrings on Potato::yukon
  Successes: 0, No fences: 1, Failures: 1
Passing: 0, Failing: 1, Missing: 1
% echo $?
1
```

And the following in the interpreter:
```
Loading module Cryptol
Cryptol> :l Potato.cry 
Loading module Cryptol
Loading module Potato
Potato> :check-docstrings 
Checking module Potato
  Docstrings on Potato
    :prove
      :prove yukon
      	Q.E.D.
      (Total Elapsed Time: 0.017s, using "Z3")
  Docstrings on Potato::yukon
Successes: 1, No fences: 1, Failures: 0
```

yav:
Fixes #1903

This is a change from the previous behavior where we use the "loaded module" from the REPL monad to determine what properties to check.  The effect is proper support for submodules.   With the new change, running `:check` with a focused submodule (or when using a docstring) should only check the things in that module, while previously it would check the whole file.

This also fixes #1903, because loading a project does not set the "loaded module" in the REPL, which lead to confusing differences between docstring validation from the REPL vs with a project.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,15 @@
 
 ## Language changes
 
+* When running validation commands (`:check`, `:prove`, `:exhaust`, etc)
+  without an explicit argument, we now run only the properties in the
+  currently *focused module*.  This is a change in behavior, because previously
+  we used to run all properties in the currently opened *file*.  This change
+  is only noticable when working with nested modules.  The new behavior works
+  better when these commands are used from docstrings (e.g., with the
+  new behavior, writing `:check` on a submodule, will only check the properties
+  in that submodule, as expected).  
+
 * When running the `:check-docstrings` command, `Bit` properties (e.g. `property
   p = True`) will be checked with `:exhaust`, unless their docstrings contain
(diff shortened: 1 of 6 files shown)


RyanGlScott (commented):
Is it worth adding a regression test for this?

PR #1909, opened 2025-09-19
Change how :t pp to include mod ctxt

Ra1issa:
Prior to this commit `:t` did not include the type parameters that were available at the module level. 
This closes issue #1867

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -353,7 +353,7 @@
 ## Bug fixes
 
 * Fix an issue where `:t` did not take into account module level type
-parameters when pretty printing type signature.
+parameters when pretty printing type signature (closes issue #1867).
 
 * The What4 backend now properly supports Boolector 3.2.2 or later.
 


src/Cryptol/REPL/Command.hs
```diff
@@ -1166,15 +1177,34 @@ typeOfCmd str pos fnm = do
 
   -- XXX need more warnings from the module system
   whenDebug (rPutStrLn (dump def))
+
```
yav, line 1180: I think there might be a more direct way to do this.   I am thinking like this: `getFocusedEnv` is a `ModContext`, which has a `NamingEnv` inside it.  You can use `toNameDisp` to convert this to a `NameDisp`, and put this inside the `PPCfg`.  Then, I think just doing `pp` on the `Schema` should print it correctly.

yav (changes requested):
Could you give the change I suggested a try.  Also, have a look add the failing tests and see if you can fix them.  When names change, try to understand why, and fix the test if needed.  In particular, if the example has some parameters with the given names in scope, I'd expect the printing of the schemes to change.  Otherwise, hopefully they shold stay as they were (if not we should try to understand why they changed)

yav:
You should add a test case with the example from the issue

Issue #1913, opened 2025-10-07
`loadedNominalTypes` is unnecessarily inefficient


RyanGlScott:
In the code for `loadedNominalTypes`, we have:

https://github.com/GaloisInc/cryptol/blob/06dc1ef0fb0e9f55adcef3a58a61ce3f097ee87a/src/Cryptol/ModuleSystem/Env.hs#L213-L217

Calling `ifNominalTypes (ifDefines i) <> ifNominalTypes (ifDefines i)` is wasteful, as it is unioning the same map together and therefore returning an identical map. We should use `ifNominalTypes (ifDefines i)` instead to avoid this inefficiency.

PR #1914, opened 2025-10-07
`Cryptol.ModuleSystem.Env`: Eliminate wasted work in `loadedNominalTypes` 

RyanGlScott:
Don't use `ifNominalTypes (ifDefines i) <> ifNominalTypes (ifDefines i)`, as this unions together identical maps and returns the same map as if you'd just used `ifNominalTypes (ifDefines i)` directly.

Fixes https://github.com/GaloisInc/cryptol/issues/1913.


mtullsen (commented):
Looks good to me.

Issue #1654, opened 2024-04-11
Mention all type class instances in the reference manual


RyanGlScott:
The Cryptol reference manual has a list of type classes and their instances [here](https://galoisinc.github.io/cryptol/3.1.0/OverloadedOperations.html). However, it's missing quite a few instances:

## `Eq`

* `instance (fin n) => Eq [n]Bit` (this works differently from `instance (fin n, Eq a, a /= Bit) => Eq [n]a`, which _is_ mentioned)

## `Cmp`

* `instance (fin n) => Cmp [n]Bit` (this works differently from `instance (fin n, Cmp a, a /= Bit) => Cmp [n]a`, which _is_ mentioned)

## `Zero`

* `instance Zero [n]Bit` (this works differently from `instance (Zero a, a /= Bit) => Zero [n]a`, which _is_ mentioned)

## `Logic`

* `instance Logic [n]Bit` (this works differently from `instance (Logic a, a /= Bit) => Logic [n]a`, which _is_ mentioned)

## `Round`

* `instance Round Rational`

## `Literal` (not mentioned at all)

* `instance (1 >= val) => Literal val Bit`
* `instance (fin val) => Literal val Integer` 
* `instance (fin val) => Literal val Rational` 
* `instance (fin val, fin n, n >= 1, n > val) => Literal val (Z n)` 
* `instance (fin val, fin n, n >= width val) => Literal val [n]`
* `instance (ValidFloat e p) => Literal val (Float e p)` (if `val` is representable)

## `LiteralLessThan` (not mentioned at all)

* `instance (2 >= val) => LiteralLessThan val Bit`
* `instance LiteralLessThan val Integer`
* `instance LiteralLessThan val Rational`
* `instance (fin n, n >= 1, n >= val) => LiteralLessThan val (Z n)`
* `instance (fin n, n >= lg2 val) => LiteralLessThan val [n]`
* `instance (ValidFloat e p) => LiteralLessThan val (Float e p)` (if `val-1` is representable)

## `FLiteral` (not mentioned at all)

* `instance (fin m, fin n, n >= 1) => FLiteral m n r Rational`
* `instance (ValidFloat e p, fin m, fin n, n >= 1) => FLiteral m n 0 (Float e p)`
* `instance (ValidFloat e p, r != 0) => FLiteral m n r (Float e p)` (if `m/n` is representable)

## Other omissions

[...]

PR #1926, opened 2025-11-01
Support deriving instances for newtypes and enums

### LINKED ISSUE #1587: Deriving instances for newtypes and enums

### LINKED ISSUE #1925: Comparison operators are too strict
According to https://github.com/GaloisInc/cryptol/issues/398#issuecomment-770077142 and the behavior of the reference interpreter, comparisons should be lazy:
```
Cryptol> :eval (False, undefined) < (True, True)
True
Cryptol> :eval (False, undefined) == (True, True)
False
```
But the current behavior in the real interpreter seems to be too strict:
```
Cryptol> (False, undefined) < (True, True)

Run-time error: undefined
-- Backtrace --
Cryptol::error called at Cryptol:1062:13--1062:18
Cryptol::undefined called at <interactive>:1:9--1:18
(Cryptol::<) called at <interactive>:1:1--1:34
Cryptol> (False, undefined) == (True, True)

Run-time error: undefined
-- Backtrace --
Cryptol::error called at Cryptol:1062:13--1062:18
Cryptol::undefined called at <interactive>:2:9--2:18
(Cryptol::==) called at <interactive>:2:1--2:35
```
This should be fixed to align with the conclusion reached in #398.

qsctr:
Adds support for the `deriving` clause after `newtype` and `enum` declarations, typechecking for derived instances, and runtime implementations of derived instances in the interpreter (both concrete and symbolic) and the reference evaluator.

Fixes #1587. Also fixes #1925.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,11 @@
 * Fix browsing of `main` modules.
   crashing. ([#1874](https://github.com/GaloisInc/cryptol/issues/1857))
 
+* Make comparison operators lazy so that they do not evaluate any more of the
+  data structure than is required to determine the comparison result, matching
+  the behavior of the reference evaluator.
+  ([#1925](https://github.com/GaloisInc/cryptol/issues/1925))
+
 ## New Features
 
 * New REPL command `:saw` to run SAW on a SAW file, usable in docstrings.


docs/RefMan/OverloadedOperations.rst
```diff
@@ -254,3 +280,71 @@ Rounding
...
+
+.. code-block:: cryptol
+
+  enum T a b = A a | B b deriving (Eq, Cmp)
+
+then ``Eq (T a b)`` will only hold if ``Eq a`` and ``Eq b`` hold, and ``Cmp (T a
+b)`` will only hold if ``Cmp a`` and ``Cmp b`` hold.
+
+Two enum values of the same type compare equal if they have the same constructor
+and the same field values for that constructor, and unequal otherwise.
```
RyanGlScott: It might be useful to illustrate this point with some examples, such as the results of evaluating things like `A 0x0 == A 0x0`, `A 0x0 == B 0x0`, `A 0x0 == A 0x1`, etc.

docs/RefMan/OverloadedOperations.rst
```diff
@@ -254,3 +280,71 @@ Rounding
...
+
+Two enum values of the same type are ordered first by their constructor, then
+for values with the same constructor, by lexicographic ordering on the fields of
+that constructor. The order of constructors is the order in which they are
+listed in the ``enum`` declaration; constructors listed earlier compare less
+than constructors listed later. In the above example of ``T a b``, any value
+with constructor ``A`` always compares less than any value with constructor
+``B``. For signed comparisons, fields are compared using signed comparisons but
+constructors themselves are still ordered in the same way.
```
RyanGlScott: Similarly, it would be helpful to illustrate this paragraph with some examples.

src/Cryptol/Eval/Reference.lhs
```diff
@@ -1284,8 +1297,10 @@ False]`, but to `error "foo"`.
 >           do val' <- val
 >              pure $ VRecord [ (f, go fty (lookupRecord f val'))
 >                             | (f, fty) <- canonicalFields fs ]
->         TVNominal {} ->
->           evalPanic "arithUnary" ["Nominal type not in `Ring`"]
+>         TVNominal _ _ ntv ->
+>           case ntv of
+>             TVStruct fs -> go (TVRec fs) val
+>             _ -> evalPanic "arithUnary" ["Enum/Abstract type not in `Ring`"]
```
RyanGlScott: Similarly, this (and the other `evalPanic` cases) should mention `"ringUnary"` instead of `"arithUnary"` here.

src/Cryptol/Eval/Reference.lhs
```diff
@@ -1332,8 +1347,10 @@ False]`, but to `error "foo"`.
 >              pure $ VRecord
 >                [ (f, go fty (lookupRecord f l') (lookupRecord f r'))
 >                | (f, fty) <- canonicalFields fs ]
->         TVNominal {} ->
->           evalPanic "arithBinary" ["Nominal type not in class `Ring`"]
+>         TVNominal _ _ ntv ->
+>           case ntv of
+>             TVStruct fs -> go (TVRec fs) l r
+>             _ -> evalPanic "arithBinary" ["Enum/Abstract type not in class `Ring`"]
```
RyanGlScott: Similarly, this (and the other `evalPanic` cases) should mention `"ringBinary"` instead of `"arithBinary"` here.

tests/deriving/deriving-tc-unsolvable.cry
RyanGlScott: I think it would be helpful to also have some failing test cases that demonstrate syntactically invalid examples of `deriving` that folks might try. The following examples come to mind:

* `deriving ()`
* `deriving Eq` (this is valid in Haskell, but not in Cryptol)
* `deriving (Literal 1)` (something like this would be valid in GHC, not in Cryptol, each class has to be a bare class name)
* `deriving (Eq) deriving (Cmp)` (such a thing is valid with GHC's `DerivingStrategies` language extension, but not in Cryptol)

docs/RefMan/OverloadedOperations.rst
```diff
@@ -254,3 +280,71 @@ Rounding
     - ``ValidFloat e p``
 
 
+.. _derived-instances:
+
+Derived instances
+-----------------
```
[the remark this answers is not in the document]
qsctr: It's unclear to me which documents in `docs/` are supposed to be up-to-date and which are old. For instance, `Syntax.md` doesn't cover `enum` syntax, and nor does the Programming Cryptol book, so I ended up assuming that the reference manual is the only thing we are keeping up-to-date.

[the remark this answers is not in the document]
RyanGlScott: Yeah, the duplication between the reference manual and the various markdown files under `docs` is a bit unfortunate. Personally, I would favor folding all of the material in the markdown files into the reference manual and making the latter the sole place where documentation of this sort is curated. (This isn't something that we necessarily need to do as part of this PR, however.)

Perhaps when we fix https://github.com/GaloisInc/cryptol/issues/1654, we can remove the material from `docs/CryptolPrims.md` about overloaded operations and simply have it point to the reference manual instead.

tests/deriving/deriving-tc-unsolvable.cry
qsctr: I implemented it that way because you said so in https://github.com/GaloisInc/cryptol/issues/1587#issuecomment-1908724303 :)

I don't have a strong opinion about it either, so I am fine with keeping it as is.

docs/RefMan/OverloadedOperations.rst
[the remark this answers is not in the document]
yav: We should obsolete `Syntax.md`---it was the starting point for the reference manual.  I don't know what `CryptolPrims.md` is, but I also think we should get rid of that and just maintain the manual and the book.

src/Cryptol/Parser.y
```diff
@@ -69,6 +69,7 @@ import Paths_cryptol
   'type'      { Located $$ (Token (KW KW_type   ) _)}
   'newtype'   { Located $$ (Token (KW KW_newtype) _)}
   'enum'      { Located $$ (Token (KW KW_enum) _)}
+  'deriving'  { Located $$ (Token (KW KW_deriving) _)}
```
yav: Does adding `deriving` to `ident` cause conflicts?  If not, we should probably add it there.

src/Cryptol/TypeCheck/Subst/Type.hs
```diff
@@ -0,0 +1,275 @@
...
+    contain module parameters, which need to be substituted when
+    instantiating a functor. -}
+    TNominal nt ts ->
+      uncurry TNominal <$> anyJust2 (applySubstToNominalType' simpProp su)
+                                    (anyJust (apSubstMaybe' simpProp su))
+                                    (nt,ts)
+
+    TVar x -> applySubstToVar' (apSubstMaybe' simpProp) su x
+
+-- | Apply a substitution without simplifying the result when it is a predicate.
```
yav: Maybe add a note here on why we need this.  As far as I understand, we use it while instantiating nominal types in the class solver, and we don't want to simplify stuff because we are in  the process of solving, so we'd get in a loop.  Is that right?

yav (approved):

src/Cryptol/Parser.y
qsctr: I see this when I build the parser with `'deriving'` in `ident`:
```
shift/reduce conflicts:  2
reduce/reduce conflicts: 2
```
so I guess the answer is yes?

docs/RefMan/OverloadedOperations.rst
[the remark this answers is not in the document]
qsctr: I agree with this approach. I've created #1932 to track this.

Meanwhile, I've only added a brief mention of derived instances in `CryptolPrims.md`, with a pointer to the manual if the reader wants to learn more.

PR #1928, opened 2025-11-05
Temporary change to test the 'what4' 1.7.1 release candidate.

aschwerdfeger-galois:


RyanGlScott:
Hah, I think the CI is failing for a very silly reason: it's because the `cry` script invokes `cabal v2-exec cryptol` on individual `cryptol` files to test them, but by default, `cabal` prints out extra information if you download a `cabal.project` package from a Hackage URL, causing the `diff`s to be different.

We might consider changing this line:

https://github.com/GaloisInc/cryptol/blob/202fd60bf87dafda9d9fd32a5d2f49a524fa9707/cry#L40

To use `  cabal -v0 v2-exec cryptol -- $*` instead. That would fix this problem, I believe, and it would be more consistent with other `cry` subcommands that are already using `-v0` to suppress these sorts of `cabal` diagnostic messages.

aschwerdfeger-galois:
I will submit the changes to suppress the diagnostic messages in a separate PR.

Issue #1894, opened 2025-08-25
REPL's tab-completion suggests unusable internal function names arising from numeric constraint guards


RyanGlScott:
To reproduce the issue, load this code into the Cryptol REPL:

```
foobar : {n} [n]
foobar
  | n == 1 => 0b1
  | n != 1 => zero
```

Now type `foobar` into the REPL and use tab completion. I was expecting `foobar` to be the only suggestion, but to my surprise, it actually came up with three different suggestions:

```
Main> foobar
foobar          foobar(n != 1)  foobar(n == 1)
```

The latter two, `foobar(n != 1)` and `foobar(n == 1)`, are internal names that are used to power `foobar`'s numeric constraint guards under the hood. They should not be suggested in tab-completion results, as they are unusable otherwise:

```
Main> :t foobar(n != 1)

[error] at <interactive>:10:11--10:12
    Value not in scope: n
```

Issue #1908, opened 2025-09-19
Incorrect shadowing warning with submodules and pattern matching


RyanGlScott:
Cryptol emits an incorrect shadowing warning for the following program:

```
f = ()
  where
    [x] = 0b1

submodule S where
  [a] = 0b0
```
```
$ ~/Software/cryptol-3.3.0/bin/cryptol Bug.cry
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.3.0 (285c30d)
https://cryptol.net  :? for help

Loading module Cryptol
Loading module Main
[warning] at Bug.cry:3:5--3:8
    This binding for `__p0` shadows the existing binding at
    Bug.cry:6:3--6:6
```

The `__p0` mentioned in this warning is an internal name that arises when pattern matching on `[x]` and `[a]`. There is no reason why Cryptol shouldn't be able to make the `__p0` for `[x]` distinct from the `__p0` for `[a]`, so I claim that this warning is incorrect. Moreover, Cryptol users can't even write code in terms of `__p0`, so even if this warning were accurate, it's not actionable.

PR #1945, opened 2025-11-20
Fix REPL Tab Completion error

Ra1issa:
This MR closes issue #1894 and #1908

### DIFF
--- a/src/Cryptol/ModuleSystem/Binds.hs
+++ b/src/Cryptol/ModuleSystem/Binds.hs
@@ -31,7 +31,7 @@ import qualified MonadLib as M
 import Cryptol.Utils.Panic (panic)
 import Cryptol.Utils.Ident(allNamespaces)
 import Cryptol.Parser.Position
-import Cryptol.Parser.Name(isGeneratedName)
+import Cryptol.Parser.Name(isSystemName)
 import Cryptol.Parser.AST
 import Cryptol.ModuleSystem.Exports(exportedDecls,exported)
 import Cryptol.ModuleSystem.Renamer.Error
@@ -434,7 +434,7 @@ newTop ::
   FreshM m => Namespace -> ModPath -> PName -> Maybe Fixity -> Range -> m Name
 newTop ns m thing fx rng =
   liftSupply (mkDeclared ns m src (getIdent thing) fx rng)
-  where src = if isGeneratedName thing then SystemName else UserName
+  where src = if isSystemName thing then SystemName else UserName
 
 newLocal :: FreshM m => Namespace -> PName -> Range -> m Name
 newLocal ns thing rng = liftSupply (mkLocalPName ns thing rng)
--- a/src/Cryptol/Parser/Name.hs
+++ b/src/Cryptol/Parser/Name.hs
@@ -20,7 +20,7 @@ module Cryptol.Parser.Name (
   , origNameToDefPName
   , getModName
   , getIdent
-  , isGeneratedName
+  , isSystemName
   , pattern UnQual
   ) where
 
@@ -100,11 +100,16 @@ getIdent (NewName p i) = packIdent ("__" ++ pass ++ show i)
            MonoValues         -> "mv"
            ExpandPropGuards _ -> "epg"
 
-isGeneratedName :: PName -> Bool
-isGeneratedName x =
+
+
+isSystemName :: PName -> Bool
+isSystemName x =
   case x of
-    NewName {} -> True
-    _          -> False
+    UnQual' _id ns ->   case ns of
+                          SystemName -> True
+                          UserName -> False
+    Qual _md _id -> False
+    NewName _p _i -> True
 
 instance PP PName where
   ppPrec _ = ppPrefixName
--- a/tests/issues/issue1894/issue_1894.icry.stdout
+++ b/tests/issues/issue1894/issue_1894.icry.stdout
@@ -237,6 +237,4 @@ Symbols
   ---------
        
     foobar : {n} [n]
-    foobar(n != 1) : {n} (n != 1) => [n]
-    foobar(n == 1) : {n} (n == 1) => [n]
      


src/Cryptol/Parser/Name.hs
```diff
@@ -71,6 +106,12 @@ getIdent (NewName p i) = packIdent ("__" ++ pass ++ show i)
            MonoValues         -> "mv"
            ExpandPropGuards _ -> "epg"
 
+getNameSource :: PName -> Maybe NameSource
```
yav: This doesn't seem to be used anywhere?   If it is not, we should remove it.

src/Cryptol/ModuleSystem/Name.hs
```diff
@@ -21,7 +21,6 @@
 module Cryptol.ModuleSystem.Name (
     -- * Names
     Name(), NameInfo(..)
-  , NameSource(..)
```
yav: We could still export `NameSource(..)` from here, even though it is not defined in this module---it would probably save on changing some imports

yav: Let's reexport this from there, so we don't need all the extra imports in the other places.

src/Cryptol/ModuleSystem/NamingEnv.hs
```diff
@@ -29,7 +29,7 @@ import Cryptol.Utils.Ident(allNamespaces,packModName,modNameChunksText)
 import Cryptol.Parser.AST
 import qualified Cryptol.TypeCheck.AST as T
 import Cryptol.ModuleSystem.Name
-import Cryptol.ModuleSystem.Names
+import Cryptol.ModuleSystem.Names
```
yav: White space at the end of the line.

src/Cryptol/ModuleSystem/Binds.hs
```diff
@@ -31,7 +31,7 @@ import qualified MonadLib as M
 import Cryptol.Utils.Panic (panic)
 import Cryptol.Utils.Ident(allNamespaces)
 import Cryptol.Parser.Position
-import Cryptol.Parser.Name(isGeneratedName)
+import Cryptol.Parser.Name(isGeneratedName, NameSource (..))
```
yav: No need for changes

src/Cryptol/REPL/Monad.hs
```diff
@@ -658,15 +660,34 @@ rPrintDoc doc =
...
+filterMap f = Map.fromList . filter f . Map.toList
+
 -- | Get visible variable names.
--- This is used for command line completition.
+-- This is used for command line completion.
 getExprNames :: REPL [String]
 getExprNames =
-  do fNames <- M.mctxNames <$> getFocusedEnv
-     return (map (show . pp) (Map.keys (M.namespaceMap M.NSValue fNames)))
+  do
```
yav: This can be simplified---there's no reason to filter the map twice.

src/Cryptol/Parser.y
```diff
@@ -504,24 +502,24 @@ qop                              :: { LPName }
...
 
     -- special cases for operators that are re-used elsewhere
-  | '*'                             { Located $1 $ mkUnqual $ mkInfix "*" }
-  | '+'                             { Located $1 $ mkUnqual $ mkInfix "+" }
-  | '-'                             { Located $1 $ mkUnqual $ mkInfix "-" }
-  | '~'                             { Located $1 $ mkUnqual $ mkInfix "~" }
-  | '^^'                            { Located $1 $ mkUnqual $ mkInfix "^^" }
-  | '<'                             { Located $1 $ mkUnqual $ mkInfix "<" }
-  | '>'                             { Located $1 $ mkUnqual $ mkInfix ">" }
+  | '*'                             { Located $1 $ mkUnqualUser $ mkInfix "*" }
```
yav: Maybe `mkUnqual` should do what `mkUnqualUser` does, and we should have `mkUnqualSystem` for the cases when we need to make a System name.  This should remove the need for all of these changes.

cryptol-remote-api/src/CryptolServer/Data/Expression.hs
```diff
@@ -46,13 +46,14 @@ import Cryptol.ModuleSystem.Monad (runModuleM, interactive)
...
 import Cryptol.ModuleSystem.Name
-  (Name, mkDeclared, NameSource( UserName ), liftSupply, nameIdent)
+  (Name, mkDeclared, liftSupply, nameIdent)
 import Cryptol.ModuleSystem.NamingEnv (singletonNS, shadowing, namespaceMap)
 
 import qualified Cryptol.Parser as CP
 import qualified Cryptol.Parser.AST as CP
 import Cryptol.Parser.Position (Located(..), emptyRange)
 import Cryptol.Parser.Selector
+import Cryptol.Parser.Name (NameSource( UserName ))
```
yav: You should import this from `Cryptol.ModuleSystem.Name` like it was before

src/Cryptol/ModuleSystem/Renamer/Monad.hs
```diff
@@ -350,11 +350,16 @@ checkOverlap env =
...
--- shadow something in the second.
+--   shadow something in the second. This warning is only emited
+--   for UserNames
 checkShadowing :: NamingEnv -> NamingEnv -> RenameM ()
 checkShadowing envNew envOld =
-  mapM_ recordWarning
-    [ SymbolShadowed p x xs | (p,x,xs) <- findShadowing envNew envOld ]
+  mapM_
+    recordWarning
+    [SymbolShadowed p x (keepUserNames xs) | (p, x, xs) <- findShadowing envNew envOld, checkUserName x && not (null (keepUserNames xs))]
```
yav: It would be better to not compute `keepUserNames xs` twice

yav (changes requested):

src/Cryptol/Transform/MonoValues.hs
```diff
@@ -79,7 +79,7 @@
 module Cryptol.Transform.MonoValues (rewModule) where
 
 import Cryptol.ModuleSystem.Name
-        (SupplyT,liftSupply,Supply,mkDeclared,NameSource(..),ModPath(..))
```
Ra1issa: import it from Crypto.MS.Name

PR #1949, opened 2025-12-09
Remove _build directory.

### LINKED ISSUE #1929: Remove `docs/RefMan/_build`
Historically, we have checked in the generated HTML pages for the Cryptol reference manual under `docs/RefMan/_build`, as that way we could ensure that folks could easily look up the manual without needing to build it themselves. These days, however, this step is unnecessary, as we go through the effort of building the reference manual in CI and publishing it to https://galoisinc.github.io/cryptol/master/RefMan.html. As such, checking the generated HTML is a waste of `git` space and just increases the likelihood for PRs that update the manual to have merge conflicts.

I propose that we simply remove `docs/RefMan/_build` and endorse https://galoisinc.github.io/cryptol/master/RefMan.html as the official way to view the manual going forward.

yav:
We build documentation in CI instead.
Fixes #1929

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -187,6 +187,7 @@ jobs:
             ) || (
              echo "Unable to generate documentation for this release" > ${tgt}/docs/RefMan/_build/html/RefMan.html
             )
+            mkdir -p docs/RefMan/_build
             mv ${tgt}/docs/RefMan/_build/html docs/RefMan/_build/$(basename ${tgt})
           done
           cp docs/index.html docs/RefMan/_build/index.html


RyanGlScott:
Looks sensible, although I think the `build-pages-docs` CI job will need to be updated to reflect the fact that the `docs/RefMan/_build` directory no longer exists (and therefore needs to be created):

https://github.com/GaloisInc/cryptol/blob/fb6d4422027c43627ca147cb60cfaf3e96586c5c/.github/workflows/docs.yml#L190

RyanGlScott (approved):

PR #1954, opened 2025-12-17
Bug fix to #1920

naomiiiiiiiii:
Fixes a small bug in `IR/Prove.hs`

Now the user can specify how many results `satProve` produces

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,7 @@
   This will run validation on all files that have not been successfully validated,
   including ones that perviously failed, and have not changed.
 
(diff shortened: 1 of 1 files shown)


RyanGlScott (commented):
Thanks for the PR! The code changes look reasonable.

Since this fixes a bug in a released version of Cryptol, could you add an entry to the changelog that briefly describes this fix? Adding a bullet point to [this section](https://github.com/GaloisInc/cryptol/blob/c0a360fa071b72a67d3c1470d85202e5ccecce9e/CHANGES.md#bug-fixes) should suffice.

RyanGlScott (approved):

PR #1962, opened 2026-01-07
Bump GHC minor versions and move default to 9.6

sauclovian-g:
Matches what we just did in SAW. (https://github.com/GaloisInc/saw-script/pull/2947)

Draft to begin with, because it's quite likely to not build straight off

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -19,10 +19,10 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -yq \
 RUN curl https://i.jpillora.com/chisel! | bash
 
 # Install GHC
-ARG GHCVER="9.4.8"
+ARG GHCVER="9.6.7"
 ENV GHCUP_INSTALL_BASE_PREFIX=/opt \
     PATH=/opt/.ghcup/bin:$PATH
-RUN curl -o /usr/local/bin/ghcup "https://downloads.haskell.org/~ghcup/0.1.22.0/x86_64-linux-ghcup-0.1.22.0" && \
+RUN curl -o /usr/local/bin/ghcup "https://downloads.haskell.org/~ghcup/0.1.50.2/x86_64-linux-ghcup-0.1.50.2" && \
     chmod +x /usr/local/bin/ghcup
 RUN ghcup install cabal --set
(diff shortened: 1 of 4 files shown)


RyanGlScott (approved):

RyanGlScott:
Oh right, I forgot about the fact that there is a variant of the `cryptol-remote-api` Dockerfile that uses a custom GHC build. Ugh.

There are a couple of things that need to be done:

1. In order to build GHC 9.6.7 from source, we need an earlier version of GHC to bootstrap it. The bootstrapping GHC version is declared [here](https://github.com/GaloisInc/cryptol/blob/df2d54f5720fec879444f9b59d2cdea7333b6497/cryptol-remote-api/Dockerfile#L6). A given GHC version supports being bootstrapped with the previous two major releases of GHC, so it likely suffices to bump `GHCVER_BOOTSTRAP` from `9.0.2` to `9.2.8`.
2. The patch that we apply the GHC ([here](https://github.com/GaloisInc/cryptol/blob/df2d54f5720fec879444f9b59d2cdea7333b6497/cryptol-remote-api/ghc-portability.patch)) will likely need to be redone for GHC 9.6.7's source code. (There is a chance that the patch will apply as-is to GHC 9.6.7, but I wouldn't count on it.)

RyanGlScott:
I have spent some time looking into what it would take to implement https://github.com/GaloisInc/cryptol/pull/1962#issuecomment-3726415933. Unfortunately, there's a third step that I forgot about: as of GHC 9.6, the GHC source tree dropped support for its legacy Make-based build system and now only supports the [Hadrian](https://gitlab.haskell.org/ghc/ghc/-/blob/322dd6726b11c7101c28ffb8aeb7cb4cee34ab56/hadrian/README.md) build system going forward. The `cryptol-remote-api` still uses the legacy Make-based build system, so if we wanted to build GHC 9.6 from source, then we'd need to migrate to Hadrian. This is doable, but a bit fiddly.

Thinking about this some more, I am thinking that we should just drop the variant of the `cryptol-remote-api` Docker CI job that uses a custom GHC build. This was introduced about half a decade ago to support a capability that we no longer use, and the amount of effort required to maintain this capability isn't paying its weight. We can always reintroduce this capability at a later date if need be (the patches will still be in the `git` commit history).

PR #1966, opened 2026-01-15
Allow building with GHC 9.10 and 9.12, add 9.12 builds, drop 9.4

sauclovian-g:
If there are code changes we need to work with newer GHC, it would be good to get them into the upcoming 3.5 release.


sauclovian-g:
The build finished; I think what's wrong now is that when I took a hatchet to the CI to avoid doing unnecessary builds, I borked it. It's probably time to drop that change

sauclovian-g:
I rebased, dropped the borked CI change, dropped the hgmp change, and dropped the non-working build hacks. Assuming it builds ok this way, we can probably merge most of the changes, but still need to:
- merge in test-lib and push the deps change there to Hackage, then drop the commit here
- then once that works, drop the last changeset that adds 9.10 and 9.12 builds

sauclovian-g:
Ok, this passed so I think it's reasonable to merge the test-lib change. I have filed that PR (https://github.com/GaloisInc/test-lib/pull/8) and I guess one of us will need to upload to Hackage.

Then we need to decide what builds we want in Cryptol (and SAW) going forward...

sauclovian-g:
ok, there is an updated test-lib on Hackage now so I've dropped the test-lib change here. Assuming this works, the next steps are to rebase on head, then decide what builds we want to run going forward and adjust the last commit accordingly... and then we can merge.

sauclovian-g:
Gah, forgot I needed to update the Hackage index date

sauclovian-g:
ok, this is ready modulo a final decision on which builds to run.

sauclovian-g:
(the force-push rebases on head to pick up the SBV changes)

RyanGlScott (approved):

sauclovian-g:
The force-push reverts the change of the default GHC version to 9.8; it's now still 9.6.7. That will avoid #2054, which we can then fix separately.

PR #1970, opened 2026-01-21
Update the solver snapshot

podhrmic:
Supports https://github.com/GaloisInc/saw-script/pull/2980

No functional change, as the solver version are the same, the difference is the added UBI9 platform.


RyanGlScott (approved):

sauclovian-g:
It's August you should ask, he's coordinating the release. I don't see any huge reason _not_ to merge it, although it might be too late now. @aschwerdfeger-galois up to you

PR #1985, opened 2026-02-04
Prepare release 0.1.3 of the VSCode extension.

aschwerdfeger-galois:
Dependabot has made enough dependency updates since the last release to warrant a patch.


sauclovian-g (commented):
Does this actually bump any of the deps yet? It doesn't look like it.

sauclovian-g (approved):

PR #2018, opened 2026-04-01
fix: make :? ~ and :? - show docs for complement and negate 

### LINKED ISSUE #1952: `:? ~` doesn't work as intended
Presumably, `:? ~` should bring up the documentation for complement, but it doesn't appear to do that currently:
```
Cryptol> :? ~
Undefined name: (~)
Cryptol> :? complement

    complement : {a} (Logic a) => a -> a

Bitwise complement. The prefix notation '~ x'
is syntactic sugar for 'complement x'.
```

MavenRain:
(fixes #1952)

### DIFF
--- a/src/Cryptol/REPL/Help.hs
+++ b/src/Cryptol/REPL/Help.hs
@@ -14,7 +14,7 @@ import Data.Foldable (for_)
 import Control.Monad(when,guard,unless,msum,mplus)
 
 import Cryptol.Utils.PP
-import Cryptol.Utils.Ident(OrigName(..),identIsNormal,identText,isInfixIdent)
+import Cryptol.Utils.Ident(OrigName(..),identIsNormal,identText)
 import qualified Cryptol.Parser.AST as P
 import qualified Cryptol.ModuleSystem as M
 import qualified Cryptol.ModuleSystem.Name as M
@@ -67,9 +67,9 @@ helpForNamed' qname =
 -- so that @:? ~@ and @:? -@ show the right documentation.
 prefixOpRedirect :: P.PName -> Maybe P.PName
 prefixOpRedirect (P.UnQual' i _)
-  | isInfixIdent i, identText i == "~" =
+  | identText i == "~" =
       Just (P.mkUnqual (P.mkIdent "complement"))
-  | isInfixIdent i, identText i == "-" =
+  | identText i == "-" =
       Just (P.mkUnqual (P.mkIdent "negate"))
 prefixOpRedirect _ = Nothing
 
--- a/src/Cryptol/REPL/Help.hs
+++ b/src/Cryptol/REPL/Help.hs
@@ -27,16 +27,16 @@ import Cryptol.TypeCheck.PP(emptyNameMap,ppWithNames)
 
 import Cryptol.REPL.Monad
 
-helpForNamed :: P.PName -> REPL Bool           
-helpForNamed pn =                                  
+helpForNamed :: P.PName -> REPL Bool
+helpForNamed pn =
   -- Prefix operators @~@ and @-@ are syntactic sugar for @complement@
-  -- and @negate@ respectively.  Redirect so that @:? ~@ and @:? -@  
+  -- and @negate@ respectively.  Redirect so that @:? ~@ and @:? -@
   -- display the help for the underlying function instead.
-  helpForNamed' (fromMaybe pn (prefixOpRedirect pn))                          
-                                                                  
-helpForNamed' :: P.PName -> REPL Bool                                         
-helpForNamed' qname =                                                         
-  do fe <- getFocusedEnv 
+  helpForNamed' (fromMaybe pn (prefixOpRedirect pn))
+
+helpForNamed' :: P.PName -> REPL Bool
+helpForNamed' qname =
+  do fe <- getFocusedEnv
      let params = M.mctxParams fe
          env    = M.mctxDecls  fe
          rnEnv  = M.mctxNames  fe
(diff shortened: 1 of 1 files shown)


RyanGlScott (commented):
Thank you for the contribution.

I want to make sure that the output of `:? ~` and `:? -` is understandable for Cryptol users. As it currently stands, however, I think their output is a bit misleading. For instance, if you run `:? ~`, you will get the following output:

```
Cryptol> :? ~

    (~) : {a} (Logic a) => a -> a

Bitwise complement. The prefix notation '~ x'
is syntactic sugar for 'complement x'.
```

This is wrong, however: `(~) : {a} (Logic a) => a -> a` would have you believe that `(~)` is the name of a function, but it's really not. This is even more confusing in the output for `:? -`:

```
Cryptol> :? -

<snip>

            ---------

    (-) : {a} (Ring a) => a -> a -> a

Precedence 80, associates to the left.

Subtract two values.
 * For type [n], subtraction is modulo 2^^n.
 * Structured values are subtracted element-wise.
 * Satisfies 'a - b = a + negate b'.
See also: 'negate'.

            ---------

    (-) : {a} (Ring a) => a -> a

Returns the additive inverse of its argument.
Over structured values, operates element-wise.
The prefix notation '- x' is syntactic sugar
for 'negate x'.

Satisfies 'a + negate a = 0'.
Satisfies 'negate a = ~a + 1' for bitvector values.
```

This reports `(-)` as having two different type signatures: one `{a} (Ring a) => a -> a -> a` and the other `{a} (Ring a) => a -> a`. That's pretty confusing, in my opinion.

-----

As an alternative, what if we added special cases in `helpForNamed` that check if the `PName`'s identifier is `~` or `-`, and if so, just print out the corresponding `:?` text for `complement` or `negate`. For instance, I am imagining something like:

```
Cryptol> :? ~

    complement : {a} (Logic a) => a -> a

Bitwise complement. The prefix notation '~ x'
is syntactic sugar for 'complement x'.
```

In my opinion, the `:?` for `complement` is all you need to display here. It shows the type of `complement` correctly, and the help text indicates that `~ x` desugars to `complement x`, which is enough context to figure out what `~` does.

RyanGlScott:
Hi @MavenRain. It would be helpful to verify that your patch builds locally before pushing, as that can avoid needing to wait for CI turnaround.

MavenRain:
> Hi @MavenRain. It would be helpful to verify that your patch builds locally before pushing, as that can avoid needing to wait for CI turnaround.

That's fair enough.  I don't usually have this kind of trouble, but I tripped myself up working on two branches concurrently.  I'll fix it soon.

src/Cryptol/REPL/Help.hs
```diff
@@ -55,6 +62,17 @@ helpForNamed qname =
...
 
+-- | Prefix operators @~@ and @-@ are syntactic sugar for @complement@
+-- and @negate@ respectively.  Redirect to the underlying function name
+-- so that @:? ~@ and @:? -@ show the right documentation.
+prefixOpRedirect :: P.PName -> Maybe P.PName
+prefixOpRedirect (P.UnQual' i _)
+  | isInfixIdent i, identText i == "~" =
+      Just (P.mkUnqual (P.mkIdent "complement"))
+  | isInfixIdent i, identText i == "-" =
+      Just (P.mkUnqual (P.mkIdent "negate"))
```
RyanGlScott: I don't think the `isInfixIdent` checks here are necessary, given that we are also checking if the identifier has the name `~` or `-` (which are inherently infix due to how Cryptol's parser works).

RyanGlScott (commented):
Thanks, @MavenRain. This is closer, although I think this approach has another problem: what does `-` refer to? For reference, here is what Cryptol currently outputs for `:? -`:

```
Cryptol> :? -

primitive type (-) : # -> # -> #

    `(-) m n` requires:
      • fin n
      • m >= n

Precedence 80, associates to the left.

Subtract numeric types.

            ---------

    (-) : {a} (Ring a) => a -> a -> a

Precedence 80, associates to the left.

Subtract two values.
 * For type [n], subtraction is modulo 2^^n.
 * Structured values are subtracted element-wise.
 * Satisfies 'a - b = a + negate b'.
See also: 'negate'.
```

The help text here reflects `-` as a binary operator (e.g., `x - y`), which has a separate meaning from `-` when used as a prefix operator (e.g., `- x`). With this PR, however, `:? -` only reports help text for `-` as a prefix operator:

```
Cryptol> :? -

    negate : {a} (Ring a) => a -> a

Returns the additive inverse of its argument.
Over structured values, operates element-wise.
The prefix notation '- x' is syntactic sugar
for 'negate x'.

Satisfies 'a + negate a = 0'.
Satisfies 'negate a = ~a + 1' for bitvector values.
```

Given that `-` can be either a prefix or a binary operator, I think `:? -` should report the help text for both uses. That is, something like:

```
Cryptol> :? -

primitive type (-) : # -> # -> #

    `(-) m n` requires:
      • fin n
      • m >= n

Precedence 80, associates to the left.

Subtract numeric types.

            ---------

    (-) : {a} (Ring a) => a -> a -> a

Precedence 80, associates to the left.

Subtract two values.
 * For type [n], subtraction is modulo 2^^n.
 * Structured values are subtracted element-wise.
 * Satisfies 'a - b = a + negate b'.
See also: 'negate'.

            ---------

    negate : {a} (Ring a) => a -> a

Returns the additive inverse of its argument.
Over structured values, operates element-wise.
The prefix notation '- x' is syntactic sugar
for 'negate x'.

Satisfies 'a + negate a = 0'.
Satisfies 'negate a = ~a + 1' for bitvector values.
```

In terms of implementation, I think we could achieve that by changing `helpForNamed` to report the help text both for the original name (`qname`), as well as potentially reporting the help text for any additional names that it desugars to (i.e., what `prefixOpRedirect` computes).

RyanGlScott:
How is this going, @MavenRain? This looks like it is very nearly done---it just needs the (hopefully minor) changes needed to make the output look more like https://github.com/GaloisInc/cryptol/pull/2018#pullrequestreview-4055587489. Let me know if I can help in any way.

yav:
This appears to be abandoned

Issue #1901, opened 2025-09-11
Add support for SAT, QBF, Sharp_SAT solvers


yav:
Currently Cryptol provides a pretty convenient way for experimenting with SMT solvers via the `:proof` and `:check` commands.   It would be nice if we also added support for:
  * SAT solvers, where we use bit blasting to translate bit-vector expressions
  * QBF solvers, which allow alternating quantifiers.    There are some interesting questions about how to examine the models produced by these solvers
  * Sharp SAT solvers, i.e., solvers that can count the number of models that satisfy a formula

A potential solver to target for QBF could be DepQBF: https://github.com/lonsing/depqbf

RyanGlScott:
I recently learned that SAW has a [`sharpSAT` command](https://github.com/GaloisInc/saw-script/blob/b328560d6fe1dcfe9da5a7df49aad410c8df130f/saw-script/src/SAWScript/Interpreter.hs#L6280-L6285) which takes a Cryptol property and computes the number of solutions using a `sharpSAT` subprocess (see the code [here](https://github.com/GaloisInc/saw-script/blob/b328560d6fe1dcfe9da5a7df49aad410c8df130f/saw-central/src/SAWCentral/Builtins.hs#L2197-L2205)). This could be a basis for an implementation of Sharp SAT support in Cryptol.

PR #2023, opened 2026-04-10
feat: add SAT, QBF, and #SAT solver support via DIMACS bit-blasting

MavenRain:
For #1901

### DIFF
--- a/cryptol-remote-api/src/CryptolServer/Sat.hs
+++ b/cryptol-remote-api/src/CryptolServer/Sat.hs
@@ -164,6 +164,7 @@ onlineProveSat proverName cmd hConsing timeout = do
     EmptyResult -> raise $ proverError "got empty result for online prover!"
     CounterExample cexType es -> Invalid cexType <$> convertSatResult es
     ThmResult _ts -> pure Unsatisfiable
+    CountResult _n -> raise $ proverError "model counting not supported via remote API"
     AllSatResult results ->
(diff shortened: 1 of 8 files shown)


RyanGlScott:
Thanks for working on this, @MavenRain!

I think it may be helpful to describe the design of this feature, as there are a fair number of moving parts. Let me see if I understand the approach that you are taking:

* This PR adds `w4-sat` and `w4-qbf` solvers, which can solve a limited number of properties (mostly bitvector-oriented ones) using SAT or QBF solvers. 
* This PR adds a `:count` command, which counts the number of inputs for a particular property. This command only works with the `w4-sat` command, and a separate option controls which particular solver to use (e.g., `sharpSAT`).

Did I get that right?

I think it would be very helpful to describe somewhere in the Cryptol documentation what each solver/command can and can't do. It would also be nice to add some tests to make sure that everything is working as expected.

RyanGlScott:
Regarding the CI failure, the remote API automatically generates its documentation (in `cryptol-remote-api/docs/Cryptol.rst`), and the CI checks that the auto-generated documentation is up to date. Since you added some more solvers, we'll need to re-generate the documentation to keep the part which lists the solvers in sync. You can do so by running:

```
$ cabal -v0 run exe:cryptol-remote-api -- doc > cryptol-remote-api/docs/Cryptol.rst
```

yav:
This appears to be abandoned

PR #2030, opened 2026-04-21
Refactorings to support sawscript

mtullsen:
- see this https://github.com/GaloisInc/saw-script/pull/3148
- see the commits: all are refactorings except one teency fix (which is only used by saw-script)


yav (approved):

mtullsen:
@yav a few more refactors (and one teency fix in one commit) were added here.  Do you want to re-review?

PR #2034, opened 2026-04-24
Look through type synonyms that are not in scope, when pretty printin…

### LINKED ISSUE #2019: Error message drops parentheses when expanding type synonym
Consider the following program, which contains a type error (`g` lacks the constraints needed to call `f`):

```
type CD c d = c + d

f : {a, b, c, d} (a == b * CD c d) => ([a], [b], [c], [d])
f = zero

g : {a, b, c, d} ([a], [b], [c], [d])
g = f
```

When Cryptol produces an error message, however, the constraint it reports lacks parentheses that I would expect to see:

```
[error] at Bug.cry:7:1--7:6:
    in the definition of 'Main::g', at Bug.cry:7:1--7:2,
    we need to show that
      for any type a, b, c, d
        • a == b * c + d
            arising from
            use of expression Main::f
            at Bug.cry:7:5--7:6
```

Note the `a == b * c + d` part, which is wrong. That should say `a == b * (c + d)`!

Note that if I remove the type synonym:

```
f : {a, b, c, d} (a == b * (c * d)) => ([a], [b], [c], [d])
f = zero

g : {a, b, c, d} ([a], [b], [c], [d])
g = f
```

Then the error message looks like what I would expect:

```
[error] at Bug.cry:5:1--5:6:
    in the definition of 'Main::g', at Bug.cry:5:1--5:2,
    we need to show that
      for any type a, b, c, d
        • a == b * (c * d)
            arising from
            use of expression Main::f
            at Bug.cry:5:5--5:6
```

yav:
…g infix operators.

Fixes #2019

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,9 @@
 
 ## Bug fixes
 
+* Fix pretty printing of types in errors messages
+  ([#2019](https://github.com/GaloisInc/cryptol/issues/2019))
+
 * Fix interface constraint scoping.  Interface constraints are kept in
   the order they were declared, but will be floated as early as possible.
   ([#1690](https://github.com/GaloisInc/cryptol/issues/1690))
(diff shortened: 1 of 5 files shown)


RyanGlScott:
What does the output of the program in #2019 look like with these changes? (Worth adding that as a test case.)

PR #2040, opened 2026-05-06
Expose parameters from functor instantiations

### LINKED ISSUE #1699: Rexporting functor parameters in an instantiation
Consider the following example (the use of submodules is not important, it is just easier to work with a single file):
```crytpol
submodule F where
  parameter type n: #
  f: [n] -> [n]
  f x = x

submodule M = submodule F where type n = 8

// g: [M::n] -> [M::n]
g = M::f
```
The definition of `g` is OK, but we cannot write the type signature, at least not in the style in the comment.
The reason for that is that `n` is a module parameter, and its instantiation is not exported my `M`.

## Workaround 1: Manual Export
```cryptol
submodule F where
  parameter type n: #
  type p_n = n             // This "exports" the parameter
  f: [n] -> [n]
  f x = x

submodule M = submodule F where type n = 8

g: [M::p_n] -> [M::p_n]
g = M::f
```
The draw-back to this approach is that you have to make a new name (e.g., `p_n`) and possibly duplicate stuff (e.g., docstrings).

## Workaround 2: Save Parameter Module
```cryptol
submodule F where                                                                
  parameter type n : #                                                              
  f : [n] -> [n]                                                                    
  f x = x                                                                        

// Now `M` has both the parameters and the instantiation                                                   
submodule M where                                                                
  submodule Param where                                                          
    type n = 8                                                                   
  submodule Impl = submodule F { submodule Param }                               
                                                                                         
g: [M::Param::n] -> [M::Param::n]                                                
g = M::Impl::f   
```
The drawback here is that things look more complex than, perhaps, they should be, and one could accidentally use the
wrong `Param` qualifier if a functor has multiple parameters.

## The Challenge
There are two reasons for the current design (i.e., to not make parameters available in the instantiated module):
  1. It is not clear what names to use for interfaces that are imported qualified (e.g., `import interface I as I`).  Note that we cannot just use `x` for parameter `I::x` because the module might already contain `x`.  It is also possible to import the same interface more than once.
  2. We'd need some mechanism the control *which* parameters should be exported.

## Al Way Forward
Here's one possible design that could allow us to "export" functor parameters in common cases:
  1. Only allow "exporting" functor parameters that are imported through an *unqualified* interface.  Importantly, the common case of using a `parameter` block is just syntactic sugar for this.  This should address issue (1) above.
  2. Allow `public`/`private` blocks in interfaces (and by extension, `parameter` blocks).  This indicates the visibility of the parameter in *instantiations*.

Related to (2) there's the question of what should be default visibility of interfaces.   The most backward compatible choice
would be to make them `private` by default (i.e., we'd basically get the current behavior where nothing is exported, and
one would have to add a `public` annotation to exported parameters).   The drawback of this choice is that is the opposite
of other modules, where things are `public` by default.  Because of that we don't actually have a `public` keyword at the
moment but, hopefully, it should not be a big deal to add this.

Thoughts?

yav:
Expose module parameters used to instantiate a functor as a virtual submodule in the instantiate module.

Fixes #1699


docs/RefMan/Modules.rst
```diff
@@ -873,6 +873,51 @@ This, in turn, is syntactic sugar for creating an anonymous module:
...
+It is also possible to import the virtual parameter module directly:
+
+.. code-block:: cryptol
+
+  submodule N where
+    import submodule M::I
+    z = x + 1           // x is in scope from M::I
+
+The name of the virtual submodule corresponds to the ``as`` name used
+in the functor's interface import.  If the functor uses an anonymous
```
RyanGlScott: What happens if the interface is not imported using `as`? That is, what happens if you write:

```
  submodule F where
    import interface submodule I
    f : T
    f = x
```

Will the name of the virtual submodule default to `I` in this case? Worth clarifying here in the manual, and it might also be worth adding a test case for this (assuming that this is allowed).

src/Cryptol/TypeCheck/Module.hs
```diff
@@ -115,16 +118,78 @@ doFunctorInst m f as instMap enclosingInScope doc =
...
+                  , dFixity     = nameFixity defName
+                  , dDoc        = Nothing
+                  }
+           NSType ->
+             do ts <- lookupTSyn paramName
+                case ts of
+                  Just origSyn ->
+                    addTySyn origSyn { tsName = defName }
+                  Nothing -> bad "Missing type synonym for parameter"
+           ns -> bad ("Unexpected namespace for parameter: " ++ show ns)
```
RyanGlScott, line 168: I assume that the `NSConstructor` and `NSModule` cases are unreachable because Cryptol's parser doesn't allow you to define enums or submodules within interfaces?

RyanGlScott (approved):
Nice!

src/Cryptol/TypeCheck/Module.hs
yav: Yep, exactly---we can't have modules or constructors in interfaces.

docs/RefMan/Modules.rst
yav: Yes.  I've updated the docs and tests to have this case as well.

PR #2045, opened 2026-05-11
Issue 1582

### LINKED ISSUE #1582: Support for parameterized interfaces (interface functors)
User feedback suggests that it might be useful to allow interfaces that are parameterized by other interfaces.   For example:
                                                                                    
```crytpol                                                                          
interface module I where                                                         
  type n : #                                                                     
  type constraint (fin n)                                                        
                                                                                 
interface module A where                                                         
  import interface module I    // This is the new feature                        
                                                                                 
  type x : #                                                                     
  type constraint (x <= n)     // `n` is imported from `I`..                     
```                                                                              
                                                                                 
The idea is that an interface import would bring in some names in scope of the interface.   Since interfaces do not contain values,  we would only ever use the type parameters and constraints from the imported interface.
                                                                                 
A parameterized interfaces is similar to a functor in that it cannot be used directly, instead we would have to first instantiate
it, for example:                                                                 
```cryptol                                                                       
                                                                                 
interface module B = A{M}                                                        
```                                                                              
                                                                                 
The more common case, however, would be to instantiate interfaces with other interfaces, for example:
```
module A where
  import interface I
  import interface J { interface I }   // interface `J` was parameterized,
                                       // and we are passing in the parameters from `I`
```

Implementing this is likely closely related to fixing #1581

yav:
Fixes #1582

I still need to write the docs and update changes


src/Cryptol/TypeCheck/Monad.hs
```diff
@@ -1158,10 +1154,10 @@ getCurDecls =
       { mName             = ()
       , mDoc              = mempty
       , mExports          = mempty
+      , mIsIfaceFunctor   = False
       , mParams           = mempty
-      , mParamTypes       = mempty
-      , mParamConstraints = mempty
-      , mParamFuns        = mempty
+      , mParamDecls       = ParamDecls mempty mempty mempty
+      , mOutputParamDecls = ParamDecls mempty mempty mempty
```
RyanGlScott: Perhaps it would be worth defining a `Monoid` instance for `ParamDecls` so that you can write `mempty` here instead of `ParamDecls mempty mempty mempty`. (Similar comments apply to other uses of `ParamDecls mempty mempty mempty` elsewhere in this PR.)

src/Cryptol/Parser/ParserUtils.hs
```diff
@@ -1394,6 +1425,42 @@ mkImport loc impName optInst mbAs mbImportSpec optImportWhere doc =
...
+  where
+  getInst =
+    case (optInst, optImportWhere) of
+      (Just _, Just _) ->
+        errorMessage (srcRange impName)
+          [ "Invalid interface import instantiation."
+          , "Import should have at most one of:"
+          , "  * { } instantiation, or"
+          , "  * where instantiation"
+          ]
```
RyanGlScott, line 1453: Worth including a program in the test suite that triggers this error message.

src/Cryptol/ModuleSystem/Renamer/Error.hs
```diff
@@ -93,15 +93,16 @@ depNameLoc x =
     ImportAt r -> Just r
 
 
-data ModKind = AFunctor | ASignature | AModule
+data ModKind = AFunctor | ASignature | AModule | AIfaceFunctor
```
RyanGlScott: Minor spelling nitpick: this should probably be named "`AnIfaceFunctor`" rather than "`AIfaceFunctor`" if the intent is for the first part to be a definite article.

RyanGlScott (approved):
Minor comments only

PR #2050, opened 2026-05-12
Add expansive test case for floating-point properties

### LINKED ISSUE #2049: More comprehensive testing of floating-point primitives
`Float.cry` has a wide variety of different primitive floating-point operations, and each of them has rather tricky semantics that can be hard to get right. (As evidence, see recent examples where we have gotten the semantics wrong, such as in #2044.)

Unfortunately, our testing story for `Float.cry` leaves a lot to be desired. At present, we have mainly been adding individual regression tests for particular functions and particular inputs, but this only gets you so far. Ideally, we would have a more comprehensive testing story that checks the correctness of floating-point operations across all possible inputs. For instance, we could imagine testing the following properties:

```
prop_fpMulCommutative :
  {e, p} ValidFloat e p => RoundingMode -> Float e p -> Float e p -> Bit
prop_fpMulCommutative m x y = fpMul m x y =.= fpMul m y x

prop_fpMulNegNeg :
  {e, p} ValidFloat e p => RoundingMode -> Float e p -> Float e p -> Bit
prop_fpMulNegNeg m x y = (fpIsNeg x /\ fpIsNeg y) ==> fpIsPos (fpMul m x y)

prop_fpMulPosPos :
  {e, p} ValidFloat e p => RoundingMode -> Float e p -> Float e p -> Bit
prop_fpMulPosPos m x y = (fpIsPos x /\ fpIsPos y) ==> fpIsPos (fpMul m x y)
```

And so on. We could also imagine testing properties that are not expected to hold, e.g.,

```
counterexample_fpMulAssociative :
  {e, p} ValidFloat e p => RoundingMode ->
  Float e p -> Float e p -> Float e p -> Bit
counterexample_fpMulAssociative m x y z =
   ~ (fpMul m x (fpMul m y z) =.= fpMul m (fpMul m x y) z)

counterexample_fpMulDistributive :
  {e, p} ValidFloat e p => RoundingMode ->
  Float e p -> Float e p -> Float e p -> Bit
counterexample_fpMulDistributive m x y z =
  ~ (fpMul m x (fpAdd m y z) =.= fpAdd m (fpMul m x y) (fpMul m x z))
```

A lot of these properties can be proven using `:prove`, but not all. For properties that properties struggle with (especially those involving `Integer`s or `Rational`s), we can use `:check` instead of `:prove`.

Aside from increasing our confidence in Cryptol's existing coverage for floating-point semantics, adding this would have the following additional advantages:

1. When we add SBV support for floating-point operations (#1997), this could help us catch potential mistakes in SBV's encoding (or Cryptol's use thereof) of floating-point semantics.
2. This would provide a starting point for eventual testing of floating-point semantics in SAW (https://github.com/GaloisInc/saw-script/issues/1237).

See also https://github.com/GaloisInc/cryptol/issues/48, which is in similar territory.

RyanGlScott:
See the comments in `FloatPropertiesGeneric.cry` (the file which defines polymorphic floating-point properties) and `GenFloatProperties.hs` (the script which generates the `.icry` file to test) for more details.

Resolves #2049.

### DIFF
--- a/tests/regression/FloatPropertiesGeneric.cry
+++ b/tests/regression/FloatPropertiesGeneric.cry
@@ -1,0 +1,840 @@
+// Various properties about floating-point operations. Naming conventions:
+//
+// - prop_*: This property should hold for all inputs, and an SMT solver can
+//   prove it within a reasonable amount of time.
+// - check_*: This property should hold for all inputs, but an SMT solver
+//   cannot prove it within a reasonable amount of time.
+// - counterexample_*: This property does not hold for all inputs.
+module FloatPropertiesGeneric where
+
+import Float
+
+///// Helpers
+
+// `True` if negative, `False` if positive
+fpSign : {e, p} (ValidFloat e p, p >= 1) => Float e p -> Bit
+fpSign x = head (fpToBits x)
+
+fpExponent : {e, p} (ValidFloat e p, p >= 1) => Float e p -> [e]
+fpExponent x = take`{e} (drop`{1} (fpToBits x))
+
+fpSignificand : {e, p} (ValidFloat e p, p >= 1) => Float e p -> [p - 1]
+fpSignificand x = drop`{1 + e} (fpToBits x)
+
+fpIsPos : {e, p} ValidFloat e p => Float e p -> Bit
+fpIsPos x = ~ (fpIsNeg x)
+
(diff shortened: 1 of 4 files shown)


yav (approved):
Good idea!

PR #2052, opened 2026-05-13
Implement module aliases

### LINKED ISSUE #1591: Aliasing modules
Currently we have the syntax `module M = F { X }` for giving the name `M` to the instantiation of the functor `F` with module parameter `X`. However we cannot simply do `module M = N` where `N` is a regular module, not a functor. Such a feature might be useful for exposing a deeply nested submodule, e.g. `submodule M = submodule A::B::C::M`. Alternatively if we have #1589 then this could be achieved by
```
submodule M where
  extend submodule A::B::C::M
```
or perhaps be syntactic sugar for it.

yav:
Fixes #1591.  Note that this is on top of #2045, so it should be merged after.  I made this with the base branch being `issue_1582`, not `master`, as supposedly this makes it a "stacked" PR, but I am not sure how this shows up when we merge stuff, so we should be a bit careful when merging things to see where they end up...

### DIFF
--- a/tests/modsys/module-aliases/ModAliasCycle.icry.stdout
+++ b/tests/modsys/module-aliases/ModAliasCycle.icry.stdout
@@ -1,0 +1,11 @@
+Loading module Cryptol
+Loading module Cryptol
+Loading module Main
+
+[error] at ModAliasCycle.cry:3:25--3:26
+    Module `B` is not yet defined.
+[error] at ModAliasCycle.cry:4:25--4:26
+    Module `A` is not yet defined.
+[error] Invalid recursive dependency:
+    • submodule Main::A, defined at 3:11--3:12
+    • submodule Main::B, defined at 4:11--4:12
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,10 @@
 * Interfaces may now be parameterized by other interfaces (interface functors).
   ([#1582](https://github.com/GaloisInc/cryptol/issues/1582))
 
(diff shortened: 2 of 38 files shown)


tests/modsys/module-aliases/ModAliasCycle.icry.stdout
```diff
@@ -0,0 +1,11 @@
...
+Loading module Cryptol
+Loading module Main
+
+[error] at ModAliasCycle.cry:3:25--3:26
+    Module `B` is not yet defined.
+[error] at ModAliasCycle.cry:4:25--4:26
+    Module `A` is not yet defined.
+[error] Invalid recursive dependency:
+    • submodule Main::A, defined at 3:11--3:12
+    • submodule Main::B, defined at 4:11--4:12
```
RyanGlScott, line 11: There appears to be something of an error message cascade here. Is it possible to just report the `Invalid recursive dependency` error without the preceding `Module [...] is not yet defined` errors?

RyanGlScott (approved):

tests/modsys/module-aliases/ModAliasCycle.icry.stdout
yav: Yeah, agreed, but it requires a bit of thinking and unfortunately I am out of hours for the month, so let's leave it like this for now, and we could revisit later.  I don't imagine it is a thing that would happen a lot, so it's probably OK to be a bit more verbose.

PR #2053, opened 2026-05-13
Add fin constraints to some primitives in Array.cry.

### LINKED ISSUE #2037: Array primitives have types that are too general
The array primitives are as yet unimplemented (#1652). But it seems like there is no hope of ever implementing them, in any backend, with their current type signatures. They are just too polymorphic.

For example, the primitives `arrayCopy`, `arraySet` and `arrayRangeEqual` are all supposed to operate on ranges of array indices. Their type signatures allow an index type of `[inf]`, which is clearly absurd. We probably need a `fin n` constraint on all three of these:
```
primitive arrayCopy : {n, a} (Array [n] a) -> [n] -> (Array [n] a) -> [n] -> [n] -> (Array [n] a)
primitive arraySet : {n, a} (Array [n] a) -> [n] -> a -> [n] -> (Array [n] a)
primitive arrayRangeEqual : {n, a} (Array [n] a) -> [n] -> (Array [n] a) -> [n] -> [n] -> Bool
```
Either that, or else the index type should be changed from `[n]` to a type `ix` with constraint `Integral ix`.

Primitive `arrayUpdate` also seems to have a too-polymorphic type. Any reasonable implementation would require the ability to compare array indices for equality, so we probably need a constraint `Eq a` here.
```
primitive arrayUpdate : {a, b} (Array a b) -> a -> b -> (Array a b)
```

brianhuffman:
* arrayCopy
* arraySet
* arrayRangeEqual

Fixes #2037.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Language changes
 
+* Primitives `arrayCopy`, `arraySet`, and `arrayRangeEqual` from the
+  built-in `Array` module now have types with `fin` constraints.
+  ([#2037](https://github.com/GaloisInc/cryptol/issues/2037))
+
 * Add typechecker simplification rule: `max(a,a) == a`
   ([#1923](https://github.com/GaloisInc/cryptol/issues/1923))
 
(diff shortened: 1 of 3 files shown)


RyanGlScott:
I take it that these constraints are needed to implement these primitives a certain way in SAWCore?

I have no objection to adding these constraints, although we should advertise the more constrained type signatures in the Cryptol changelog.

brianhuffman:
That is correct: Without these constraints, it is impossible to implement the corresponding operations in SAWCore without `error` cases.

RyanGlScott (approved):

Issue #2054, opened 2026-05-14
Tests fail with sbv > 11.0


sauclovian-g:
Now that we have moved the default GHC version to 9.8 (#2055), using SBV versions greater than 11.0 (or perhaps, greater than 11.5; 11.1 through 11.5 don't build with GHC 9.8) causes test failures as described below. This is because the Z3 version we're using is older than the minimum supported by SBV. SBV upstream is also not particularly interested in providing compatibility for old versions, and to be fair the version we're using is very old, not just a little old.

In order to lift this restriction and use current SBV, we need to update the Z3 we're using for the regression tests. That in turn requires updating the default Z3 in what4-solvers, and _that_ in turn requires updates to SAW to allow it to run alternate Z3 versions and/or provide config to the solver under user control. Currently a couple of things in SAW's test suite do not work with recent Z3. (See https://github.com/GaloisInc/saw-script/issues/2404.) Those updates are in the works, but expected to take a while, so I'm leaving this issue open to track the situation on the Cryptol side.

Note also that:
- SBV does publish a minimum Z3 version upstream, but apparently doesn't check it at runtime; it's possible that we should.
- All of this happened because the CI only tests the builds done with the default GHC version; thus, any problem that arises with libraries or version skew with some other GHC version will not be detected.

Either of those points should maybe get their own issue someday.

Original description follows:

------

Moving the default GHC version to 9.8 (which was going to be part of #1966) causes the tests to be run on the GHC 9.8 build, and this fails with z3-related SBV exceptions, for example:
```
SBV exception:

*** Data.SBV: Unexpected non-success response from Z3:
***
***    Sent      : (define-fun s1 () Int (ubv_to_int s0))
***    Expected  : success
***    Received  : (error "line 8 column 36: unknown constant ubv_to_int ((_ BitVec 2)) ")
***
***    Exit code : ExitFailure (-15)
***    Executable: /home/runner/work/cryptol/cryptol/bin/z3
***    Options   : -nw -in -smt2
***
***    Reason    : Check solver response for further information. If your code is correct,
***                please report this as an issue either with SBV or the solver itself!
```

[...]

PR #2055, opened 2026-05-14
Address test failures and move default GHC to 9.8

sauclovian-g:


sauclovian-g:
This does not fix #2054, but lets us move to GHC 9.8 by default without falling over it.

cryptol.cabal
```diff
@@ -80,7 +80,11 @@ library
                        primitive,
                        process           >= 1.2,
                        rme-what4         ^>= 0.1,
-                       sbv               >= 9.1 && < 14.1,
+                       -- See #2054 before changing the sbv pattern.
+                       -- sbv 11.[1-5] do not build with ghc 9.8, and
+                       -- 11.6+ do not work with the old z3 we have in
+                       -- what4-solvers and use in CI.
+                       sbv               >= 9.1 && <= 11.0,
```
RyanGlScott: For consistency with how we usually dictate version bounds in `.cabal` files, the usual convention is to use `<` for the upper bounds so as to allow minor bugfixes:



(It doesn't matter much here, since it's unlikely that there will be future bugfixes to the `11.0` series, but consistency is nice.)
```suggestion
                       sbv               >= 9.1 && < 11.1,
```

RyanGlScott (approved):

PR #2061, opened 2026-05-21
Fix reference evaluator's implementation of integer division and remainder

### LINKED ISSUE #2060: Concrete backend and reference evaluator have different `Integer` division semantics
The following Cryptol session computes the same expression involving integer division twice, once using the concrete backend and once again using the reference evaluator (`:eval`). I'd expect to get the same answer both times, but they instead return different answers: 

```
Cryptol> (7 : Integer) / (-2 : Integer)
-4
Cryptol> :eval (7 : Integer) / (-2 : Integer)
-3
```

RyanGlScott:
Cryptol's integer division (`/`) is intended to round towards negative infinity, not towards zero, so it should be implemented using `div`, not `quot`. By similar reasoning, integer remainder (`%`) should be implemented using `mod`, not `rem`.

To fix this, I tweaked the behavior of `divWrap` and `modWrap` to use `div` and `mod` instead of `quot` and `rem`. The functions that were previously named `divWrap` and `modWrap` have been renamed to `signedDivWrap` and `signedModWrap`, respectively, as these names reflect the fact that they're only used to implemented signed bitvector division (`/$`) and remainder (`%$`).

Fixes #2060.


brianhuffman:
@RyanGlScott What do you think about the online documentation for `(/)` and `(%)`? Is it sufficient to make it clear to users how division and modulus are supposed to work on negative integers?

RyanGlScott:
Which bit of documentation in particular are you referring to? I see that `(/)`/`(%)` are very briefly mentioned in https://galoisinc.github.io/cryptol/master/OverloadedOperations.html#integral-operations, but that section doesn't really make any effort to define the semantics of any of the mentioned operators. Is that the place you had in mind for adding this sort of extra documentation?

brianhuffman:
I was just thinking of the REPL documentation you get with `:? /` and `:? %`.
```
    (/) : {a} (Integral a) => a -> a -> a

Precedence 90, associates to the left.

Divide two values, rounding down (toward negative infinity).
 * For type [n], the arguments are treated as unsigned.
 * Division by zero is undefined.
```
```
    (%) : {a} (Integral a) => a -> a -> a

Precedence 90, associates to the left.

Compute the remainder from dividing two values.
 * For type [n], the arguments are treated as unsigned.
 * Remainder of division by zero is undefined.
 * Satisfies 'x % y == x - (x / y) * y'.
```
It does say that division rounds toward negative infinity, but it always takes me a couple of mental steps to decode what that phrase means. And I guess the defining equation for `x % y` together with the documentation for `(/)` does fully specify its behavior, but I'd have to think fairly hard to conclude that `%` with a negative modulus should yield a negative remainder.

RyanGlScott:
To clarify the semantics of `(%$)`, perhaps we could mimic what the documentation for `(%$)` does?

```
Cryptol> :? (%$)

    (%$) : {n} (fin n, n >= 1) => [n] -> [n] -> [n]

Precedence 90, associates to the left.

2's complement signed remainder.  Division rounds toward 0.
Division by 0 is undefined.  Satisfies the following for 'y != 0'

 * 'x %$ y == x - (x /$ y) * y'.
 * 'x >=$ 0 ==> x %$ y >=$ 0'
 * 'x <=$ 0 ==> x %$ y <=$ 0'
```

In addition to giving the defining equation `x %$ y == x - (x /$ y) * y'`, this also gives some intuition about what the sign of `x %$ y` will be. We could do something similar for `(%)`, namely:

* `y < 0 ==> x % y <= 0`
* `y > 0 ==> x % y >= 0`

I'm not sure how best to clarify the semantics of `(/)` (and `(/$)`), but perhaps we could also list some additional defining equations. For `(/)`, we could clarify that it is "floor division", i.e.,

* `y != 0 ==> toInteger (x / y) == floor (ratio (toInteger x) (toInteger y))` (a.k.a. the [`bvdiv_char`](https://github.com/GaloisInc/cryptol/blob/bccdc96273516b5ec5c9e70e6d925cf41629f08a/tests/regression/bvdiv.cry#L6-L9) property)

For `(/$)`, we could clarify that it rounds towards zero, i.e.,

* `~ (y == 0 \/ (x == -(2^^(n-1)) /\ y == -1)) ==> toSignedInteger (x /$ y) == trunc (ratio (toSignedInteger x) (toSignedInteger y))` (a.k.a. the [`bvsdiv_char`](https://github.com/GaloisInc/cryptol/blob/bccdc96273516b5ec5c9e70e6d925cf41629f08a/tests/regression/bvdiv.cry#L37-L41) property)
* `(x == -(2^^(n-1)) /\ y == -1) ==> toSignedInteger (x /$ y) == -(2^^(n-1))` (where `-(2^^(n-1))` is the minimum possible signed integer value of bit width `n`)

yav:
So, to be clear there are at least three possible definitions.  To illustrate this here are the results of dividing 5 by 3:
Lib SMT: (reminder always positive)
```
((pos_pos 1)
 (neg_pos (- 2))
 (pos_neg (- 1))
 (neg_neg 2))
```
div: (round toward -inf, reminder matches sign of the amount we are diving by)
```
ghci> [div 5 3, div (-5) 3, div 5 (-3), div (-5) (-3)]
[1,-2,-2,1]
```
quot: (round toward 0, reminder matches the sign of the thing we are diving)
```
ghci> [quot 5 3, quot (-5) 3, quot 5 (-3), quot (-5) (-3)]
[1,-1,-1,1]
```

I am not sure how intentional we were with the definitions in the interpreters, but I think languages like C and Rust tend to use quot/rem semantics for signed division, so I would lean toward that being our interpretation as well.   This does mean that we probably need to do something a bit special when translating to SMT.

RyanGlScott:
> I am not sure how intentional we were with the definitions in the interpreters, but I think languages like C and Rust tend to use quot/rem semantics for signed division, so I would lean toward that being our interpretation as well.

This is the opposite conclusion that was reached in https://github.com/GaloisInc/cryptol/issues/662, for what it's worth. Cryptol's concrete backend currently implements div/mod semantics for `Integer` division, and the symbolic SBV/What4 backends take special care to ensure that this these semantics are also used:

https://github.com/GaloisInc/cryptol/blob/7dc48dc0171bc52d17fd51fdd06815980ae37ed4/src/Cryptol/Backend/SBV.hs#L309-L313
https://github.com/GaloisInc/cryptol/blob/7dc48dc0171bc52d17fd51fdd06815980ae37ed4/src/Cryptol/Backend/What4.hs#L375-L393

Moreover, SAW's SBV backend and What4 backend also take special case to ensure that div/mod semantics are used (as a result of https://github.com/GaloisInc/saw-script/issues/2415). It's really only Cryptol's reference interpreter that's the odd one out. I'd favor just changing the reference interpreter to match everything else, as this PR accomplishes.

yav:
Hm, but our signed division on bit vectors seems to do `quot/rem` (as I would expect), so it seems really confusing to do something different for `Integer`:
```
Cryptol> [5 /$ 3, (-5) /$ 3, 5 /$ (-3), (-5) /$ (-3)] : [_][64]
[0x0000000000000001, 0xffffffffffffffff, 0xffffffffffffffff,
 0x0000000000000001]
 ```

**Comment by RyanGlScott** [comment]
I'm not sure if one convention is unambiguously less confusing than the other---they're just different conventions. Given that we already offer both conventions for bitvectors, perhaps we should offer both conventions for `Integer`s as well? That is, continue to make `(/)`/`(%)` use div/mod semantics for both bitvectors and `Integer`s, but also expose an operation that uses quot/rem semantics for `Integer`s. We might consider generalizing the type of `(/$)`/`(%$)` to work over both bitvectors and `Integer`s, as `(/$)`/`(%$)` already use quot/rem semantics.

**Comment by sauclovian-g** [comment]
Well, yes, except that integer division rounding towards zero is nearly universal in software (and hardware) manifestations of bitvectors. I feel like doing anything else is usually a mistake (although it looks like that ship sailed years ago) but at least needs big bright banner warnings that it doesn't work the way people will expect it to.

**Comment by RyanGlScott** [comment]
I'd hesitate to claim that rounding towards zero is nearly universal, given that languages such as Python, Ruby, Perl, and Tcl round towards negative infinity. (I'm not trying to make a judgment value on whether this was the right call or not, just that there are plenty of languages that pick a different convention.)

I agree that the ship has sailed already, however. For quite a long time, Cryptol has used `(/)` and `(/$)` to denote bitvector division with rounding towards negative infinity and rounding towards zero, respectively. I think we should be consistent about this when dividing `Integer`s as well.

**Comment by yav** [comment]
Interesting---I hadn't thought of just saying that both `/` and `/$` work on negative numbers, but just do different rounding modes.  This makes for a more consistent explanation, so I am more comfortable with merging this PR.

I think that does suggest that we may also want to generalize the type of `/$` to allow using it on `Integer` with the `quot/rem` semantics.

**Comment by yav** [comment]
Hm, but `/` does not match the round toward `-inf` when used on unsigned numbers:
```cryptol
[5 / 3, (-5) / 3, 5 / (-3), (-5) / (-3)] : [_][64]
[0x0000000000000001, 0x5555555555555553, 0x0000000000000000,
 0x0000000000000000]
```

RyanGlScott:
My apologies, I misunderstood what `(/)` does on unsigned integers. I think a more accurate claim is that `(/)` implements the semantics of SMT-LIB's `bvudiv` operation, i.e., first convert each argument to a 2's complement representation and then compute the quotient. By a similar token, `(/$)` implements the semantics of SMT-LIB's `bvsdiv` operation, which uses signed division rounding towards zero. (We could imagine adding a variant of `(/$)` for bitvectors that uses signed division rounding towards negative infinity, but Cryptol doesn't currently have such a thing.)

This leaves my earlier claims about consistency on shaky ground. If Cryptol is trying to be consistent with SMT-LIB, then `(/)` on `Integer`s isn't doing a very good job of this, as `(/)` doesn't implement the semantics of SMT-LIB's `div` operation (which rounds towards zero) properly. As such, I'm much less convinced about my earlier conviction that `(/)` should round towards negative infinity.

RyanGlScott:
Actually, my intuition of "do whatever SMT-Lib does" is still likely not what we want. This is because SMT-LIB's `div` implements neither truncated division (i.e., rounding towards zero, as in C) nor floor division (i.e., rounding towards negative infinity, as in Python), but rather [Euclidean division](https://dl.acm.org/doi/10.1145/128861.128862). (See https://github.com/GaloisInc/cryptol/pull/2061#issuecomment-4510219438 for an example of where `div`'s outputs differ.)

RyanGlScott:
I talked about this with @yav. The consensus was that we should change the semantics of `(/)`/`(%)` to use quot/rem, as this would be the least surprising option. We should also carefully document how `(/)`/`(%)` works for both unsigned bitvectors and `Integer`s (which would now have different semantics for the same operators). This means that we will also have to change the semantics of `(/)`/`(%)` in SAW yet again.

Under this proposal, the reference evaluator's implementation of `(/)` is actually the only one that is correct. As such, this PR is moot, so I'll close this.

Issue #1249, opened 2021-07-22
Missing primitives in the reference interpreter


robdockins:
When we made `pmult`, `pmod` and `pdiv` into primitives, we forgot to  update the reference interpreter.

```
Cryptol> :eval pmod 0xff 0x80
cryptol: You have encountered a bug in Cryptol's implementation.
*** Please create an issue at https://github.com/GaloisInc/cryptol/issues

%< ---------------------------------------------------
  Revision:  ef37437cd0bdfc2e1c8b5faabec25bdaddc97924
  Branch:    what4-online (uncommited files present)
  Location:  [Reference Evaluator]evalPrim
  Message:   Unimplemented primitive
             Cryptol::pmod
CallStack (from HasCallStack):
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-2.11.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/Eval/Reference.lhs:1748:19 in cryptol-2.11.0.99-inplace:Cryptol.Eval.Reference
%< ---------------------------------------------------
```

robdockins:
Maybe a clean solution to this would be to allow primitives to have optional definitions.

robdockins:
I spent a little bit of time working on implementing primitives with optional definitions.  It was tricker than I expected, as the typechecker and module system aren't really set up to consider that primitive definitions may reference other definitions, etc.

Perhaps a better way would be to allow ordinary definitions to have an annotation indicating that they were declared with the `primitive` keyword.  Then, the simulator can look for alternate realizations of those names before falling back on the given definition.

RyanGlScott:
The following primitives are missing definitions in the reference evaluator:

* [x] `toSignedInteger` (https://github.com/GaloisInc/cryptol/pull/1784)
* [x] `deepseq` (https://github.com/GaloisInc/cryptol/pull/1784)
* [ ] `foldl`
* [ ] `foldl'`
* [ ] `foldWhile`
* [ ] `scanl`
* [ ] `pmult`
* [ ] `pdiv`
* [ ] `pmod`
* [ ] `foldWhile`

RyanGlScott:
https://github.com/GaloisInc/cryptol/pull/1784 implements all of the remaining primitives in the reference interpreter that do not have alternative implementations in `lib/Cryptol/Reference.cry` (i.e., it implements `toSignedInteger` and `deepseq`).

RyanGlScott:
[...]

PR #2068, opened 2026-06-12
Add `foldWhile` primitive

### LINKED ISSUE #2028: Add `foldlMaybe` primitive
The idea is that this primitive will behave like fold, but provide a way to stop iteration:
```cryptol
foldlMaybe: {n, a, b} (fin n) => (a -> b -> Option a) -> a -> [n]b -> a
```

marsella:
Closes #2028.

Adds a breaking-fold primitive to Cryptol. I chose `foldWhile` for the name; to me, that makes it easier to infer the behavior of the conditional (e.g. you fold while it returns true, just like a normal loop). Open to debate on that front. 

I added some docs in fairly outdated places, but didn't update the reference manual since similar functions (`foldl`, `scanl`, etc.) are not in there.


RyanGlScott:
Nice! I think the only thing missing is a `foldWhile` implementation in the reference interpreter. Then again, there are quite a few definitions from `Cryptol::Reference` that aren't yet implemented in the reference interpreter (see https://github.com/GaloisInc/cryptol/issues/1249), so perhaps we could defer that task to https://github.com/GaloisInc/cryptol/issues/1249.

Could you:

* Mention `foldWhile` in the `CHANGES.md` file
* Update the new test output in the FFI-related tests

Then I think this will be ready to go.

lib/Cryptol/Reference.cry
```diff
@@ -55,6 +55,24 @@ pmod x y = if y == 0 then 0/0 else last zs
 foldl : {n, a, b} (fin n) => (a -> b -> a) -> a -> [n]b -> a
 foldl f z bs = last (scanl f z bs)
 
+/**
```
yav, line 58: I think the result should always be what the function returned (i.e., we don't need the if).  This is more flexible as the callback function has full control over what's returned.  For translating to Rust this has the nice benefit that the state can travel linearly, so we can update in place.  With the semantics below, we can't just "move" the `z` into `f`, we'd need to clone it instead.  

This is the main reason we decided to go with the `s -> a -> (Bool,s)` form rater than `s -> a -> Option s` form.

src/Cryptol/Eval/Generic.hs
```diff
@@ -1942,6 +1942,38 @@ foldl'V sym =
...
+
+  go _f a [] = a
+  go f a (b:bs) =
+    do
+      f' <- fromVFun sym <$> f a
+      tup <- fromVTuple <$> f' b
+      case tup of
+        [condition, a'] ->
+          do
+            bit <- fromVBit <$> condition
```
yav, line 1973: Related to my other comment I think the 4th argument to `iteValue` should be `a'`

marsella:
Thanks! I updated the tests and `CHANGES.md` here, and the list in #1249. Happy to take on adding the reference impl if it seems important, but I think it'll take me some time to do correctly so I'd rather do it in a separate PR.

yav (changes requested):
I think we should change the semantics a bit as I mentioned in the comments, but otherwise LGTM

yav:
Don't worry about the reference interpreter---I think we have an unrelated bug (which I am about to file), because it looks like the Reference interpreter is written with the idea that the Reference module should be loaded, but the `:eval` command does not seem to do that.  So `foldl` also crashes in the reference interpreter, which is obviously not good.

You've added a definition to `Reference.cry`, so when we fix the loading of this module, I think the new primitive should start working.

lib/Cryptol/Reference.cry
marsella: Ah, that makes sense. Because the function can always compute the bool and decide whether to return the old vs new state, but if this implementation locks us into the old state it can't do anything that "looks ahead" to the first failing element. Fixing.

marsella: That said I think we still need the `if` to decide whether to keep folding or not. I just replaced the `else` case with the updated state.

PR #2094, opened 2026-07-21
Set the prop guard flag properly during ExpandPropGuards

### LINKED ISSUE #2090: `resolveCurBind` panic in XTS module
The weekly cryptol-specs build failed on the [XTS spec](https://github.com/GaloisInc/cryptol-specs/blob/82023f4bee71e8d6f00e24e926b04e33cf2aacd6/Primitive/Symmetric/Cipher/Block/Modes/XTS.cry) with the following error:
```
  Revision:  UNKNOWN
  Branch:    UNKNOWN
  Location:  resolveCurBind
  Message:   Unexpected current binding
             Expected: encrypt(0 == n % 128) /__w/cryptol-specs/cryptol-specs/Primitive/Symmetric/Cipher/Block/Modes/XTS.cry:196:1--196:8
             Actual: encrypt
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.5.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/ModuleSystem/Renamer/Monad.hs:602:11 in cryptol-3.5.0.99-inplace:Cryptol.ModuleSystem.Renamer.Monad
```

yav:
The issue was that NoPat can introduce where clauses so we have to look for lambdas through `wehre` as well.

Fixes #2090

### DIFF
--- a/tests/issues/issue2090.icry.stdout
+++ b/tests/issues/issue2090.icry.stdout
@@ -1,0 +1,7 @@
+Loading module Cryptol
+Loading module Cryptol
+Loading module Main
+[warning] at issue2090.cry:2:9--2:11
+    Unused name: ys
+[warning] at issue2090.cry:2:4--2:6
+    Unused name: xs
--- a/src/Cryptol/Parser/AST.hs
+++ b/src/Cryptol/Parser/AST.hs
@@ -1435,7 +1435,7 @@ instance (Show name, PPName name) => PP (Expr name) where
                                  text "->" </> pp e)
         where
         des = maybe mempty ppNm (funDescrName d)
-        ppNm x = "/*" <.> pp x <.> "*/"
+        ppNm x = "/*" <.> pp x <.> (if funDescrFromPropGuard d then " prop-guard " else mempty) <.> "*/"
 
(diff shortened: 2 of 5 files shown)


tests/issues/issue2090.icry.stdout
```diff
@@ -0,0 +1,7 @@
+Loading module Cryptol
+Loading module Cryptol
+Loading module Main
+[warning] at issue2090.cry:2:9--2:11
+    Unused name: ys
+[warning] at issue2090.cry:2:4--2:6
+    Unused name: xs
```
RyanGlScott, line 7: These warnings are rather strange, aren't they? Normally, you can fix `Unused name: foo` warnings by replacing the binding for `foo` with a wildcard pattern (`_`), but if you replaced `xs` or `ys` here with a wildcard pattern, then the function would no longer typecheck.

yav: Yeah, that's a good point.   Since this is unrelated to the fix in this PR I made https://github.com/GaloisInc/cryptol/issues/2096 to track the issue.

RyanGlScott (approved):

PR #2095, opened 2026-07-21
Simplify away trivial `Prop`s in numeric constraint guards

### LINKED ISSUE #2093: Numeric constraint guard definition fails `coreLint`
The following program causes Cryptol's `coreLint` check to panic on 3.5.0 and on `master` (as of commit 471c5bb69f5ae23677e741e7f24ee1af7af4e2e7):

```
// Test.cry
f : {n} [n] -> Bit
f _ | n != n => False
    | n == n  => True
```
```
$ ~/Software/cryptol-3.5.0/bin/cryptol -c ":set coreLint=on" -c ":load Test.cry"
Loading module Cryptol
Loading module Cryptol
{n, a, b, c} n == min n n
{a, n} (fin n) => inf == inf * (1 * (1 * 1))
{a, n} (fin n) => n / 2 + n /^ 2 == n
{n, a, b} n == min n n
{n} (n >= 1, fin n) => (fin n, n >= 1)
Loading module Main

  Revision:  6173b60ad40822a1563ea4f569b04a70806b67fd
  Branch:    release-3.5.0 (uncommited files present)
  Location:  Core lint failed:
  Message:   Location: :1:1--1:1
             Proof application without a proof abstraction

  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.5.0-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/ModuleSystem/Base.hs:785:29 in cryptol-3.5.0-inplace:Cryptol.ModuleSystem.Base
```

Note that not all definitions with numeric constraint guards have this problem. For instance, the following variation of the program above passes `coreLint`:

```
g : {n} [n] -> Bit
g _ | n != n => False
    | () => True
```

RyanGlScott:
When typechecking a `Schema`, Cryptol sometimes simplifies away `Prop`s corresponding to "trivial" constraints (e.g., `n == n`). As a result, when desugaring numeric constraint guards, we must also be careful to simplify away trivial constraints that are mentioned in a guard, as the type of the corresponding guard function will not have these constraints in its type signature. This ensures that we apply the correct number of `EProofApp`s.

Fixes #2093.

### DIFF
--- a/src/Cryptol/TypeCheck/Solver/Numeric.hs
+++ b/src/Cryptol/TypeCheck/Solver/Numeric.hs
@@ -52,7 +52,11 @@ cryIsEqual ctxt t1 t2 =
 
 -- | Try to solve @t1 /= t2@
 cryIsNotEqual :: Ctxt -> Type -> Type -> Solved
-cryIsNotEqual _i t1 t2 = matchDefault Unsolved (pBin (/=) t1 t2)
+cryIsNotEqual i t1 t2 =
+  matchDefault Unsolved $
+        (pBin (/=) t1 t2)
+    <|> (aNat' t1 >>= tryNeqK i t2)
+    <|> (aNat' t2 >>= tryNeqK i t1)
 
 -- | Try to solve @t1 >= t2@
 cryIsGeq :: Ctxt -> Type -> Type -> Solved
@@ -139,6 +143,25 @@ tryGeqKThan _ ty (Nat n) =
                 Inf   -> [ b =#= tZero ]
                 Nat 0 -> []
                 Nat k -> [ tNum (div n k) >== b ]
+  <|>
+  -- K1 >= K2 ^^ t    ~~> logBase K2 K1 >= t
+  do (rk, b) <- matches ty ((|^|), aNat, __)
+     case genLog n rk of
+       Just (a,True) -> pure $ SolvedIf [ tNum a >== b ]
+       _ -> pure Unsolved
+  <|>
+  -- K1 >= 1 + (K2 ^^ t)    ~~> logBase K2 K1 >= 1 + t
+  --
+  -- Or, equivalently,
+  --
+  -- K1 > K2 ^^ t           ~~> logBase K2 K1 > t
+  do (oneTy,ty') <- anAdd ty
+     oneK <- aNat oneTy
+     guard (oneK == 1)
+     (rk, b) <- matches ty' ((|^|), aNat, __)
+     case genLog n rk of
+       Just (a,True) -> pure $ SolvedIf [ tNum a >== tf2 TCAdd (tNum (1::Int)) b ]
+       _ -> pure Unsolved
 
 -- | Try to solve @t >= K@
 tryGeqThanK :: Ctxt -> Type -> Nat' -> Match Solved
@@ -151,7 +174,21 @@ tryGeqThanK _ t (Nat k) =
      return $ SolvedIf $ if n >= k
                             then []
                             else [ b >== tNum (k - n) ]
-  -- XXX: K1 ^^ n >= K2
+  <|>
+  -- K1 ^^ t >= K2    ~~> t >= logBase K1 K2
+  do (rk, a) <- matches t ((|^|), aNat, __)
+     case genLog k rk of
+       Just (b,True) -> pure $ SolvedIf [ a >== tNum b ]
+       _ ->
+         -- K1 ^^ t >= 1 + K2 ~~> t >= 1 + logBase K1 K2
+         --
+         -- Or, equivalently,
+         --
+         -- K1 ^^ t > K2      ~~> t > logBase K2 K1
+         do guard (k > 0)
+            case genLog (k-1) rk of
+              Just (b,True) -> pure $ SolvedIf [ a >== tNum (1+b) ]
+              _ -> pure Unsolved
 
 
 -- (K >= 2 && K^a >= K^b) => a >= b
@@ -412,6 +449,17 @@ tryEqK ctxt ty lk =
   -- 2  = min (2,y)   --> y >= 2
   -- 10 = min (2,y)   --> impossible
 
+
+-- e.g., 10 = t
+tryNeqK :: Ctxt -> Type -> Nat' -> Match Solved
+tryNeqK _ ty lk =
+
+  -- K1 != K2 ^^ t    ~~> t != logBase K2 K1
+  do (rk, b) <- matches ty ((|^|), aNat, __)
+     return $ case lk of
+                Inf | rk > 1 -> SolvedIf [ b =/= tInf ]
+                Nat n | Just (a,True) <- genLog n rk -> SolvedIf [ b =/= tNum a]
+                _ -> SolvedIf []
 
 -- | K1 * t1 + K2 * t2 + ... = K3 * t3 + K4 * t4 + ...
 tryEqMulConst :: Type -> Type -> Match Solved
--- a/src/Cryptol/TypeCheck/Solver/Numeric.hs
+++ b/src/Cryptol/TypeCheck/Solver/Numeric.hs
@@ -145,9 +145,10 @@ tryGeqKThan _ ty (Nat n) =
                 Nat k -> [ tNum (div n k) >== b ]
   <|>
   -- K1 >= K2 ^^ t    ~~> logBase K2 K1 >= t
-  do (rk, b) <- matches ty ((|^|), aNat, __)
-     case genLog n rk of
-       Just (a,True) -> pure $ SolvedIf [ tNum a >== b ]
+  do let k1 = n
+     (k2, t) <- matches ty ((|^|), aNat, __)
+     case genLog k1 k2 of
+       Just (logBaseK2K1,True) -> pure $ SolvedIf [ tNum logBaseK2K1 >== t ]
        _ -> pure Unsolved
   <|>
   -- K1 >= 1 + (K2 ^^ t)    ~~> logBase K2 K1 >= 1 + t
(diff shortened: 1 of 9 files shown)


RyanGlScott:
A surprising consequence of this change is that Cryptol now cannot prove that the numeric constraint guards in [`tests/issues/issue1796.cry`](https://github.com/GaloisInc/cryptol/blob/471c5bb69f5ae23677e741e7f24ee1af7af4e2e7/tests/issues/issue1796.cry) are exhaustive, whereas it could prove them to be exhaustive before:

```
[warning] at issue1796.cry:3:5--5:24:
  Could not prove that the constraint guards used in defining Main::exhaustive1 were exhaustive.
[warning] at issue1796.cry:9:5--11:24:
  Could not prove that the constraint guards used in defining Main::exhaustive1Reordered1 were exhaustive.
[warning] at issue1796.cry:15:5--17:24:
  Could not prove that the constraint guards used in defining Main::exhaustive1Reordered2 were exhaustive.
[warning] at issue1796.cry:28:5--29:24:
  Could not prove that the constraint guards used in defining Main::exhaustive3 were exhaustive.
```

One example is [`exhaustive1`](https://github.com/GaloisInc/cryptol/blob/471c5bb69f5ae23677e741e7f24ee1af7af4e2e7/tests/issues/issue1796.cry#L1-L5):

```
exhaustive1 : {w} [2 ^^ (2 ^^ w)] -> [2 ^^ (2 ^^ w)]
exhaustive1 x
  | (2 ^^ (2 ^^ w)) < 2 => x
  | (2 ^^ (2 ^^ w)) == 2 => ~ x
  | (2 ^^ (2 ^^ w)) > 2 => x
```

Note that Cryptol's ability to reason about type-level exponentiation (`^^`) is limited, but Cryptol can still see that the type `(2 ^^ (2 ^^ w))` is checked to be less than, equal to, or greater than `2`, which is definitely exhaustive. After the changes in this PR, however, Cryptol now simplifies `(2 ^^ (2 ^^ w)) == 2` to simply `w == 0`. Due to Cryptol's limited reasoning over exponentiation, this means that Cryptol is unsure is the numeric constraint guards check if `(2 ^^ (2 ^^ w))` is ever checked to be equal to `2`, so Cryptol now gives a spurious warning about potentially inexhaustive guards.

I'm not quite sure what to do about this. I suppose we could make Cryptol's simplification of numeric constraint guards a little less aggressive such that it does simplify trivial constraints like `n == n`, but it doesn't perform more subtantial rewrites such as going from `(2 ^^ (2 ^^ w)) == 2` to `w == 0`. Currently, all of these rewrites are implemented in the [`simplify`](https://github.com/GaloisInc/cryptol/blob/471c5bb69f5ae23677e741e7f24ee1af7af4e2e7/src/Cryptol/TypeCheck/SimpleSolver.hs#L24-L35) function. Unless I'm missing some other existing functionality, this suggests that we'd need to split out the trivial-constraint-elimintation functionality in `simplify` so that we can use it on its own for numeric constraint guards.

yav:
Hm, I thought we tried to aways preserve the number of constraints when simplifying and existing schemas, to avoid this problem. This is why we have the "true" and "and" constraints. It might be worth having a look why this is not happening in this case.

RyanGlScott:
> Hm, I thought we tried to aways preserve the number of constraints when simplifying and existing schemas, to avoid this problem.

Unfortunately, this is the not the case, as seen in the current implementation of `checkSchema`:

https://github.com/GaloisInc/cryptol/blob/471c5bb69f5ae23677e741e7f24ee1af7af4e2e7/src/Cryptol/TypeCheck/Kind.hs#L63-L67

We could try not doing this, I suppose, but it's unclear to me what the consequences of doing so are.

yav:
Ah, I was thinking about the schemas in the type-checker syntax, but this happens when we turn the parsed schemas into the TC ones, so it is not that unreasonable to do the simplification, perhaps.

For the time being, I am leaning toward just accepting that we get some spurious warnings for potentially non-exhaustive patterns---this check is incomplete anyways.

In the future, we may want to also update the exhaustiveness check for the prop guards to distinguish between "these are not exhaustive for sure" (as in "here's an example" that would crash), vs. "we couldn't show that these are exhaustive but we also don't know of an example".

yav (approved):

RyanGlScott:
I certainly agree that there will always be some numeric constraint guards for which Cryptol won't be smart enough to conclude that they're exhaustive. (I'm pretty sure the general problem is undecidable.) Still, I think it would be a pity to let this particular test case regress. It's meant to serve as a regression test for https://github.com/GaloisInc/cryptol/issues/1796, where a user requested that this specific program be marked as exhaustive. On the other hand, this PR is meant to fix a somewhat obscure corner case in checking constraint guards. It feels somewhat perverse to let this obscure corner case take precedence over an actual user request!

Looking at `exhaustive1` again, it strikes me that if Cryptol can simplify `(2 ^^ (2 ^^ w)) == 2` to `w == 0`, then it really ought to be the case that Cryptol should be able to simplify `(2 ^^ (2 ^^ w)) > 2` to `w > 0` and `(2 ^^ (2 ^^ w)) < 2` to `w < 0`. I think the reason it can handle the `(2 ^^ (2 ^^ w)) == 2` case is because of this rewrite rule:

https://github.com/GaloisInc/cryptol/blob/f22f0c7e6007216eb96734c492243946a6189ccf/src/Cryptol/TypeCheck/Solver/Numeric.hs#L403-L408

On the other hand, Cryptol doesn't have any corresponding rewrite rules for inequalities. There is a comment [here](https://github.com/GaloisInc/cryptol/blob/f22f0c7e6007216eb96734c492243946a6189ccf/src/Cryptol/TypeCheck/Solver/Numeric.hs#L154) acting as a reminder to add them, though. Perhaps we should!

RyanGlScott:
Commit 8919860 fixes the regression by adding some more typechecker simplification rules. This is now ready for another round of review!

src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -151,7 +174,21 @@ tryGeqThanK _ t (Nat k) =
      return $ SolvedIf $ if n >= k
                             then []
                             else [ b >== tNum (k - n) ]
-  -- XXX: K1 ^^ n >= K2
+  <|>
+  -- K1 ^^ t >= K2    ~~> t >= logBase K2 K1
+  do (rk, a) <- matches t ((|^|), aNat, __)
+     case genLog k rk of
+       Just (b,True) -> pure $ SolvedIf [ a >== tNum b ]
```
yav: Do we need the matching on `True` here?  It looks like the log is on the RHS, so even if we round down, the inequality should still hold?

src/Cryptol/TypeCheck/Solver/Numeric.hs
```diff
@@ -151,7 +174,21 @@ tryGeqThanK _ t (Nat k) =
...
+       _ ->
+         -- K1 ^^ t >= 1 + K2 ~~> t >= 1 + logBase K2 K1
+         --
+         -- Or, equivalently,
+         --
+         -- K1 ^^ t > K2      ~~> t > logBase K2 K1
+         do guard (k > 0)
+            case genLog (k-1) rk of
+              Just (b,True) -> pure $ SolvedIf [ a >== tNum (1+b) ]
+              _ -> pure Unsolved
```
yav, line 201: Ditto

yav (approved):

src/Cryptol/TypeCheck/Solver/Numeric.hs
yav: Wait, shouldn't this be:  `t >= logBase K1 K2`?

As for my previous comment, I think the reasoning is wrong as we'd like to replace only constraints with equivalent ones.  We certainly have that `K1 ^^ t >= K2` implies `t >= floor (logBase K1 K2)`.  I am not 100% sure about the other direction, although I think it might be OK as well...

RyanGlScott: The inequality will still hold, yes, but the inequality may be weaker than before. That may actually matter when it comes to typechecking. Consider this example:

```
f : {n} (2^^(2^^n) >= 3) => [n]
f = [False] # zero // Requires (n > 0) to typecheck
```

This typechecks with both Cryptol 3.5.0 and with the current state of this PR. Now suppose we change the `Just (b,True)` patterns to `Just(b,_exact)` such that `logBase` is allowed to round down. If we apply this change, then Cryptol no longer typechecks the definition above:

```
[error] at Bug.cry:2:1--2:19:
    in the definition of 'Main::f', at Bug.cry:2:1--2:2,
    we need to show that
      for any type n
        • n >= 1
            arising from
            matching types
            at Bug.cry:2:5--2:19
```

The reason this happens is because under the proposed redesign, `2^^(2^^n) >= 3` would "simplify" to `2^^n >= 1`, since `logBase 2 3 = floor(lg2(3)) = 1`. In turn, Cryptol will simplify `2 ^^ n >= 1` to `n >= 0`, which holds trivially. The first rewrite in this simplification process is questionable, however, as it turns a constraint that implies `n > 0` (needed for the body of the function to typecheck) into a constraint that is trivially true.

As such, I don't think we want to apply this rewrite. I can add a comment to this effect if you'd like.

RyanGlScott: See https://github.com/GaloisInc/cryptol/pull/2095/commits/3e03638a12a4568763621c4997d987c27e9ff5de.

yav: I was worried about the rounding too, that's why I wasn't sure.

Isn't the base and the argument of the log swapped though?

`K1 ^^ t >= K2`  we take log base `K1` on both sides to get `t >= logBase K1 K2`  

Btw, it might make sense to use similar names in the comments and the code to make it easier to match the two---I know we are not consistent about this in other places as well, but it wouldn't hurt if we started updating that as we change things.

yav (approved):
Sorry, I realized that some of my comments were in "review" mode so weren't actually posted.

I think the changes generally look good, except for the bit where the arguments of the log look like they are swapped, so once that's fixed I think we are good to merge.

src/Cryptol/TypeCheck/Solver/Numeric.hs
RyanGlScott: Oops, you are correct: the comment should say `K1 ^^ t >= K2    ~~> t >= logBase K1 K2` instead. The implementation in the code was doing this correctly, but the comments were wrong due to a copy-paste error. I've pushed a new version that corrects this and the corresponding references to this rule in the Note.

I also agree that the code is quite out of sync with the accompanying comments, which makes it harder to figure to spot these sorts of inconsistencies. Porting all of the code in `Cryptol.TypeCheck.Solver.Numeric` to be consistent will be a lot of work, but for now, I've opted to just rename the variables in rewrite rules involving `logBase`---see d304e0d0c5eac42af156c39f70afba06c4b43820.

While I was in town, I also flipped the order of arguments to the `genLog` function such that `genLog base x` now models `logBase base x` (previously, it was `genLog x base` instead). See commit 2ff212c2fc01a97b70c728357b0a6633b382793c.

PR #2100, opened 2026-07-28
Fix misleading comments and test cases mentioning NaN's sign bit

### LINKED ISSUE #2099: `Float.cry`: Misleading comments about NaN being positive
The docstring for `fpToBits` in `lib/Float.cry`: currently states the following:

https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/lib/Float.cry#L105-L112

I think the comments about NaN being represented as positive are slightly misleading. While it is true that Cryptol chooses a NaN representation whose sign bit is unset, that doesn't imply that NaN itself is positive. In fact, most floating-point libraries will say that NaN is neither positive nor negative:

* In LibBF, `bfIsNeg bfNaN == bfIsPos bfNaN == False`
* In SMT-LIB, the `fp.isNegative` and `fp.isPositive` operations will both return `false` when given `(_ NaN eb sb)` as an argument.

As such, I think we should revise the docstring here to say something to the effect of "the sign bit is unset" rather than "positive".

By a similar token, there are some misleading definitions and comments in `tests/regression/FloatPropertiesGeneric.cry`:

* https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L14-L16

  This is misleading for the same reasons discussed above. We should change this docstring to something like the following:

  ```
  // `True` if the sign bit is set, `False` otherwise. If a number's sign bit is
  // set, then that usually implies that the number is negative, and inversely,
  // if a number's sign bit is unset, then that usually implies that the number
  // is positive. The one exception to this rule is NaN, whose sign bit is unset
  // but is considered to be neither a negative nor a positive number. Cryptol's
  // NaN simply uses a 0 sign bit as a sentinel value; it otherwise has no
  // mathematical significance.
  ```
* https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L24-L25

  This really ought to be the following instead:

  ```
  fpIsPos x = ~ (fpIsNeg x \/ fpIsNaN x)
  ```
* https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L38-L40

  This is misleading for the same reasons discussed above. I think we should remove this property in favor of something like `prop_fpNaNSign = ~ (fpSign (fpNaN`{e, p}))`
* https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L209-L210

  This property does hold for `x = NaN`, even with the more restrictive definition of `fpIsPos` proposed above. That being said, it is somewhat of a coincidence that it holds for NaN; if Cryptol picked a different representation for NaN, then it might not continue to hold. As such, we might want to change this to:

  ```
  prop_fpIsNegSign x = fpIsNeg x ==> fpSign x
  ```

  By a similar token, if we ever introduce an `fpIsPos` operation in the future (see https://github.com/GaloisInc/cryptol/issues/2046), then we would have the property `fpIsPos x ==> ~ (fpSign x)` (note that the stronger property `fpIsPos x == ~ (fpSign x)` does _not_ hold when `x = NaN`).
* https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L443-L452

  Saying that NaN is defined to be positive is misleading for the same reasons discussed above. With the strengthened definition of `fpIsPos` above, we can change this to be:

  ```
  // Special cases:
  //
  // * Infinite/zero: `-inf * 0 =.= -0 * inf =.= NaN`, and NaN is not considered
  //   to be positive.
  prop_fpMulNegPos :
    {e, p} ValidFloat e p => RoundingMode -> Float e p -> Float e p -> Bit
  prop_fpMulNegPos m x y =
    ~ ((fpIsInf x /\ fpIsZero y) \/ (fpIsZero x /\ fpIsInf y)) ==>
      (fpIsNeg x /\ fpIsPos y) ==> fpIsNeg (fpMul m x y)
  ```

  Similar comments apply [here](https://github.com/GaloisInc/cryptol/blob/2afd2bf6310dd71b325eccf8b84af2c259b65981/tests/regression/FloatPropertiesGeneric.cry#L454-L463) as well.

RyanGlScott:
NaN's sign bit being 0 should not be miscontrued as the NaN value itself being considered positive.

Fixes #2099.

### DIFF
--- a/tests/regression/FloatPropertiesGeneric.cry
+++ b/tests/regression/FloatPropertiesGeneric.cry
@@ -11,7 +11,13 @@ import Float
 
 ///// Helpers
 
-// `True` if negative, `False` if positive
+// `True` if the sign bit is set, `False` otherwise. If a number's sign bit is
+// set, then that usually implies that the number is negative, and inversely,
+// if a number's sign bit is unset, then that usually implies that the number
+// is positive. The one exception to this rule is NaN, whose sign bit is unset
+// but is considered to be neither a negative nor a positive number. The
+// IEEE-754 standard does not interpret the sign of a NaN, so the sign bit has
+// no mathematical meaning.
 fpSign : {e, p} (ValidFloat e p, p >= 1) => Float e p -> Bit
 fpSign x = head (fpToBits x)
 
@@ -22,7 +28,7 @@ fpSignificand : {e, p} (ValidFloat e p, p >= 1) => Float e p -> [p - 1]
 fpSignificand x = drop`{1 + e} (fpToBits x)
 
 fpIsPos : {e, p} ValidFloat e p => Float e p -> Bit
-fpIsPos x = ~ (fpIsNeg x)
+fpIsPos x = ~ (fpIsNeg x \/ fpIsNaN x)
 
 ///// fpNaN
 
@@ -35,9 +41,9 @@ prop_fpNaNIsNotInf = ~ (fpIsInf (fpNaN`{e, p}))
 prop_fpNaNIsNotZero : {e, p} ValidFloat e p => Bit
 prop_fpNaNIsNotZero = ~ (fpIsZero (fpNaN`{e, p}))
 
-// Cryptol chooses to represent NaN as a positive value.
-prop_fpNaNIsPos : {e, p} ValidFloat e p => Bit
-prop_fpNaNIsPos = fpIsPos (fpNaN`{e, p})
+// Cryptol chooses to represent NaN with a sign bit of 0:
+prop_fpNaNSign : {e, p} (ValidFloat e p, p >= 1) => Bit
+prop_fpNaNSign = ~ (fpSign (fpNaN`{e, p}))
 
 prop_fpNaNIsNotNormal : {e, p} ValidFloat e p => Bit
 prop_fpNaNIsNotNormal = ~ (fpIsNormal (fpNaN`{e, p}))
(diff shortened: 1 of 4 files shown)


tests/regression/FloatPropertiesGeneric.cry
```diff
@@ -35,9 +41,9 @@ prop_fpNaNIsNotInf = ~ (fpIsInf (fpNaN`{e, p}))
 prop_fpNaNIsNotZero : {e, p} ValidFloat e p => Bit
 prop_fpNaNIsNotZero = ~ (fpIsZero (fpNaN`{e, p}))
 
-// Cryptol chooses to represent NaN as a positive value.
-prop_fpNaNIsPos : {e, p} ValidFloat e p => Bit
-prop_fpNaNIsPos = fpIsPos (fpNaN`{e, p})
+// Cryptol chooses to represent NaN with a sign bit of 0:
```
yav: This seems like  a really weird function.  Do we need it all?

RyanGlScott: Technically, we don't "need" anything in this file. It's just a collection of properties to test. I found it interesting to test what representation Cryptol happens to choose in practice for NaN (see also the `prop_fpNaNIsQuiet` property below), but if you think that's too intricately tied to Cryptol's internals, then I could just as well remove these properties.

yav (approved):

tests/regression/FloatPropertiesGeneric.cry
RyanGlScott: Thinking about this some more, I think I agree that we shouldn't be testing these, as they rely on Cryptol having coincidentally picked a particular NaN bit pattern. (It is unlikely that Cryptol will ever change this convention, but still, relying on this coincidence feels iffy.)

I've removed the `prop_fpNaNSign` and `prop_fpNaNIsQuiet` properties in favor of more general properties that should always hold for any IEEE-754 NaN bit pattern.

PR #2107, opened 2026-08-06
`Cryptol.Testing.Random`: Fix how subnormal/normal float values are sampled

### LINKED ISSUE #2106: `Cryptol.Testing.Random`'s claims about randomly sampled subnormal/normal `Float` values are incorrect
When randomly generating values of floating-point types (e.g., `Float32` or `Float64`) for commands like `:check`, Cryptol uses the `randomFloat` function:

https://github.com/GaloisInc/cryptol/blob/3ee85db11172d8c16c773431d0ce32dac1d0af7d/src/Cryptol/Testing/Random.hs#L348-L389

As the comments indicate, an effort is made to sample values in a such a way that a certain percentage of values are subnormal or normal values. Unfortunately, the code that implements sampling subnormal and normal values (`genSubnormal` and `genNormal`, respectively) don't actually do what they claim. To explain what I mean, I have distilled the code above into a standalone `cabal` script:

```hs
#!/usr/bin/env cabal
{- cabal:
build-depends: base, libBF, tf-random
-}
module Main (main) where

import Data.Bits
import Data.Ratio
import LibBF
import System.Random.TF
import System.Random.TF.Gen
import System.Random.TF.Instances

main :: IO ()
main = do
    let g = mkTFGen (-432432)
    let (bfSubnormal, _) = genSubnormal e p g
    putStrLn $ "Is " ++ show bfSubnormal ++ " subnormal? " ++ show (bfIsSubnormal opts bfSubnormal)
    let (bfNormal, _) = genNormal e p g
    putStrLn $ "Is " ++ show bfNormal ++ " normal? " ++ show (bfIsNormal opts bfNormal)
  where
    e = 8
    p = 24
    opts = fpOpts e p NearEven

-- generates floats corresponding to subnormal values.  These are
-- values with 0 biased exponent and nonzero mantissa.
genSubnormal :: RandomGen g => Integer -> Integer -> g -> (BigFloat, g)
genSubnormal e p g =
  let (sgn, g1) = random g
      (v, g2)   = randomR (1, bit (fromInteger p) - 1) g1
      bf        = bfFromBits opts v
      bf'       = if sgn then bfNeg bf else bf
  in (bf', g2)
  where
    opts = fpOpts e p NearEven

-- generates floats where the exponent and mantissa are scaled by the size
genNormal :: RandomGen g => Integer -> Integer -> g -> (BigFloat, g)
genNormal e p g =
  let (sgn, g1) = random g
      (ex,  g2) = randomR ((1-emax)*sz `div` 100, (sz*emax) `div` 100) g1
      (mag, g3) = randomR (1, max 1 ((sz*smax) `div` 100)) g2
      r  = fromInteger mag ^^ (ex - widthInteger mag)
      r' = if sgn then negate r else r
   in (floatFromRational e p NearEven r', g3)
  where
    sz :: Integer
    sz = 5

    emax = bit (fromInteger e) - 1
    smax = bit (fromInteger p) - 1

-- | Make LibBF options for the given precision and rounding mode.
fpOpts :: Integer -> Integer -> RoundMode -> BFOpts
fpOpts e p r =
  case ok of
    Just opts -> opts
    Nothing   -> error $ unlines [ "Invalid Float size"
                                 , "exponent: " ++ show e
                                 , "precision: " ++ show p
                                 ]
  where
  ok = do eb <- rng expBits expBitsMin expBitsMax e
          pb <- rng precBits precBitsMin precBitsMax p
          pure (eb <> pb <> allowSubnormal <> rnd r)

  rng f a b x = if toInteger a <= x && x <= toInteger b
                  then Just (f (fromInteger x))
                  else Nothing

-- | Check that we didn't get an unexpected status.
fpCheckStatus :: (BigFloat,Status) -> BigFloat
fpCheckStatus (r,s) =
  case s of
    MemError  -> error "libBF: Memory error"
    _         -> r

-- | Make a floating point number from a rational, using the given rounding mode
floatFromRational :: Integer -> Integer -> RoundMode -> Rational -> BigFloat
floatFromRational e p r rat =
  fpCheckStatus $
  if den == 1 then bfRoundFloat opts num
              else bfDiv opts num (bfFromInteger den)
  where
  opts  = fpOpts e p r

  num   = bfFromInteger (numerator rat)
  den   = denominator rat

-- | Compute the number of bits required to represent the given integer.
widthInteger :: Integer -> Integer
widthInteger x = go' 0 (if x < 0 then complement x else x)
  where
    go s 0 = s
    go s n = let s' = s + 1 in s' `seq` go s' (n `shiftR` 1)

    go' s n
      | n < bit 32 = go s n
      | otherwise  = let s' = s + 32 in s' `seq` go' s' (n `shiftR` 32)
```

This script tests the `genSubnormal` and `genNormal` functions in isolation. Despite their names, they don't actually do what they claim, as they can generate numbers that are not subnormal or normal, respectively. To demonstrate this, the script produces the following output:

```
Is 0x7.a1179p-128 subnormal? False
Is 0 normal? False
```

`genSubnormal` is buggy because of this part of its implementation:

https://github.com/GaloisInc/cryptol/blob/3ee85db11172d8c16c773431d0ce32dac1d0af7d/src/Cryptol/Testing/Random.hs#L377-L379

This should be `bit (fromInteger p - 1) - 1`, not `bit (fromInteger p) - 1`, as we need to ensure that the ones bit of the mantissa is also zero.

`genNormal` is buggy because it is possible to randomly generate an `r' :: Rational` whose numerator is extremely small and denominator is extremely large, and when we convert this to a `BigFloat`, rounding will occur such that the rounded result will be `0` (which is not considered normalized). One way to fix this would be to take advantage of the fact that a float is normalized iff its exponent value's bits are not all zeros and not all ones, i.e.,

```hs
genNormal :: RandomGen g => Integer -> Integer -> g -> (BigFloat, g)
genNormal e p g =
  let (sgn, g1) = random g
      (v, g2)   = randomR (1, bit (fromInteger e) - 2) g1
      v'        = v `shiftL` (fromInteger p - 1)
      bf        = bfFromBits opts v'
      bf'       = if sgn then bfNeg bf else bf
  in (bf', g2)
  where
    opts = fpOpts e p NearEven
```

RyanGlScott:
Fixes #2106.

### DIFF
--- a/src/Cryptol/Testing/Random.hs
+++ b/src/Cryptol/Testing/Random.hs
@@ -369,6 +369,9 @@ randomFloat sym e p w g0 =
 
     -- generates floats uniformly chosen from among all bitpatterns
     genBinary g =
+      -- NB: Use the size (e+p) below: 1 bit for the sign bit, e bits for the
+      -- exponent, and (p - 1) bits for the mantissa for a total of
+      -- (1 + e + (p - 1)) = (e+p) bits.
       let (v, g1) = randomR (0, bit (fromInteger (e+p)) - 1) g
        in (VFloat <$> (fpFromBits sym e p =<< wordLit sym (e+p) v), g1)
 
--- a/src/Cryptol/Testing/Random.hs
+++ b/src/Cryptol/Testing/Random.hs
@@ -54,7 +54,7 @@ import Cryptol.Eval(evalEnumCon)
 import Cryptol.Eval.Type      ( TValue(..), TNominalTypeValue(..), ConInfo(..)
                               , enumTagWidth, isNullaryCon )
 import Cryptol.Eval.Value     ( GenValue(..), ppValue, defaultPPOpts, fromVFun, mkSeq, unsafeToFinSeq, finSeq)
-import Cryptol.TypeCheck.Solver.InfNat (widthInteger, Nat' (..))
+import Cryptol.TypeCheck.Solver.InfNat (Nat' (..))
 import Cryptol.Utils.Ident    (Ident)
 import Cryptol.Utils.Panic    (panic)
 import Cryptol.Utils.RecordMap
@@ -361,12 +361,9 @@ randomFloat sym e p w g0 =
            | x < 10   -> (VFloat <$> (fpNeg sym =<< fpLit sym e p 0), g')
            | x <= sz       -> genSubnormal g'  -- about 10% of the time
            | x <= 4*(sz+1) -> genBinary g'     -- about 40%
-           | otherwise     -> genNormal (toInteger sz) g'  -- remaining ~50%
+           | otherwise     -> genNormal g'     -- remaining ~50%
 
   where
-    emax = bit (fromInteger e) - 1
-    smax = bit (fromInteger p) - 1
-
     -- generates floats uniformly chosen from among all bitpatterns
     genBinary g =
       -- NB: Use the size (e+p) below: 1 bit for the sign bit, e bits for the
@@ -385,14 +382,14 @@ randomFloat sym e p w g0 =
           (v, g2)   = randomR (1, bit (fromInteger p - 1) - 1) g1
        in (VFloat <$> ((if sgn then fpNeg sym else pure) =<< fpFromBits sym e p =<< wordLit sym (e+p) v), g2)
 
-    -- generates floats where the exponent and mantissa are scaled by the size
-    genNormal sz g =
+    -- generates floats corresponding to normal values. These are values where
+    -- the exponent bits are not all zeros and not all ones.
+    genNormal g =
       let (sgn, g1) = random g
-          (ex,  g2) = randomR ((1-emax)*sz `div` 100, (sz*emax) `div` 100) g1
-          (mag, g3) = randomR (1, max 1 ((sz*smax) `div` 100)) g2
-          r  = fromInteger mag ^^ (ex - widthInteger mag)
-          r' = if sgn then negate r else r
-       in (VFloat <$> fpLit sym e p r', g3)
+          (ex, g2)  = randomR (1, bit (fromInteger e) - 2) g1
+          (si, g3)  = randomR (0, bit (fromInteger p - 1) - 1) g2
+          v         = (ex `shiftL` (fromInteger p - 1)) .|. si
+       in (VFloat <$> ((if sgn then fpNeg sym else pure) =<< fpFromBits sym e p =<< wordLit sym (e+p) v), g3)
 
 
 -- | A test result is either a pass, a failure due to evaluating to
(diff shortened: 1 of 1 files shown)


src/Cryptol/Testing/Random.hs
```diff
@@ -361,32 +361,34 @@ randomFloat sym e p w g0 =
...
+    genNormal g =
       let (sgn, g1) = random g
-          (ex,  g2) = randomR ((1-emax)*sz `div` 100, (sz*emax) `div` 100) g1
-          (mag, g3) = randomR (1, max 1 ((sz*smax) `div` 100)) g2
-          r  = fromInteger mag ^^ (ex - widthInteger mag)
-          r' = if sgn then negate r else r
-       in (VFloat <$> fpLit sym e p r', g3)
+          (v, g2)   = randomR (1, bit (fromInteger e) - 2) g1
+          v'        = v `shiftL` (fromInteger p - 1)
+       in (VFloat <$> ((if sgn then fpNeg sym else pure) =<< fpFromBits sym e p =<< wordLit sym (e+p) v'), g2)
```
sauclovian-g: I am likely misreading this, but it looks to me like it's generating a random biased exponent, shifting it left into place (p - 1 because p counts the ones bit) ... then not actually adding in a significand value before creating the float.

RyanGlScott: Oops, you are quite right. I'll add a random significand value as well.

yav (approved):
Seems fine to me.  It seems that previously we tried to bias the generated floats based on the size and we don't do that anymore.  Is that something we are OK with?  The idea with biasing is that we may want to generate more numbers in a range that is likely to be useful to the program in question, which I imagine is usually not super huge.

RyanGlScott:
In principle, I'm not opposed to biasing the sampled floats, although I'd want to make sure we do so in a way that actually guarantees that the resulting float is normalized. Are there examples elsewhere in `Cryptol.Testing.Random` of how to do this sort of biasing over a fixed-length interval like what we are using in `genNormal`? For instance, I note that `randomWord` and `randomIntMod` (which also work over fixed-size intervals) ignore the `sz` argument altogether.

yav:
I am surprised we don't use it for words, I would have expected smaller sizes to correspond to smaller numbers (e.g., most significant bits are set to 0).  For floats we could do something similar, and use rejection sampling if we accidentally generate a "bad" number.   I wonder where we use `sz` at all, maybe for `Integer`?  

Anyway, if we are not doing it at other base types, we should probably leave it as is for now, and we can revisit all base types later if the need arises.

RyanGlScott:
>  I wonder where we use `sz` at all, maybe for Integer?

We use `sz` to influence the sampling of types with unbounded size, such as `Integer`s and `Rational`s. As far as I can tell, that's it. I agree that this is a bit weird, but I think it's fine for now. If we want, we can revisit the treatment of words, `Z`s, and `Float`s later.

Issue #2108, opened 2026-08-10
Cryptol crashes (`arithmetic overflow`) when converting float to bits


RyanGlScott:
The following crashes Cryptol:

```
$ ~/Software/cryptol-3.5.0/bin/cryptol
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.5.0 (6173b60)
https://cryptol.net  :? for help

Loading module Cryptol
Cryptol> :m Float
Loading module Cryptol
Loading module Float
Float> fpToBits (fromInteger 4294967295 : Float32)
cryptol: arithmetic overflow
```

RyanGlScott:
Interestingly, the crash does not occur with the reference evaluator:

```
Float> :eval fpToBits (fromInteger 4294967295 : Float32)
0x4f800000
```

Or the `what4` backend:

```
Float> :sat \bits -> bits == fpToBits (fromInteger 4294967295 : Float32)
Satisfiable
(\bits -> bits == fpToBits (fromInteger 4294967295 : Float32))
  0x4f800000
  = True
(Total Elapsed Time: 0.010s, using "Bitwuzla")
```

So the bug appears to be unique to the concrete backend. The error message is very reminiscent of https://github.com/GaloisInc/what4/issues/378#issuecomment-4473395091, where I observed a bug that arose when failing to apply LibBF's `bfRoundFloat` function to a `BigFloat` that held a value that wasn't representable as a single-precision float. Compare the code in the concrete backend:

https://github.com/GaloisInc/cryptol/blob/41fd5bbb0c7d3f5477a082db7518523899733582/src/Cryptol/Backend/Concrete.hs#L400-L406

To the corresponding code in the reference evaluator:

https://github.com/GaloisInc/cryptol/blob/41fd5bbb0c7d3f5477a082db7518523899733582/src/Cryptol/Eval/Reference.lhs#L1807-L1809

The former uses `bfRoundInt` while the latter uses `bfRoundFloat opts`. Given that only the latter actually makes use of LibBF options (which is where the size of the float itself is encoded), I believe that the latter is actually the correct approach.

RyanGlScott:
Fixed in https://github.com/GaloisInc/cryptol/pull/2109.

PR #2109, opened 2026-08-10
Concrete backend: Fix `fpFromInteger`'s implementation

RyanGlScott:
After converting the input integer to a float, the concrete backend's implementation of `fpFromInteger` was failing to round the float to the appropriate size. This is unlike the reference implementation, which properly rounded the float using `bfRoundFloat`. We now do the same thing in the concrete backend, thereby fixing #2108.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -95,6 +95,10 @@
   constraints (e.g., `n == n`) could generate ill-typed code.
   ([#2093](https://github.com/GaloisInc/cryptol/issues/2093))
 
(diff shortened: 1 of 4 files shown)


yav (approved):
Interesting, apparently round to int produces an "infinite precision" float.  We should make this more obvious in the libBF docs

RyanGlScott:
Good idea. I've opened https://github.com/GaloisInc/libBF-hs/issues/37 to track this.

PR #2118, opened 2026-08-26
Issue 2105

### LINKED ISSUE #2105: Add commands to facilitate easier multi-line REPL input
Currently, the REPL is optimized towards entering single lines of input at a time. This makes it somewhat challenging to interactively copy-paste Cryptol definitions into the REPL. This is especially the case you are copy-pasting from a location that is not a local file on your filesystem, and therefore you can't simply `:load` the file in the REPL.

To this end, it would be convenient if the REPL could make it easier to enter multiple lines of code at once. I am envisioning an interaction like the following, using syntax inspired by GHCi:

```
Cryptol> :{
Cryptol> f : [8]
Cryptol> f = 0x2a
Cryptol> :}
Cryptol> f
0x2a
```

That is, `:{` delimits the start of a multi-line input, and hitting Enter after this separates each line. After entering the `:}` command, the REPL exits multi-line input mode, which then allows you to enter expressions or REPL commands as before (e.g., entering the expression `f` causes the REPL to evaluate it). The idea would be that you could first enter `:{`, paste one or more definitions, hit `:}`, and then proceed.

yav:
Implements multi-line support for the repl.  Fixes #2105

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,13 @@
   intermediate Cryptol passes.
   ([#2000](https://github.com/GaloisInc/cryptol/issues/2000))
 
+* The REPL now supports multi-line definition blocks, delimited by `:{`
+  and `:}`.  Everything between the delimiters is
+  processed as a group of top-level declarations, without needing `\`
+  continuations or a leading `let`, which makes pasting definitions
+  easier.
+  ([#2105](https://github.com/GaloisInc/cryptol/issues/2105))
+
 * Support `prime` constraints in numeric constraint guards.
   ([#1658](https://github.com/GaloisInc/cryptol/issues/1658))
 
--- a/cryptol-repl-internal/REPL/Haskeline.hs
+++ b/cryptol-repl-internal/REPL/Haskeline.hs
@@ -107,11 +107,14 @@ crySession replMode stopOnError =
 
   continueAfter success consumed lineNum status =
     case crSuccess status of
-      False | isBatch && stopOnError -> return (return status)
-      _ -> do goOn <- shouldContinue
-              return $ if goOn
-                then loop (crSuccess status && success) (lineNum + consumed)
-                else return status
+      False | isBatch && stopOnError -> pure (pure status)
+      _ ->
+        do
+          goOn <- shouldContinue
+          pure $
+            if goOn
+              then loop (crSuccess status && success) (lineNum + consumed)
+              else return status
 
 
 -- | The result of reading a chunk of input from the user.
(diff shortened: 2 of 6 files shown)


RyanGlScott:
A couple of odd things I noticed when giving this new feature a test drive:

* Only certain types of declarations appear to be accepted between `:{ ... :}`. For instance, top-level functions are accepted:

  ```
  Cryptol> :{
      2| f : {a} a -> a
      3| f x = x
      4| :}
  ```

  But not `newtype` declarations:

  ```
  Cryptol> :{
      6| newtype T = { unT : [8] }
      7| :}

  Parse error at <interactive>:6:1,
    unexpected: newtype
    expected: a declaration
  ```
* Speaking of which, what's with the line numbers before the vertical bars (e.g., `2|`) on each line?

yav:
Yes, at the moment we support only the things that can appear in `where` blocks (function declarations and type synonym declarations).   I only did the conservative thing, as it was similar to what we could do before (I guess not `type synonyms before, but that was for no reason).   To do full top-level declarations we'd need to update the dynamic environment, which would be good to do, but would require more work, which I think we should separately.

The line numbers you see in the blocks is what you'd see if there was an error in the definition, basically they increase every time you press `\n`, kind of as if the REPL was in an `.icry` file.

RyanGlScott:
> To do full top-level declarations we'd need to update the dynamic environment, which would be good to do, but would require more work, which I think we should separately.

OK. Can you open a separate issue for this?

RyanGlScott:
I found a panic:

```
$ cabal run exe:cryptol
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.5.0.99 (addd889, modified)

Loading module Cryptol
Cryptol> :{
    2| f : {n} (fin n) => [n]
    3| f | n >= 2 => 1
    4|   | n < 2 => 0
    5| :}

  Revision:  addd889f9618c268dfefa3f5bbaa60f0dc7b4dd4
  Branch:    T2046-add-more-float-operations (uncommited files present)
  Location:  checkPropGuardCase
  Message:   Expression is not a variable.
  panic, called at src/Cryptol/Utils/Panic.hs:21:9 in cryptol-3.5.0.99-inplace:Cryptol.Utils.Panic
  panic, called at src/Cryptol/TypeCheck/Infer.hs:1516:13 in cryptol-3.5.0.99-inplace:Cryptol.TypeCheck.Infer
```

yav:
Good catch---previously we didn't allow these so we were not running the prop guards pass.   Should be fixed now.  I opened #2120 to track the other issue.

RyanGlScott (approved):

PR #2122, opened 2026-08-27
Add a timeout setting for the type-checking solver

### LINKED ISSUE #2117: Allow configurable timeout for typechecker-related SMT sessions
If you're really unlucky, Cryptol's typechecker will hang forever if it tries to check code that gives rise to particularly complex SMT queries. Or, more accurately, Z3 will hang forever, and Cryptol's typechecker will wait indefinitely until Z3 finishes. For instance, here is an example that hangs during typechecking if you use Z3 4.8.14 (which is bundled with `what4-solvers` as of the time of writing) or 5.1.0 (which is the most recent Z3 version as of the time of writing):

```
f : {n, k} (fin k, k >= 1, k % 2 == 0, 64 >= width (max 1 (n /^ k) - 1)) => [n /^ k * k]
f = zero : [n / k * k + n % k /^ k * k]
```
```
$ ~/Software/cryptol-3.5.0/bin/cryptol Bug.cry
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.5.0 (6173b60)

Loading module Cryptol
Loading module Main
# Hangs here
```

In such situations, it would be nice to introduce a configurable timeout settings. This would be similar to the existing `proverTimeout` setting, which only applies when you are running proofs (i.e., with the `:prove` command). I propose that we call this new setting `tcTimeout`. `tcTimeout` would be to `tcSolver` as `proverTimeout` is to `prover`.

yav:
Fixes #2117


cryptol-remote-api/docs/Errors.rst
```diff
@@ -156,6 +156,7 @@ Type errors (``20700``–``29999``)
 -  ``20720``: “No include errors”
    ``{ source: String, errors: [String] }``
 -  ``20730``: “Typechecking failed”
-   ``{ source: String, errors: [String] }`` (could be split in future
-   into many separate errors)
+   ``{ source: String, errors: [String] }``, or “Typechecker SMT solver timed
+   out” ``{ timeout: Number }``. Typechecking errors could be split into
+   separate errors in the future.
```
RyanGlScott: Co-opting the same error code for both type errors and timeouts seems unnecessarily confusing. Surely we should give timeouts their own error code?

cabal.GHC-9.6.7.config
```diff
@@ -279,7 +279,7 @@ constraints: any.BoundedChan ==1.0.3.0,
              any.simple-get-opt ==0.4,
              any.simple-sendfile ==0.2.32,
              simple-sendfile +allow-bsd -fallback,
-             any.simple-smt ==0.9.8,
+             any.simple-smt ==0.9.9,
```
RyanGlScott: Note that you'll also need to bump the `index-state` date at the bottom of each freeze file, as `simple-smt-0.9.9` wasn't yet released at the time of the current `index-state`.

tests/issues/issue_2117.cry
```diff
@@ -0,0 +1,4 @@
+module issue_2117 where
+
+f : {n, k} (fin k, k >= 1, k % 2 == 0, 64 >= width (max 1 (n /^ k) - 1)) => [n /^ k * k]
+f = zero : [n / k * k + n % k /^ k * k]
```
RyanGlScott: I think it would be worth documenting which versions of Z3 this is known to hang with. It is conceivable that a future version of Z3 might become smarter and solve this in a reasonable amount of time, at which point we'd need to figure out a different way to test for the property of "complex type that takes a long time to typecheck".

cryptol-remote-api/src/CryptolServer.hs
```diff
@@ -141,7 +183,7 @@ loadedPath = lens _loadedPath (\v n -> v { _loadedPath = n })
 data ServerState =
   ServerState { _loadedModule :: Maybe LoadedModule
               , _moduleEnv :: ModuleEnv
-              , _tcSolver :: SMT.Solver
+              , _tcSolver :: MVar (Maybe SMT.Solver)
```
RyanGlScott, line 186: Can you document here what `Nothing` and `Just` mean in the context of `_tcSolver`?

language-server/server/State.hs
```diff
@@ -27,13 +27,17 @@ data Config = Config {
...
+  -- Cryptol state (`cryEnv`, e.g., loading modules).
   -- We do this instead of taking the `stateRef` so that the server can still
   -- read the state while we are working on computing the new Cryptol state.
 
   cryLog :: MVar (Text -> IO ()),
   -- ^ A callback to use to send messages to the client
 
+  cryTCSolver :: MVar (Maybe Solver),
+  -- ^ The current typechecker solver.  A solver exit clears this, and the
+  -- next module command starts a fresh solver.
```
RyanGlScott: Similarly here.

tests/issues/issue_2117.cry
yav: This is probably unnecessarily complicated, and can be replaced with a simpler example.  As it happens, this does hang `z3` (or at least it takes more than about 10s on my machine), but that's accidental: note that in the `.icry` file we change to solver to a custom shell script that hangs reliably (it is not really a solver at all).    All we need from this example is to make a call to the solver, so that the script can hang.   I'll update the example with something simpler, and add a comment to explain this.

RyanGlScott: If there is a reliable way to make an SMT solver take a predictable amount of time (even if it's a finite amount), I'd love to see it :)

RyanGlScott: Oh, I overlooked the fact that you're using a fake SMT solver. That does make things more predictable!

RyanGlScott (approved):
LGTM, although you'll need to resolve the merge conflicts in the `CHANGES.md` file.

PR #2130, opened 2026-09-03
Relax the constraints on expressiosn in the head of a record update

### LINKED ISSUE #2127: Relax syntax for record updates
Currently, when writing a record update `{ e | x = 2 }` we require that `e` be is an `aexpr`, which means that unless it is simple expressions we need to wrap it in parens.   This is mostly done to avoid parser issues due to 1 lookahead, but it'd be nice to relax this, so one can write `{ f a | x = 2 }`, for example, instead of `{ (f a) | x = 2 }`.

yav:
Fixes #2127


tests/issues/issue_2127.icry
```diff
@@ -0,0 +1,7 @@
+:{
+mkRecord : [8] -> { x : [8][8], y : [8] }
+mkRecord x = { x @i = x + i, y = x + 1 }
```
RyanGlScott, line 3: I had no idea that Cryptol allowed to write things like `{ x @ i = x + 1 }`. Neat!

Is the full range of patterns that you can embed to the left of the equals sign documented somewhere? I ask because...

src/Cryptol/Parser/ParserUtils.hs
```diff
@@ -863,6 +863,77 @@ mkIndexedExpr (ps, ixs) body
...
+-- | Split a left-associated expression application into its head and
+-- arguments.
+splitFieldApp :: Expr PName -> (Expr PName, [Expr PName])
+splitFieldApp = go Nothing []
+  where
+  go mbLoc args expr =
+    case expr of
+      ELocated e r -> go (Just r) args e
+      EApp f x     -> go mbLoc (x : args) f
+      _            -> (at mbLoc expr, args)
```
RyanGlScott, line 910: ...I was about to ask why we don't look through parentheses (`EParens`) here. But now that I check, it appears that Cryptol doesn't support parenthesized patterns in record expressions anyway:

```
Cryptol> { x @ i = i } : { x : [8]Integer }
{x = [0, 1, 2, 3, 4, 5, 6, 7]}
Cryptol> { (x @ i) = i } : { x : [8]Integer }

Parse error at <interactive>:19:3--19:10
  Invalid label in record update.
```

This is fine, I think, although it would be helpful to have a more formal description somewhere of what the intended grammar is so that we can verify if we've covered all of the necessary cases.

RyanGlScott (commented):
Please make sure to document this in the changelog, since this allows Cryptol to typecheck strictly more programs.

src/Cryptol/Parser/ParserUtils.hs
sauclovian-g: ...yes please, these kinds of parser hacks are hard to reason about and expensive to maintain :-(

yav: I added some documentation that function and sequence notation may be used in records.  This is a bit orthogonal to this PR, but seems like a good thing to have.

I also updated the CHANGES file to note the change, but I couldn't find any docs describing to old behavior, so not sure if I should update something else.   I think the new behavior is what one would naturally expect, as it matches the notation for sequence comprehensions.   I imagine the only reason I'd added the atomic expression restriction is that I probably didn't feel like trying to figure out the ambiguity :)

RyanGlScott (approved):

PR #2132, opened 2026-09-03
Refactor CI so that we build and test things by platform.

### LINKED ISSUE #2131: Refactor CI so that test for an OS depend only on the build for that OS
Currently we wait for builds on all platforms to succeed before running tests, which is slow and wasteful.

yav:
Previously we used to do all builds, and only start testing after.

Fixes #2131

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,124 @@
+# GitHub Actions workflows
+
+The workflows in this directory are split into event-triggered entry points and
+reusable workflows called by the main CI workflow.
+
+## CI workflow dependencies
+
+```text
+ci.yml
+├── config
+├── platform matrix (after config)
+│   └── ci-platform.yml
+│       ├── ci-build.yml
+│       └── ci-test.yml (after build, using its artifacts)
+└── ci-images.yml (after config, for scheduled, manual, and release builds)
+```
+
+Each platform in the matrix is processed independently. Tests for a platform
+can therefore start as soon as that platform's build artifacts are available.
+The Docker image workflow does not depend on any platform build.
+
+## Event-triggered workflows
+
+### `ci.yml`
+
+The main Cryptol CI entry point. It runs on pull requests, selected branches
+and tags, a daily schedule, and manual dispatches. Its `config` job determines
+the version, release status, event type, artifact name, and retention period.
+It then invokes `ci-platform.yml` for each platform matrix entry and invokes
+`ci-images.yml` when Docker images should be built.
+
+This workflow has no caller-supplied parameters.
+
+### `docs.yml`
+
+Builds the Cryptol documentation on pull requests, selected branches and tags,
+and manual dispatches. It also assembles and deploys the versioned GitHub Pages
+site when running in the main repository. Its two jobs are independent and may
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,281 @@
...
+      ghc-version:
+        required: true
+        type: string
+      cabal:
+        required: true
+        type: string
+      test-kind:
+        description: "Tests to run: none, test-lib, or full"
+        required: true
+        type: string
```
RyanGlScott, line 18: Can you document in more detail what exactly the distinction between `none`, `test-lib`, and `full` is? I am guessing that `none` builds Cryptol but runs no tests, and that `full` runs the full test suite. Is `test-lib` in the middle somewhere? If so, what tests does it run (and not run)?

.github/workflows/ci.yml
RyanGlScott, line 1: If I'm reading this correctly, various CI YAML files now depend on each other. For instance, `ci.yml` runs after `ci-platform.yml`, `ci-test.yml` runs after `ci-build.yml`, and so on. Could you include a small `README.md` file in `.github/workflows` that documents how this pipeline works? (You can reverse engineer it by squinting at the `uses:` stanzas in each file, but I find this information less discoverable.)

RyanGlScott: In the same vein as other changes in this PR, I wonder if it would make more sense to split out the Docker-related configuration into its own `ci-docker.yml` file. That way, `ci.yml` would simply be a small, top-level YAML file that calls `ci-docker.yml`, as well as the other YAML files.

RyanGlScott (commented):
Awesome work! I saw the CI jobs go by when this was first opened, and I can confirm that this allows the CI to run to the Linux test jobs before the macOS build jobs have a chance to complete.

Out of an abundance of caution, I'd like to suggest waiting until after the upcoming Cryptol release before merging this. This is because there are some aspects of CI that only run on nightly jobs (and not on PR-related jobs), and I'd hate to hold of the impending release in case we have to debug subtle changes that were inadvertently introduced here. (I trust that you didn't intend to cause any such changes here, but I've often introduced them myself without meaning to!)

.github/workflows/README.md
```diff
@@ -0,0 +1,124 @@
...
+Optional secrets:
+
+- `APPLE_P12_CERTIFICATE`
+- `APPLE_P12_PASSWORD`
+- `APPLE_P12_IDENTITY_NAME`
+- `APPLE_P8_KEY_ID`
+- `APPLE_P8_ISSUER_ID`
+- `APPLE_P8_API_KEY`
+- `SIGNING_PASSPHRASE`
+- `SIGNING_KEY`
```
RyanGlScott: I'm not sure it does much good to repeat the names of all of the optional secrets here, especially since there is no accompanying documentation to describe what they mean. If we do wish to document these further, I think it would be better to do so in the YAML file itself.

RyanGlScott (approved):
LGTM (but again, I think we should wait until after the upcoming release to land this).

RyanGlScott:
The 3.6.0 release has been cut, so let's land this!

PR #2135, opened 2026-09-08
Prepare release 3.6.0

aschwerdfeger-galois:


cryptol-remote-api/python/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## next
 
+## 3.6.0 -- 2026-09-08
+
```
RyanGlScott, line 6: It might be worth having a small bullet point like "Sync with the Cryptol 3.6.0 release" so that the 3.6.0 section of the changelog isn't completely empty. (We _could_ make the changelog empty, but it does raise the question of what caused the release in the first place.)

RyanGlScott (approved):

aschwerdfeger-galois:
Issues resolved via force-push:

* The changelog for `cryptol-remote-api` 3.6.0 is no longer empty.
* The post-release development version was originally created with a five-part version number, while the packager for the Windows installer requires four-part version numbers.

PR #2136, opened 2026-09-08
Implement :location command

### LINKED ISSUE #2066: REPL command to print source location of a declaration
At the Cryptol REPL, the `:?` or `:help` command with the name of a function or type will print out the docstring attached to the function or type's declaration. It would be useful to have a similar command, perhaps `:location`, that would print out the source location of the named function or type, including the file name and line number.

The use case I have in mind is where we have another tool that is using the Cryptol REPL as a subprocess, where we send REPL commands and parse results. So we want to have an output that is both human-readable and machine parseable.

yav:
Fixes #2066

### DIFF
--- a/src/Cryptol/REPL/Command.hs
+++ b/src/Cryptol/REPL/Command.hs
@@ -101,7 +101,9 @@ import Cryptol.Parser
     (parseExprWith,parseReplWith,parseDeclsWith,ParseError()
     ,Config(..),defaultConfig
     ,parseModName,parseHelpName,parseImpName)
-import           Cryptol.Parser.Position (replPosition,startOfLine,Range(..),HasLoc(..))
+import           Cryptol.Parser.Position
+                   (replPosition,startOfLine,Range(..),HasLoc(..))
+import qualified Cryptol.Parser.Position as Pos
 import qualified Cryptol.TypeCheck.AST as T
 import qualified Cryptol.TypeCheck.Docstrings as T
 import qualified Cryptol.TypeCheck.Error as T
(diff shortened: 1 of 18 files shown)


src/Cryptol/REPL/Command.hs
```diff
@@ -251,6 +253,16 @@ nbCommandList  =
...
     )
+  , CommandDescr [ ":location" ] ["NAME"] (HelpArg locationCmd)
+    "Display the location where a name is defined."
+    (unlines
+      [ "Locations have the form NAMESPACE:LOCATION:LINE:COLUMN"
+      , "  * NAMESAPCE is one of `value`, `type`, or `module`."
+      , "  * File names are relative to the Cryptol search path when possible,"
+      , "    and absolute otherwise."
+      , "  * Definitions in builtin modules are shown as `module NAME`."
+      , "  * Definitions from the REPL use `interactive`."
```
RyanGlScott: Are all three of these bullet points talking about the format of the `LOCATION` part of this output format? If so, worth clarifying.

src/Cryptol/REPL/Command.hs
```diff
@@ -1578,6 +1590,57 @@ browseCmd input
...
+           pure emptyCommandResult { crSuccess = False }
+      Just pname ->
+        do fe <- getFocusedEnv
+           let names = nub
+                     $ concatMap (\ns -> M.lookupListNS ns pname (M.mctxNames fe))
+                                 [ M.NSValue
+                                 , M.NSConstructor
+                                 , M.NSType
+                                 , M.NSModule
+                                 ]
```
RyanGlScott: Consider using `allNamespaces` instead of manually enumerating all the `Namespace` constructors here.

src/Cryptol/REPL/Command.hs
```diff
@@ -1578,6 +1590,57 @@ browseCmd input
...
+       case M.lookupTCEntity (M.nameTopModule name) menv of
+         Just lm ->
+           case M.lmFilePath lm of
+             M.InFile {} ->
+               show <$> io (M.makeRelativeToSearchPath menv (M.lmModuleId lm))
+             M.InMem {} -> pure ("module " ++ pretty (M.lmName lm))
+         Nothing -> pure "interactive"
+     pure
+       (namespace ++ ":" ++ prefix ++ ":" ++
+        show (Pos.line (from loc)) ++ ":" ++ show (Pos.col (from loc)))
```
RyanGlScott: For symmetry with the help text, I wonder if "`prefix`" here should be named "`location`" instead.

RyanGlScott (approved):

PR #2140, opened 2026-09-14
Ensure fresh `Rational`s have non-zero denominators in What4 backend

### LINKED ISSUE #2020: `:sat` crashes when querying zero `Rational` value with What4-based prover
Cryptol can use `:sat` to query a zero `Rational` value. With an SBV-based prover, this works as expected:

```
$ ~/Software/cryptol-3.5.0/bin/cryptol
┏━╸┏━┓╻ ╻┏━┓╺┳╸┏━┓╻
┃  ┣┳┛┗┳┛┣━┛ ┃ ┃ ┃┃
┗━╸╹┗╸ ╹ ╹   ╹ ┗━┛┗━╸
version 3.5.0 (6173b60)

Loading module Cryptol
Cryptol> :set prover=sbv-z3
Cryptol> :sat \(x : Rational) -> x == 0
Satisfiable
(\(x : Rational) -> x == 0) (ratio 0 1) = True
(Total Elapsed Time: 0.017s, using "Z3")
```

With a What4-based prover, on the other hand, this crashes the REPL:

```
Cryptol> :set prover=w4-z3
Cryptol> :sat \(x : Rational) -> x == 0
Satisfiable
cryptol: Ratio has zero denominator
```

RyanGlScott:
We implement an approach similar to what is done in the SBV backend, where fresh `Rational` values add an assumption (via `addDefEqn`) that the denominator is non-zero. This avoids crashing the Cryptol REPL when attempting to use fresh `Rational` values (e.g., with the `:sat` command).

Note that this causes the `:prove QordCompatible` check in the `rational_properties` test to start returning `UNKNOWN` with the `w4-cvc4` solver, whereas the check previously passed. This is a curious change, but not unprecedented, as several other checks in the same test also cause CVC4 to return `UNKNOWN` (e.g., `:prove QordTrans`). Moreover, every other solver I've tried on this check either times out or returns an `UNKNOWN`-like result. As such, I'm going to accept the change.

Fixes #2020.


RyanGlScott:
Very curiously, this change caused the `rational_properties` test's `:prove QordCompatible` check to go from being solvable to returning `UNKNOWN` (when using the `w4-cvc4` solver). I am not entirely sure why CVC4 was able to solve it before, but I don't think this is worth spending a lot of time investigating, as various other properties in the same file (e.g., `QordTrans`) also cause `w4-cvc4` to return `UNKNOWN`. It is also worth noting that every other solver I tried on this check (`w4-cvc5`, `w4-yices`, `w4-z3`, etc.) either times out or returns a result similar to `UNKNOWN`.

src/Cryptol/Symbolic/What4.hs
```diff
@@ -356,12 +356,32 @@ freshBV sym w hi =
        Nothing -> pure ()
      pure x
 
+freshBoundedInt ::
```
yav, line 361: Might be nice to add a comment noting that the bounds are inclusive.

RyanGlScott: Done in https://github.com/GaloisInc/cryptol/commit/54947fe091afd92e0fc275467dad7fe2d6e79285.

